PR #60, opened 2024-12-11
fix: return an error for non-2xx responses using error_for_status()

coreyphillips:
- Previously, the request function returned `Ok(())` even if the HTTP response status was an error (4xx/5xx). By adding `response.error_for_status()?`, we correctly propagate HTTP errors rather than masking them, ensuring that timeouts and other non-2xx status codes produce an error as expected.

### DIFF
--- a/pubky/src/shared/auth.rs
+++ b/pubky/src/shared/auth.rs
@@ -143,10 +143,12 @@ impl PubkyClient {
         path_segments.push(&channel_id);
         drop(path_segments);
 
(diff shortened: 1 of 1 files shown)


Nuhvi:
@coreyphillips do you need this to be published to crates io or can you just use main branch? I am in the middle of a very large update so there will be a new version by the end of the year anyway.

coreyphillips:
@Nuhvi, I can just use the main branch until the next update.

PR #63, opened 2025-01-13
fix: ARM (Apple Silicon) build support in Dockerfile

catch-21:
Still defaults to x86. Pass build arg when using ARM architecture: `docker build --build-arg TARGETARCH=aarch64 .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,10 @@
 # ========================
 FROM rust:1.82.0-alpine3.20 AS builder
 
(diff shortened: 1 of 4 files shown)


SHAcollision (approved):
Thank you @catch-21 , I can't test but looks good! Feel free to merge

catch-21:
I tested on M3 Mac with the new arg and x86 AMD linux machine without the new arg.

PR #67, opened 2025-02-13
fix: Dockerfile build pubky-testnet

catch-21:


SHAcollision (approved):
If it works, it works :heavy_check_mark: 

We might anyway want have one Dockerfile per binary and separate `homeserver` and `pubky-testnet` services.

Dockerfile
```diff
@@ -45,11 +45,11 @@ COPY Cargo.toml Cargo.lock ./
 COPY . .
 
 # Build the project in release mode for the MUSL target
-RUN cargo build --release --bin pubky-homeserver --target $TARGETARCH-unknown-linux-musl
+RUN cargo build --release --bin pubky-homeserver --bin pubky-testnet --target $TARGETARCH-unknown-linux-musl
```
Nuhvi, line 48: do you still need to run the homeserver? If so, I can just run one in the binary of the testnet as well as an http_relay...

Nuhvi: I will do you one better, I publish `pubky-testnet` crate tonight, and you remove this entire thing, and wherever you need docker just `cargo install pubky-testnet` and then run it, no need for the docker file to be here.

catch-21: No, it's for optionality. I've now made it build one or the other, defaulting to testnet.

catch-21:
> We might anyway want have one Dockerfile per binary and separate homeserver and pubky-testnet services.

@SHAcollision It now builds one or the other, defaulting to testnet.

Dockerfile
catch-21: Just seen your second message. As long as we can retain control of compatibility, I'm open to this.

Nuhvi: Published `pubky-testnet` and now you can test it as I described.

Nuhvi:
Other than what I said about deleting this docker entirely ... I don't have an opinion one way or another, so I can merge it if you want.

catch-21:
Let's merge. I'll try pulling from cargo another time, need to understand how it might affect my version control.

PR #73, opened 2025-02-20
fix(tests): update apis on docs

SHAcollision:
As we mostly use `cargo nextest run` , we stoped testing our readmes, as it seems only `cargo test` is now running those snippets. As consequence, the readmes have been outdated/untested for some time.

This fixes `cargo test` . Therefore it seems introducing `tokio-shared-rt` is not necessary as seen in https://github.com/pubky/pubky-core/pull/71

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -20,7 +20,7 @@ anyhow = "1.0.95"
 base64 = "0.22.1"
 clap = { version = "4.5.29", features = ["derive"] }
 pubky = { path = "../pubky" }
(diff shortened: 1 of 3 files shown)


tipogi (commented):
yes it was not need it that macro, you are right! it was my machine default limitation with the current file descriptor limits. Updated the limit (`ulimit -n [number]`) and all the tests pass 🚀 I will close the PR

tipogi (approved):

MiguelMedeiros (approved):

PR #76, opened 2025-02-22
fix: improve testnet start up times in some environments

SHAcollision:
This PR reduces the number of Dht testnet nodes from 10 to 3. 

Both should suffice for testing. But 3 might improve start up times in some specific environments where `mainline::Testnet` after the `5.3.1` fix still takes a considerable amount of time to start for some unknown reason. More context in https://github.com/pubky/mainline/issues/52#issuecomment-2676219391


MiguelMedeiros (approved):
ACK

PR #77, opened 2025-02-22
chore: add doctests to CI

SHAcollision:
Some crates were testing 0 tests. For example `pubky-testnet` only test is the Readme snippet. This is because `cargo nextest` does not support doctests yet. We can run `cargo test --docs` sequentially. This can be simplified in the future if `cargo nextest` starts to support doctests.


MiguelMedeiros:
Good catch! ACK!

MiguelMedeiros (approved):

PR #80, opened 2025-02-26
feat: signup tokens

### LINKED ISSUE #74: feat: homeserver signup tokens
There is 2 features here: **1) admin endpoints, and 2) signup tokens. Luckily for us: we can use admin endpoints to create new signup tokens. And we can use a special signup token, to create an admin. BOOOOOOM 🤯** 
   
1. **Signup Tokens:**  
   - After the initial admin is created, new signup tokens can be generated through **admin** endpoint `/generate_tokens`.  
   - These tokens can potentially be time-limited, one-time use, as needed.

2. **Admin/Master Signup Token:**  
   - Defined in `config.toml` and used only during initial setup.  
   - The first user who registers using this token will gain **Admin privileges**.  
   - This approach allows for secure bootstrap of the first admin user without pre-existing endpoints.
   - This willl allow us to reuse all existing pubky id tooling for admins as well. An admin is just a pubky user with a different role.

# Implementation Plan Proposal
I've been looking in detail for every step involved into implementation of this feature. Of course, every plan is to be broken and improved as we go :) Consider this a "mental dry run" of how to go about it:

**Admin Infrastructure**
- [ ] Extend the user schema to support roles. For example, we could add a simple `admin: bool` field.
- [ ] Implement admin authentication middleware: Create a helper/middleware that validates a session’s token for admin privileges.
- [ ] Reserve a new API route prefix (e.g. `/admin`) for admin endpoints.

**Add a "master" signup token**
Users who do signup with the "master" signup token, will get the `admin: bool` `true` .
- [ ] Update `config.toml` and the configuration parser to add an optional field `master_signup_token`. Extend the `Config` struct to include this master token.

**Signup Tokens Table**
- [ ] Define a new LMDB table (e.g. `signup_tokens`). Design the schema with fields:  
    - `token: String`  
    - `created_at: u64`  
    _- `expires_at: Option<u64>`  (just an idea: not needed just yet, we don't need codes to expire for now!)_
    - `used: Option<PublicKey>` (if `None`, token is unused; if `Some(user_id)`, record the consuming user). Advanced: could (should?) also be a vector of users, just in case we want to allow multiple signups with a token (e.g., we need it if we want several admins with the same `master_signup_token`)
- [ ] Implement serialization and deserialization for the signup token record.
- [ ] Update the migration scripts to create this table.

**Modify the Signup Endpoint**
- [ ] Update the client API signature to include a `signup_token` query parameter.
- [ ] In the signup handler:

1.  Extract the `signup_token` from query parameters.
2.  **Master Token Check:**  
2.1. If the provided token equals the configured master token, proceed and mark the new user as admin.
3. **Regular Token Check:**  
3.1. If not master, look up the token in the `signup_tokens` table.
3.2. Verify that the token exists, is not expired, and is unused.
3.3. Mark the token as used by storing the consuming user’s public key.
4. Continue with the existing AuthToken verification and session creation.
5. Update the user record creation to store the proper role (admin vs. regular).

 **Admin Endpoints for Signup Token Management**
- [ ] Implement an endpoint (e.g. `POST /admin/generate_signup_token`):
  - [ ] Secure this endpoint using the admin middleware.
  - [ ] Generate a secure random token (e.g. 16 random bytes, BASE32 encoded, all caps, possibly add hyphens for readability, would look like `JBSW-Y3DP-EHPK-3PXQ-XR6M-7N2F-LM` this).
  - [ ] Create a signup token record with the current timestamp and (optionally) an expiration.
  - [ ] Save the new token to the `signup_tokens` table and return the token.

Advanced: optionally in the future add endpoints for listing and revoking tokens. These would be useful for the Homeserver Dashboard.
- List Endpoint: Return a list of tokens with metadata (creation time, usage, expiration).
- Revoke Endpoint: Allow an admin to mark a token as revoked or delete it from the table.

**Pubky Client**
- [ ] Make sure the Pubky Client is updated with the new param of `client.signup(... signup_token)`
- [ ] Add a new method for admins `client.gen_signup_token()` that returns a new valid signup token.
- [ ] Implement tests and release for native and wasm

**Optional Signup Tokens**
- [ ] Please, allow a homeserver to be created and ran without requiring signup tokens. The client will submit a `None` instead. This will be tremendously useful, so testing does not become cumbersome. Every single test creates a homeserver, we don't want to always need to signup a admin, generate a token, etc in all testing workflows. We could also make it so `testnet` homeserver are always codeless/unprotected by default.

SHAcollision:
Fixes #74

- [x] admin password and new config
- [x] new table for signup_tokens
- [x] new `/admin/generate_signup_token` endpoint
- [x] homeserver can run with and without signup token requirements
- [x] client can signup with a valid token. Fails to signup with an invalid token. And can signup always when no token required.
- [x] signup is not a wrapper of signin anymore. Now, sign in does not create new users (lmao). Sign up will return an error if user existed already.
- [x] Client test for native signup flow with a signup token.
- [x] Wasm API updated and wasm building
- [x] Client test for wasm signup flow with a signup token.

### DIFF
--- a/pubky-homeserver/src/config.example.toml
+++ b/pubky-homeserver/src/config.example.toml
@@ -1,6 +1,14 @@
 # Secret key (in hex) to generate the Homeserver's Keypair
 # secret_key = "0000000000000000000000000000000000000000000000000000000000000000"
 
+[admin]
+# Set an admin password to protect admin endpoints.
+# WARNING: if no password is set, the admin endpoints will be unprotected.
+password = "mySecretAdminPassword"
+# Set signup_mode to "open" to allow anyone to signup a new user,
+# otherwise, "closed" (the default) to require a valid invite token on signup;
+signup_mode = "closed"
+
 [database]
 # Storage directory Defaults to <System's Data Directory>
 #
--- a/pubky-homeserver/src/config.example.toml
+++ b/pubky-homeserver/src/config.example.toml
@@ -3,7 +3,7 @@
 
 [admin]
 # Set an admin password to protect admin endpoints.
-# WARNING: if no password is set, the admin endpoints will be unprotected.
+# If no password is set, the admin endpoints will not be accessible. 
 password = "mySecretAdminPassword"
 # Set signup_mode to "open" to allow anyone to signup a new user,
 # otherwise, "closed" (the default) to require a valid invite token on signup;
--- a/pubky-homeserver/src/config.example.toml
+++ b/pubky-homeserver/src/config.example.toml
@@ -4,10 +4,10 @@
 [admin]
 # Set an admin password to protect admin endpoints.
 # If no password is set, the admin endpoints will not be accessible. 
-password = "mySecretAdminPassword"
+password = "admin"
 # Set signup_mode to "open" to allow anyone to signup a new user,
-# otherwise, "closed" (the default) to require a valid invite token on signup;
-signup_mode = "closed"
+# otherwise, "token_required" (the default) to require a valid invite token on signup;
+signup_mode = "token_required"
 
 [database]
 # Storage directory Defaults to <System's Data Directory>
--- a/pubky-homeserver/src/config.rs
+++ b/pubky-homeserver/src/config.rs
@@ -41,7 +41,6 @@ struct LegacyBrowsersTompl {
     pub domain: Option<String>,
 }
 
-// Add these new types near the top:
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 struct AdminToml {
     pub password: Option<String>,
--- a/pubky-homeserver/src/config.rs
+++ b/pubky-homeserver/src/config.rs
@@ -44,7 +44,7 @@ struct LegacyBrowsersTompl {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 struct AdminToml {
     pub password: Option<String>,
-    /// "open" or "closed" (defaults to "closed", i.e., a signup token is required)
+    /// "open" or "token_required" (defaults to "token_required", i.e., a signup token is required)
     pub signup_mode: Option<String>,
 }
 
@@ -192,7 +192,7 @@ impl TryFrom<ConfigToml> for Config {
                 password: admin_toml.password.clone(),
                 signup_mode: match admin_toml.signup_mode.as_deref() {
                     Some("open") => SignupMode::Open,
-                    _ => SignupMode::Closed,
+                    _ => SignupMode::TokenRequired,
                 },
             }
         } else {
@@ -283,7 +283,7 @@ mod tests {
                     ..Default::default()
                 },
                 admin: AdminConfig {
-                    password: Some("test_admin_password".to_string()),
+                    password: Some("admin".to_string()),
                     signup_mode: SignupMode::Open
                 }
             }
--- a/pubky-homeserver/src/config.rs
+++ b/pubky-homeserver/src/config.rs
@@ -283,7 +283,7 @@ mod tests {
                     ..Default::default()
                 },
                 admin: AdminConfig {
-                    password: None,
+                    password: Some("test_admin_password".to_string()),
                     signup_mode: SignupMode::Open
                 }
             }
--- a/pubky-homeserver/src/config.rs
+++ b/pubky-homeserver/src/config.rs
@@ -10,7 +10,10 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use crate::{core::CoreConfig, io::IoConfig};
+use crate::{
+    core::{AdminConfig, CoreConfig, SignupMode},
+    io::IoConfig,
+};
 
 // === Core ==
 pub const DEFAULT_STORAGE_DIR: &str = "pubky";
@@ -46,29 +49,6 @@ struct AdminToml {
     pub signup_mode: Option<String>,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Default)]
-pub enum SignupMode {
-    Open,
-    #[default]
-    Closed,
-}
-
-#[derive(Debug, Clone, PartialEq, Eq, Default)]
-pub struct AdminConfig {
-    /// The password used to authorize admin endpoints.
-    pub password: Option<String>,
-    /// Determines whether new signups require a valid token.
-    pub signup_mode: SignupMode,
-}
-
-impl AdminConfig {
-    pub fn test() -> Self {
-        AdminConfig {
-            password: None,
-            signup_mode: SignupMode::Open,
-        }
-    }
-}
 #[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq)]
 struct IoToml {
     pub http_port: Option<u16>,
--- a/pubky-homeserver/src/config.rs
+++ b/pubky-homeserver/src/config.rs
@@ -38,6 +38,37 @@ struct LegacyBrowsersTompl {
     pub domain: Option<String>,
 }
 
+// Add these new types near the top:
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+struct AdminToml {
+    pub password: Option<String>,
+    /// "open" or "closed" (defaults to "closed", i.e., a signup token is required)
+    pub signup_mode: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub enum SignupMode {
+    Open,
+    #[default]
+    Closed,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct AdminConfig {
+    /// The password used to authorize admin endpoints.
+    pub password: Option<String>,
+    /// Determines whether new signups require a valid token.
+    pub signup_mode: SignupMode,
+}
+
+impl AdminConfig {
+    pub fn test() -> Self {
+        AdminConfig {
+            password: None,
+            signup_mode: SignupMode::Open,
+        }
+    }
+}
 #[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq)]
 struct IoToml {
     pub http_port: Option<u16>,
@@ -51,9 +82,9 @@ struct IoToml {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 struct ConfigToml {
     secret_key: Option<String>,
-
     database: Option<DatabaseToml>,
     io: Option<IoToml>,
+    admin: Option<AdminToml>,
 }
 
 /// Server configuration
@@ -63,9 +94,20 @@ pub struct Config {
     ///
     /// Defaults to a random keypair.
     pub keypair: Keypair,
-
     pub io: IoConfig,
     pub core: CoreConfig,
+    pub admin: AdminConfig,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            keypair: Keypair::random(),
+            io: IoConfig::default(),
+            core: CoreConfig::default(),
+            admin: AdminConfig::default(),
+        }
+    }
 }
 
 impl Config {
@@ -112,17 +154,8 @@ impl Config {
                 ..Default::default()
             },
             core: CoreConfig::test(),
+            admin: AdminConfig::test(),
             ..Default::default()
-        }
-    }
-}
-
-impl Default for Config {
-    fn default() -> Self {
-        Self {
-            keypair: Keypair::random(),
-            io: IoConfig::default(),
-            core: CoreConfig::default(),
         }
     }
 }
@@ -175,14 +208,26 @@ impl TryFrom<ConfigToml> for Config {
             }
         };
 
+        let admin = if let Some(admin_toml) = value.admin {
+            AdminConfig {
+                password: admin_toml.password.clone(),
+                signup_mode: match admin_toml.signup_mode.as_deref() {
+                    Some("open") => SignupMode::Open,
+                    _ => SignupMode::Closed,
+                },
+            }
+        } else {
+            AdminConfig::default()
+        };
+
         Ok(Config {
             keypair,
-
             io,
             core: CoreConfig {
                 storage,
                 ..Default::default()
             },
+            admin,
         })
     }
 }
@@ -258,6 +303,10 @@ mod tests {
 
                     ..Default::default()
                 },
+                admin: AdminConfig {
+                    password: None,
+                    signup_mode: SignupMode::Open
+                }
             }
         )
     }
--- a/pubky-homeserver/src/core/database/tables/entries.rs
+++ b/pubky-homeserver/src/core/database/tables/entries.rs
@@ -431,7 +431,7 @@ impl<'db> EntryWriter<'db> {
     }
 }
 
-impl<'db> std::io::Write for EntryWriter<'db> {
+impl std::io::Write for EntryWriter<'_> {
     /// Write a chunk to a Filesystem based buffer.
     #[inline]
     fn write(&mut self, chunk: &[u8]) -> std::io::Result<usize> {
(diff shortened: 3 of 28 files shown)


SHAcollision:
@SeverinAlexB made a big push here. All the basic functionality is in place. Lots of room to improve and clean up here and there. Consider it a draft to iterate over, but it would already benefit from a review and another pair of eyes on it.

Managed to build the wasm and API is alright, it's untested though. Will try to implement wasm tests for signup with tokens tmr.

pubky-homeserver/src/core/database/tables/entries.rs
```diff
@@ -431,7 +431,7 @@ impl<'db> EntryWriter<'db> {
     }
 }
 
-impl<'db> std::io::Write for EntryWriter<'db> {
+impl std::io::Write for EntryWriter<'_> {
```
SHAcollision, line 434: `clippy` did these. I guess it makes sense?

SeverinAlexB: I think that's ok but not sure.

SeverinAlexB:
Something more that comes to my mind: What are John requirements again? Do tokens expire? Do we need the ability to deactivate/delete them?

pubky-homeserver/src/config.example.toml
```diff
@@ -1,6 +1,14 @@
 # Secret key (in hex) to generate the Homeserver's Keypair
 # secret_key = "0000000000000000000000000000000000000000000000000000000000000000"
 
+[admin]
+# Set an admin password to protect admin endpoints.
+# If no password is set, the admin endpoints will not be accessible. 
+password = "mySecretAdminPassword"
```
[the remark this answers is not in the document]
SHAcollision: I agree. I take this is a very early iteration of "admin passwords" with the only purpose of getting signup codes working. A proper nominative admin system will be needed eventually to know "who performed this admin action" as this seems to be a requirement.

This file needs to be kept secret in its current design because it also has the homeserver private key in it. That's arguably even more sensitive than admin password because once leaked, it cannot be changed.

Let me know your thoughts, because my feeling is to leave it "as is" but we can of course be improve this. Storing a plaintext password in a config file is not that unusual and potential users likely know how to keep this file out of public access, after all, Bitcoind, `bitcoin.conf`, MySql `my.cnf`, rabbitmq.conf, reddis.conf, Django's settings.py, and many other store unsalted plaintext passwords in one way or another.

pubky-homeserver/src/config.example.toml
```diff
@@ -1,6 +1,14 @@
...
+# Set signup_mode to "open" to allow anyone to signup a new user,
+# otherwise, "closed" (the default) to require a valid invite token on signup;
+signup_mode = "closed"
```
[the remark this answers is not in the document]
SHAcollision: What do you think about `restricted` ? I think it's my OCD but if it's only 2 options I like to keep it Enums to 1 word. We could make it a `bool` too. `signup_open = false`

pubky-homeserver/src/config.rs
```diff
@@ -175,14 +187,26 @@ impl TryFrom<ConfigToml> for Config {
...
+            AdminConfig {
+                password: admin_toml.password.clone(),
+                signup_mode: match admin_toml.signup_mode.as_deref() {
+                    Some("open") => SignupMode::Open,
+                    _ => SignupMode::Closed,
+                },
+            }
+        } else {
+            AdminConfig::default()
+        };
```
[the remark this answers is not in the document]
SHAcollision: Yes, agree. I thought about this as well. It's clucky that each config struct needs 2 variants `*Config` and `*ConfigToml` . Though I get this is because not all config is intended to be set from the `.toml` file.

pubky-homeserver/src/core/database/tables/signup_tokens.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl DB {
+    pub fn generate_signup_token(&mut self) -> anyhow::Result<String> {
+        let token = generate_random_token();
+        let signup_token = SignupToken {
+            token: token.clone(),
+            created_at: Timestamp::now().as_u64(),
+            used: None,
+        };
```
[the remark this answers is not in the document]
SHAcollision: Nice!

pubky-homeserver/src/core/routes/admin.rs
```diff
@@ -0,0 +1,48 @@
...
+use axum::{
+    extract::State,
+    http::{HeaderMap, StatusCode},
+    response::IntoResponse,
+};
+
+pub async fn generate_signup_token(
+    State(mut state): State<AppState>,
+    headers: HeaderMap,
+) -> Result<impl IntoResponse> {
```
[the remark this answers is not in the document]
SHAcollision: Yes, we can call this endpoint from anywhere we want. I will document how to generate a new token using `curl` .

pubky-homeserver/src/core/database/tables/signup_tokens.rs
```diff
@@ -0,0 +1,87 @@
...
+            self.tables
+                .signup_tokens
+                .put(&mut wtxn, token, &signup_token.serialize())?;
+            wtxn.commit()?;
+            Ok(())
+        } else {
+            anyhow::bail!("Invalid token");
+        }
+    }
+}
```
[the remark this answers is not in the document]
SHAcollision: Yes, makes sense!

pubky-testnet/src/lib.rs
```diff
@@ -107,6 +107,16 @@ impl Testnet {
         Homeserver::run_test(&self.dht.bootstrap).await
     }
 
+    /// Returns a [HomeserverBuilder] preconfigured with this testnet's DHT bootstrap nodes.
+    pub fn homeserver_builder(&self) -> HomeserverBuilder {
+        // Use the same test admin password as the run_test() homeserver.
+        let mut builder = Homeserver::builder().admin_password("test_admin_password".to_string());
```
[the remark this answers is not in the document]
SHAcollision: sure!

pubky/pkg/test/auth.js
```diff
@@ -11,7 +11,7 @@ test('Auth: basic', async (t) => {
   const keypair = Keypair.random()
   const publicKey = keypair.publicKey()
 
-  await client.signup(keypair, HOMESERVER_PUBLICKEY )
+  await client.signup(keypair, HOMESERVER_PUBLICKEY, null)
```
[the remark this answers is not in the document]
SHAcollision: Yes. It is not required here. I left here mostly as a pointer to come back and place an actual token. But first we need to implement a `pubky-testnet` with hardcoded configuration that ALSO requires a signup token so we can also add a test token signups from the nodejs/browser.

pubky/src/native/api/auth.rs
```diff
@@ -23,26 +23,51 @@ impl Client {
...
+        signup_token: Option<&str>,
+    ) -> Result<Session> {
+        // 1) Construct the base URL: "https://<homeserver>/signup"
+        let mut url = Url::parse(&format!("https://{}", homeserver))?;
+        url.set_path("/signup");
+
+        // 2) If we have a signup_token, append it to the query string.
+        if let Some(token) = signup_token {
+            url.query_pairs_mut().append_pair("signup_token", token);
+        }
```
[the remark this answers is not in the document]
SHAcollision: True, will fix.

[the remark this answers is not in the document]
SHAcollision: Moving `signup_token` into the POST body makes sense to align with expected REST semantics (avoiding side effects in GET parameters). However, the current approach keeps AuthToken as raw bytes, and signup_token does not belong inside it. To handle this, we'd need to introduce a structured payload (e.g., JSON with `auth_token` as bytes and `signup_token` as an optional string), requiring additional dependencies (`serde`, `serde_bytes`) on both the homeserver and client.

Before committing to JSON, we should confirm whether this aligns with the original design choice of using raw bytes. Alternatively, we could send both as concatenated bytes, but that might be less clear. If we prefer to avoid unnecessary complexity for now, would it be acceptable to keep `signup_token` as a URL parameter until we fully commit to a structured request format?

SHAcollision:
I fixed the PR comments that were attainable by today. Those that are not fixed, I left a comment on the tradeoffs. We can take it slowly and in the meantime release `v0.4.2.rc1` 

I added instructions for generating signup tokens. These will now be needed on @catch-21 and @MiguelMedeiros e2e-tests . A default `pubky-testnet` homeserver can we queried to generate a new signup token like this:

from JS

```js
const url = `http://${homeserver_ipport}/admin/generate_signup_token`;
const response = await pubky_client.fetch(url, {   // would work with regular `fetch()` too
  method: "GET",
  headers: {
    "X-Admin-Password": "admin", // use your admin password, defaults to testnet password.
  },
});
const signupToken = await response.text();
```

via CLI with `curl`

```bash
curl -X GET "https://<homeserver_ip:port>/admin/generate_signup_token" \
     -H "X-Admin-Password: admin"
     # Use your admin password. This is testnet default pwd.
```
or from rust

```rust
let response = pubky_client
    .get(&format!("https://{homeserver_pubkey}/admin/generate_signup_token"))
    .header("X-Admin-Password", "admin") // Use your admin password. This is testnet default pwd.
    .send()
    .await
    .unwrap();
let signup_token = response.text().await.unwrap();
```

pubky-homeserver/src/config.example.toml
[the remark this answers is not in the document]
SeverinAlexB: Let's keep it simple in this PR. We do have some time for refactors and we can check more advanced secret strategies then.

pubky/src/native/api/auth.rs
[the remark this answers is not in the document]
SeverinAlexB: I am fine with hacking it for the moment to deliver the feature but we should solve and refactor this soon.

pubky-homeserver/src/config.example.toml
```diff
@@ -1,6 +1,14 @@
...
 # secret_key = "0000000000000000000000000000000000000000000000000000000000000000"
 
+[admin]
+# Set an admin password to protect admin endpoints.
+# If no password is set, the admin endpoints will not be accessible. 
+password = "admin"
+# Set signup_mode to "open" to allow anyone to signup a new user,
+# otherwise, "token_required" (the default) to require a valid invite token on signup;
+signup_mode = "token_required"
+
```
SeverinAlexB, line 11: I don't like the `admin` section. What has the `signup_mode` to do with admin? Why not put both in the general section like the secret key?

We need to refactor the config.rs anyway in the future. So maybe we should merge it as it is and then do it better later?

pubky-homeserver/src/core/database/tables/signup_tokens.rs
```diff
@@ -0,0 +1,87 @@
...
+use serde::{Deserialize, Serialize};
+
+pub const SIGNUP_TOKENS_TABLE: &str = "signup_tokens";
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct SignupToken {
+    pub token: String,
+    pub created_at: u64,
+    /// If Some(pubkey), the token has been used.
+    pub used: Option<PublicKey>,
```
SeverinAlexB, line 19: NIT: rename to `used_by`? Not so important

pubky-homeserver/src/core/database/tables/signup_tokens.rs
```diff
@@ -0,0 +1,87 @@
...
+        &self,
+        token: &str,
+        user_pubkey: &PublicKey,
+    ) -> anyhow::Result<()> {
+        let mut wtxn = self.env.write_txn()?;
+        if let Some(token_bytes) = self.tables.signup_tokens.get(&wtxn, token)? {
+            let mut signup_token = SignupToken::deserialize(token_bytes);
+            if signup_token.used.is_some() {
+                anyhow::bail!("Token already used");
+            }
```
SeverinAlexB, line 73: NIT: Create a method `is_used()` in `impl SignupToken`

pubky-homeserver/src/core/routes/auth.rs
```diff
@@ -1,79 +1,154 @@
...
+};
+use std::collections::HashMap;
 use tower_cookies::{cookie::SameSite, Cookie, Cookies};
 
-use pubky_common::{crypto::random_bytes, session::Session, timestamp::Timestamp};
-
-use crate::core::{database::tables::users::User, error::Result, AppState};
-
+/// Creates a brand-new user if they do not exist, then logs them in by creating a session.
+/// 1) Check if signup tokens are required (signup mode is Closed).
```
SeverinAlexB, line 22:
```suggestion
/// 1) Check if signup tokens are required (signup mode is token_required).
```

pubky-homeserver/src/core/routes/auth.rs
```diff
@@ -1,79 +1,154 @@
...
+    let users = state.db.tables.users;
+    if users.get(&txn, public_key)?.is_some() {
+        return Err(Error::new(
+            StatusCode::CONFLICT,
+            Some("User already exists"),
+        ));
+    }
+    txn.commit()?;
+
+    // 3) If signup_mode == closed, require & validate a `signup_token` param.
```
SeverinAlexB, line 49:
```suggestion
    // 3) If signup_mode == token_required, require & validate a `signup_token` param.
```

pubky-homeserver/src/core/routes/auth.rs
```diff
@@ -1,79 +1,154 @@
...
-            },
-        )?;
+    let user_exists = users.get(&txn, public_key)?.is_some();
+    txn.commit()?;
+    if !user_exists {
+        return Err(Error::new(
+            StatusCode::NOT_FOUND,
+            Some("User does not exist"),
+        ));
     }
```
SeverinAlexB, line 104: Comment, no need to take action: "does user exist" seems to be a common question and in the future, we should think of a extracting these methods in it's own struct next to the tables/database. Maybe a repository pattern?

pubky/pkg/test/utils.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Util to request a signup token from the given homeserver as admin.
+ *
+ * @param {Client} client - An instance of your client.
+ * @param {string} homeserver_address - The homeserver's public key (as a domain-like string).
+ * @param {string} [adminPassword="admin"] - The admin password (defaults to "admin").
+ * @returns {Promise<string>} - The signup token.
+ * @throws Will throw an error if the request fails.
+ */
+export async function getSignupToken(client, homeserver_address ="localhost:6286", adminPassword = "admin") {
```
SeverinAlexB, line 11: `get` indicates that the token already exists. Can we do `createSignupToken`?

SeverinAlexB (changes requested):
We are getting there!

pubky-homeserver/src/config.example.toml
SHAcollision: Yes, I think there simply was not a right place to placed these two. Needs refactoring. So we could leave as is and refactor next.

pubky-homeserver/src/core/database/tables/signup_tokens.rs
SHAcollision: We could rename, but that would be a breaking change against currently deployed prod homeserver that will need some sort of migration script. Maybe not worth it?

pubky-homeserver/src/core/routes/auth.rs
SHAcollision: Nice catches!

pubky-homeserver/src/core/routes/auth.rs
SHAcollision: Yes, agree

pubky/pkg/test/utils.js
SHAcollision: Yes, that makes sense

pubky-homeserver/src/core/database/tables/signup_tokens.rs
SeverinAlexB: Let's keep it as it is and potentially address this when we look over the db setup again.

SeverinAlexB (approved):

PR #86, opened 2025-03-17
Feat(client): getHomeserver

SeverinAlexB:
Adds the Pubky client method `getHomeserver`. Reads the pkarr packet and returns the homeserver id.
Throws an error in case the value can't be found or in case of a public key parse error.


pubky/pkg/test/auth.js
```diff
@@ -94,3 +94,37 @@ test("Auth: 3rd party signin", async (t) => {
...
+  const keypair = Keypair.random()
+  const publicKey = keypair.publicKey()
+
+  await client.signup(keypair, HOMESERVER_PUBLICKEY)
+
+  const client2 = Client.testnet();
+
+  let homeserver = await client2.getHomeserver(publicKey);
+  t.is(homeserver, "http://localhost:15412", "homeserver is correct");
+})
```
SeverinAlexB, line 130: @SHAcollision Do you know why this test fails?

pubky/src/wasm/api/auth.rs
```diff
@@ -101,6 +101,20 @@ impl Client {
...
+    #[wasm_bindgen(js_name = "getHomeserver")]
+    pub async fn get_homeserver(&self, public_key: &PublicKey) -> Result<PublicKey, JsValue> {
+        let val = self.0.get_homeserver(public_key.as_inner()).await;
+        if val.is_none() {
+            return Err(JsValue::from_str("No homeserver found"));
+        }
+        let val = val.unwrap();
+        let js_val = JsValue::from_str(val.as_str());
+        PublicKey::try_from(js_val)
+    }
```
SHAcollision, line 116: Looks good!!

SHAcollision (approved):
Hey @SeverinAlexB nice work! This PR is great. There is strong synergies with the recently merged work in #79 that could lead to even more code re-usage and simplification. Left some comments and questions.

The functionality is 100% ready and to me it's good to go.

pubky/src/native/internal/pkarr.rs
```diff
@@ -110,6 +110,32 @@ impl Client {
...
+
+        if pubky_records.is_empty() {
+            return None;
+        }
+
+        let record = pubky_records.first().unwrap();
+        if let RData::HTTPS(svc) = record {
+            return Some(svc.target.to_string());
+        }
+        None
```
SHAcollision, line 137: NIT: just a comment here since I already you stance about `match` . I don't even think this is a matter of personal choice or style. This is just not how `rust` is intended to be written. The `match` statement is a cornerstone of safe rust.
```suggestion
        match record {
            RData::HTTPS(svc) => Some(svc.target.to_string()),
            _ => None,
        }
```
In fact, the `is_empty()` and then the `.unwrap()` above are also very far from idiomatic rust
```rust
if pubky_records.is_empty() {
            return None;
        }

        let record = pubky_records.first().unwrap();
```
it all could just be this straight forward to understand and review statement (I only need to look at syntax to know is complete and correct, no even think about un-handled cases).
```rust
pubky_records.first().and_then(|record| match record {
    RData::HTTPS(svc) => Some(svc.target.to_string()),
    _ => None,
})
```
I guess we clash here, but I think I stand with the consensus. Certainly my head has to work 2 times has hard when I read the original code. We can ask other in-house crabs :crab:  to chime in for opinion @tipogi

tipogi: I guess, since `RData` is an enum, using a `match` clause is the way to go for readability and maintainability. However, in this case, where it only handles a specific pattern, I find `if let` to be an acceptable choice here. Still, it does feel a bit odd.

Talking about top snippet (agree with SHA), `.first()` already returns `None` if the vector is empty, so the first check feels unnecessary.

pubky/src/native/internal/pkarr.rs
```diff
@@ -110,6 +110,32 @@ impl Client {
                 _ => None,
             })
     }
+
+    /// Get the homeserver for a given Pubky public key.
+    /// Looks up the pkarr packet for the given public key and returns the content of the first `_pubky` SVCB record.
+    pub async fn get_homeserver(&self, pubky: &PublicKey) -> Option<String> {
+        let packet = self.pkarr.resolve_most_recent(pubky).await?;
```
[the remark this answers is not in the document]
SeverinAlexB: In most cases `resolve()` works. There is an edge case though when you have old and new packets mixed on the DHT `resolve()` returns the old one and you override the new homeserver with the old one.

Be aware, resolve and resolve_most_recent doesn't just ask the relay but also the DHT nodes. So I strongly suggest using `resolve_most_recent()`.

pubky/src/native/internal/pkarr.rs
```diff
@@ -136,4 +162,27 @@ mod tests {
         assert_eq!(extracted_host.as_deref(), Some(host));
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_get_homeserver() {
+        let dht = mainline::Testnet::new(3).unwrap();
```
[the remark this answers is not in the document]
SeverinAlexB: Using `pubky_testnet` in the pubky client is dangerous because you run into a cycling dependency.
- pubky_testnet depends on `pubky`.
- `pubky` depends on `pubky_testnet`.

So I recommend not using `pubky_testnet` here.

pubky/src/native/internal/pkarr.rs
SeverinAlexB: The match is literally the same as if let else. There are functionally equivalent and there is zero reason to use the match here.

```rust
match record {
    RData::HTTPS(svc) => Some(svc.target.to_string()),
    _ => None,
}
```

```rust
if let RData::HTTPS(svc) = record {
    return Some(svc.target.to_string());
} else {
   None
}
```

I reused your `Self::extract_host_from_packet` method so there is nothing to do here anymore.

I agree with the `first()?` usage.

pubky/src/native/internal/pkarr.rs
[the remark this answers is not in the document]
SHAcollision: Sounds good

pubky/src/native/internal/pkarr.rs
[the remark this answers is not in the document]
SHAcollision: Yes, the cycling dependency could be an issue, yet as it is part of the same workspace the main real problem is to not be able to test internals of `pubky-client` or use `#[cfg(test)]` to define different client behaviour during tests. 

Let's either use `pubky-testnet` or not, but let's be consistent across all tests?

pubky/src/native/internal/pkarr.rs
SHAcollision: `if let`, when there is only 2 variants in the Enum and using the `else` is effectively the same as a `match`, indeed.

 I keep supporting it's not OK to avoid `match` in `rust`, specially on `Enums` .

pubky/src/native/internal/pkarr.rs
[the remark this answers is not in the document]
SeverinAlexB: I just run into a cyclying dependency because I wanted to add a feature flag to testnet to only optionally install the pubky-client 😬 Crashed the whole rust-analyzer. I will try to remove testnet and make it clean.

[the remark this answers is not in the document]
SHAcollision: > Crashed the whole rust-analyzer.

This can happen from time to time.

pubky/src/native/internal/pkarr.rs
```diff
@@ -101,7 +101,7 @@ impl Client {
 
     /// Helper to extract the current homeserver host from a signed PKarr packet.
     /// Iterates over the records with name "_pubky" and returns the first SVCB target found.
-    fn extract_host_from_packet(packet: &SignedPacket) -> Option<String> {
+    pub(crate) fn extract_host_from_packet(packet: &SignedPacket) -> Option<String> {
```
SHAcollision, line 104: Nice

pubky/src/native/api/auth.rs
```diff
@@ -4,7 +4,7 @@ use base64::{alphabet::URL_SAFE, engine::general_purpose::NO_PAD, Engine};
 use reqwest::{IntoUrl, Method, StatusCode};
 use url::Url;
 
-use pkarr::{Keypair, PublicKey};
+use pkarr::{dns::rdata::RData, Keypair, PublicKey};
```
SHAcollision, line 7: Maybe this `RData` import is a WIP left over from previous fixes

pubky/src/native/api/auth.rs
```diff
@@ -322,6 +322,13 @@ impl Client {
         )
         .await
     }
+
+    /// Get the homeserver for a given Pubky public key.
+    /// Looks up the pkarr packet for the given public key and returns the content of the first `_pubky` SVCB record.
+    pub async fn get_homeserver(&self, pubky: &PublicKey) -> Option<String> {
+        let packet = self.pkarr.resolve_most_recent(pubky).await?;
+        Self::extract_host_from_packet(&packet)
+    }
```
SHAcollision, line 331: :rocket:

pubky/src/native/internal/pkarr.rs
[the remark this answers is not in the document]
SeverinAlexB: It happened predictably when enabling the feature. This was not a time to time thing.

PR #88, opened 2025-03-19
chore: Moved e2e tests / Fixes circular dependency

SeverinAlexB:
Moved e2e tests from pubky-client to it's own e2e create. This removes the circular dependency in pubky client.


e2e/README.md
```diff
@@ -0,0 +1,17 @@
+# Pubky End2End Tests
+
+This workspace member contains Pubky End2End tests. Run them with `cargo test`.
```
SHAcollision, line 3: Using `cargo test` is sufficient for most usecases, however the main method we are using to run tests in CI/CD is `cargo nextest run` for the added performance, logging and handling of flaky tests.

Passing tests with `cargo test` does not guarantee tests will also pass on `cargo nextest` , as nextest runs tests in more than 1 CPU thread. For example, `nextest` could fail but `cargo test` still succeed if tests spin up multiple homeservers that try to bind the same port.

SHAcollision (approved):
Looks great! It's ready to merge, nothing to change IMO.

PR #90, opened 2025-03-21
chore: Renamed pubky member dir to pubky-client

SeverinAlexB:
As the title suggests. create name stays the same.


SHAcollision (approved):
Looks good!!

PR #94, opened 2025-03-25
feat(homeserver): fallback missing pubky-host header to query param

SHAcollision:
A simple fix so the `pubky-host` header, if missing, can be extracted from the URL. So in addition of this working
```bash
curl -H "pubky-host: b8pf41opxrtgwjox7a49u6d677snmutyukzdy5wdueet5sne4q8o" https://homeserver.pubky.app/pub/pubky.app/profile.json
```
this would also work
```
curl https://homeserver.pubky.app/pub/pubky.app/profile.json&pubky-host=b8pf41opxrtgwjox7a49u6d677snmutyukzdy5wdueet5sne4q8o
```
Useful for fetching data from contexts where headers cannot be set.

### DIFF
--- a/pubky-homeserver/src/core/layers/pubky_host.rs
+++ b/pubky-homeserver/src/core/layers/pubky_host.rs
@@ -6,7 +6,7 @@ use pkarr::PublicKey;
 use std::{convert::Infallible, task::Poll};
 use tower::{Layer, Service};
 
-/// A Tower Layer to handle authorization for write operations.
+/// A Tower Layer to extract and inject the PubkyHost into request extensions.
 #[derive(Debug, Clone)]
 pub struct PubkyHostLayer;
 
@@ -18,7 +18,7 @@ impl<S> Layer<S> for PubkyHostLayer {
     }
 }
 
-/// Middleware that performs authorization checks for write operations.
+/// Middleware that extracts the public key from headers or query parameters.
 #[derive(Debug, Clone)]
 pub struct PubkyHostLayerMiddleware<S> {
     inner: S,
@@ -28,8 +28,8 @@ impl<S> Service<Request<Body>> for PubkyHostLayerMiddleware<S>
 where
     S: Service<Request<Body>, Response = axum::response::Response, Error = Infallible>
         + Send
-        + 'static
-        + Clone,
+        + Clone
+        + 'static,
     S::Future: Send + 'static,
 {
     type Response = S::Response;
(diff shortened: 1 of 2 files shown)


pubky-homeserver/src/core/layers/pubky_host.rs
```diff
@@ -40,25 +37,44 @@ where
...
+    // Fallback: check query string for "pubky-host"
+    req.uri().query().and_then(|query| {
+        query.split('&').find_map(|pair| {
+            let mut parts = pair.splitn(2, '=');
+            if let (Some(key), Some(value)) = (parts.next(), parts.next()) {
+                if key == "pubky-host" {
+                    return PublicKey::try_from(value).ok();
+                }
+            }
+            None
```
SeverinAlexB, line 77: Isn't there a way to automatically parse the query params. Do you need to parse it yourself? Not a blocker for the merge but curious

SeverinAlexB (approved):

pubky-homeserver/src/core/layers/pubky_host.rs
```diff
@@ -40,25 +37,44 @@ where
...
+        }
+    }
 
-            inner.call(req).await.map_err(|_| unreachable!())
+    // Fallback: check query string for "pubky-host"
+    req.uri().query().and_then(|query| {
+        query.split('&').find_map(|pair| {
+            let mut parts = pair.splitn(2, '=');
+            if let (Some(key), Some(value)) = (parts.next(), parts.next()) {
+                if key == "pubky-host" {
```
dzdidi, line 73:
```suggestion
                if (key == "pubky-host" || key == "host") {
```
Simply by analogy with this one https://github.com/pubky/pubky-core/pull/94/files#diff-9fd5ac0adee2ef08e643ab5ed36387677497e1c670bad561673564fa5c208363R58

dzdidi (approved):
One thing I'm not sure about. Nonetheless LGTM

pubky-homeserver/src/core/layers/pubky_host.rs
SHAcollision: Are you sure? I think in query param we should settle for one. I believe the original intention of the header check is to fallback to `host` (because browsers always attach this header) if the client user does not set a `pubky-host` header. I think this is makes sense because `host` will always be corrct and convertible to a `PublicKey` if the user is visiting himself on the browser (`https://{pubky_id}` or `pubky://{pubky_id}`). But when it comes to query params, this is an intentional decision on how you construct the URL and you either pick the convention to use `pubky-host` or `host` but having both does not add any benefit.

dzdidi: I did not mean that I need it, just thought that we might need make behavior identical. Thanks for clarifying, makes sense

SHAcollision:
Added one more client `assert` to confirm the new query param works as expected when fetching data from regular web. Merging!

PR #95, opened 2025-03-26
fix(homeserver): simplified config with serde(default)

SeverinAlexB:
Remove `unwrap()` in SignupMode and uses more serde(default) macros to simplify the code.

`Display` is still there for debug/optic purposes.

### DIFF
--- a/pubky-homeserver/src/data_directory/signup_mode.rs
+++ b/pubky-homeserver/src/data_directory/signup_mode.rs
@@ -1,10 +1,10 @@
 use core::fmt;
-use std::{fmt::Display, str::FromStr};
-
 use serde::{Deserialize, Serialize};
+use std::fmt::Display;
 
 /// The mode of signup.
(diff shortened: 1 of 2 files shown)


pubky-homeserver/src/data_directory/signup_mode.rs
```diff
@@ -1,10 +1,10 @@
 use core::fmt;
-use std::{fmt::Display, str::FromStr};
-
 use serde::{Deserialize, Serialize};
+use std::fmt::Display;
```
SHAcollision, line 3: Maybe `Debug` is the right one for what we need
```
#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize, Debug)]
```

SeverinAlexB: Guess Debug is good enough here.

PR #104, opened 2025-04-18
feat: add --homeserver-config arg to testnet binary

SeverinAlexB:
Adds the ability to run the `StaticTestnet` binary with a custom homeserver config

```bash
cargo run -- --homeserver-config=config.toml  
```

The old way still works too

```bash
cargo run
```

In both ways, the data is not persisted after the binary is stopped.

### DIFF
--- a/pubky-testnet/src/ephemeral_testnet.rs
+++ b/pubky-testnet/src/ephemeral_testnet.rs
@@ -10,35 +10,35 @@ use crate::Testnet;
 /// - An admin server for the homeserver.
 pub struct EphemeralTestnet {
     /// Inner flexible testnet.
-    pub flexible_testnet: Testnet,
+    pub testnet: Testnet,
 }
 
 impl EphemeralTestnet {
     /// Run a new simple testnet.
     pub async fn start() -> anyhow::Result<Self> {
         let mut me = Self {
-            flexible_testnet: Testnet::new().await?,
+            testnet: Testnet::new().await?,
         };
 
-        me.flexible_testnet.create_http_relay().await?;
-        me.flexible_testnet.create_homeserver_suite().await?;
+        me.testnet.create_http_relay().await?;
+        me.testnet.create_homeserver_suite().await?;
 
         Ok(me)
     }
 
     /// Create a new pubky client builder.
     pub fn pubky_client_builder(&self) -> pubky::ClientBuilder {
-        self.flexible_testnet.pubky_client_builder()
+        self.testnet.pubky_client_builder()
     }
 
     /// Create a new pkarr client builder.
     pub fn pkarr_client_builder(&self) -> pkarr::ClientBuilder {
-        self.flexible_testnet.pkarr_client_builder()
+        self.testnet.pkarr_client_builder()
     }
 
     /// Get the homeserver in the testnet.
     pub fn homeserver_suite(&self) -> &pubky_homeserver::HomeserverSuite {
-        self.flexible_testnet
+        self.testnet
             .homeservers
             .first()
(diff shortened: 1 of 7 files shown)


pubky-testnet/src/main.rs
```diff
@@ -1,16 +1,33 @@
...
     tracing_subscriber::fmt()
         .with_env_filter(
             "pubky_homeserver=debug,http_relay=debug,pkarr_relay=info,tower_http=debug,pubky_testnet=debug"
                 .to_string(),
         )
         .init();
 
-    let testnet = StaticTestnet::start().await?;
+    let testnet = if let Some(config_path) = args.homeserver_config {
+        StaticTestnet::start_with_homeserver_config(config_path).await?
```
SHAcollision, line 26: I am a little bit confused about starting a testnet Homeserver with a full blown `homeserver.config.toml` .
Are all of the fields used always? Are some ignored? If we have a  `config.toml` that looks just like this:
```toml
[general]
signup_mode = "token_required"
```
Does it use every other param from `ConfigToml::test()` ?

If all arguments are given on the `.toml` will it replace for the hardcoded silently these `pubky_listen_socket` , `icann_listen_socket` , `listen_socket` ,`public_ip` , `public_pubky_tls_port`, `public_icann_http_port` and `dht_bootstrap_nodes` ?

pubky-testnet/src/ephemeral_testnet.rs
```diff
@@ -10,43 +10,43 @@ use crate::Testnet;
 /// - An admin server for the homeserver.
 pub struct EphemeralTestnet {
     /// Inner flexible testnet.
-    pub flexible_testnet: Testnet,
```
SHAcollision, line 13: Good catch of old naming

SHAcollision (commented):
Just added a question regarding configuring using `config.toml`

pubky-testnet/src/main.rs
SeverinAlexB: The StaticTestnet takes the provided config and overrides the following properties to make the config work as expected:
https://github.com/pubky/pubky-core/blob/cb8de7e02d70d8592f73dbf6016c213c8c21b6c7/pubky-testnet/src/static_testnet.rs#L172

Everything else is taken from the config file. Happy to talk about a better way to do it if you have proposals.

SHAcollision: > Everything else is taken from the config file.

Are all fields in the config.toml mandatory then?

If a user wants to launch a testnet with a homeserver that simply requires signup tokens, do we need to provide an `example.config.toml` or can a user figure out what he needs to configure, what will be ignored, and what can be skipped?

It feels there is not much left at the moment that a user could change in the `config.toml` for a testnet homeserver: `signup_mode` , `dht_request_timeout_ms` , `lmdb_backup_interval_s` and `user_keys_republisher_interval` .

SeverinAlexB: All fields in the config are optional. With this [PR](https://github.com/pubky/pubky-core/pull/106) even the headers will be optional. Therefore, an empty file is a valid config.

>If a user wants to launch a testnet with a homeserver that simply requires signup tokens, do we need to provide an example.config.toml to ease that use case or can we reasonably expect that the user will figure out what he needs to configure, what will be ignored, and what he can skip in config.toml?

The user simply needs to add this config:
```toml
[general]
signup_mode = "token"
```

I think this is simple enough, isn't it?

SHAcollision: Understood, all clear! Ready to go for me.

SHAcollision (approved):

PR #105, opened 2025-04-21
fix: trim secret when reading

SeverinAlexB:
While moving the secret from the toml file to the new `secret` file, editors add a newline at the end. This trips the homeserver as it can't decode the hex + newline. Same happened in pkdns.

This PR trims the secret aka removes newlines and whitespaces. This ensures the secret is read even though the user added a newline.


SHAcollision (approved):
ACK!

SeverinAlexB:
Run into this problem myself in https://github.com/pubky/pubky-stack/pull/64

PR #106, opened 2025-04-21
feat: Make config headers optional

SeverinAlexB:
All config values are optional but the headers (`[general]`) aren't. This PR makes them optional too. This way, an empty config file is a valid config too.


SHAcollision (approved):
ACK, looks good!

PR #108, opened 2025-04-21
feat(homeserver): add user disk quota

SHAcollision:
Add user disk limits.

WIP


pubky-homeserver/src/core/routes/tenants/write.rs
```diff
@@ -39,19 +67,67 @@ pub async fn put(
...
+    // stream body, counting only when we *didn’t* have a hint
+    let mut writer = state.db.write_entry(&pk, full_path)?;
+    let mut seen: u64 = 0;
     let mut stream = body.into_data_stream();
-    while let Some(next) = stream.next().await {
-        let chunk = next?;
-        entry_writer.write_all(&chunk)?;
+
+    while let Some(chunk) = stream.next().await.transpose()? {
+        if hint.is_none() {
```
SHAcollision, line 88: In other words, we are trusting here that users cannot abuse and send a `content-length` header that is "1" byte but then send 5 GB. Is this something we can take for granted? @SeverinAlexB

pubky-homeserver/src/persistence/lmdb/tables/users.rs
```diff
@@ -17,17 +19,53 @@ pub const USERS_TABLE: &str = "users";
...
+    /// Retrieves the current data usage (in bytes) for a given user.
+    /// Returns the `used_bytes` value for the specified `public_key`, or zero if no record exists.
+    pub fn get_user_data_usage(&self, pk: &PublicKey) -> anyhow::Result<u64> {
+        let rtxn = self.env.read_txn()?;
+        let usage = self
+            .tables
+            .users
+            .get(&rtxn, pk)?
+            .map(|u| u.used_bytes)
+            .unwrap_or(0);
```
SeverinAlexB, line 63: We need a migration setting all existing user `used_bytes` to `0`.

SeverinAlexB (changes requested):
I think the general approach is good. Great tests. Good use of return codes. I assume the codes match with the [WebDav standard](https://grok.com/share/bGVnYWN5_d231aede-39cc-42f3-869c-261abb34bbdd)?

Here my comment about the 1GB default: https://synonymworkspace.slack.com/archives/D06Q193NX9T/p1745300669699449?thread_ts=1745258043.287039&cid=D06Q193NX9T

`content-length` can be [faked](https://grok.com/share/bGVnYWN5_bfabb293-83e3-4cdd-96bd-0628a1ec68f3). So we can't rely on it. It's smart to validate the `content-length` though so the request fails early but it shouldn't replace the stream check.

I haven't commented on naming and so on. Let me know when you think the PR is ready for that and I give it another round.

pubky-homeserver/src/core/routes/tenants/write.rs
SeverinAlexB: I think we do need to measure the size of the body every time 😐

e2e/src/tests/public.rs
```diff
@@ -69,6 +69,69 @@ async fn put_get_delete() {
...
+    // First 600 KB → OK
+    let data: Vec<u8> = vec![0; 600_000];
+    client
+        .put(&url)
+        .body(data.clone())
+        .send()
+        .await
+        .unwrap()
+        .error_for_status()
+        .unwrap();
```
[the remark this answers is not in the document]
SHAcollision: Yes, it does. I'll add the 201 assert

pubky-homeserver/src/persistence/lmdb/tables/users.rs
SHAcollision: First real migration, right? Right now, given the deserializer defaults to 0, even if this field does not exist in current DB user table, this feat will still work and `used_bytes` will effectively be 0 for everyone until they write again and start growing it (without migration needed). Well, that's the theory. Would it be OK that all existing users start now at `0` (current beta usage so far is condoned ) or do we want a "smart migration" where we pre-compute the used quota so far for all existing users? That would be quite more work.

SHAcollision:
Re-requesting review here @SeverinAlexB Did not yet test whether the `.unwrap_or(0)` works as intended with an old database. It's on my TODO list.

pubky-homeserver/src/core/routes/tenants/write.rs
```diff
@@ -39,19 +71,65 @@ pub async fn put(
...
+    let used = state.db.get_user_data_usage(&pk)?;
 
-    let mut entry_writer = state.db.write_entry(&public_key, path.0.path())?;
+    // Upfront check when we have an exact Content‑Length
+    let hint = body.size_hint().exact();
+    if let Some(exact) = hint {
+        enforce_quota(existing, exact, used, quota)?;
+    }
 
+    // Stream body, counting only when we didn’t have a hint
```
SeverinAlexB, line 86:
```suggestion
    // Stream body
```

pubky-homeserver/src/persistence/lmdb/tables/users.rs
```diff
@@ -17,17 +19,57 @@ pub const USERS_TABLE: &str = "users";
...
+    pub fn update_data_usage(&mut self, public_key: &PublicKey, delta: i64) -> anyhow::Result<()> {
+        let mut wtxn = self.env.write_txn()?;
+        let mut user = self
+            .tables
+            .users
+            .get(&wtxn, public_key)?
+            .unwrap_or(User::default());
+
+        if delta >= 0 {
+            user.used_bytes = user.used_bytes.saturating_add(delta as u64);
```
SeverinAlexB, line 48: Just a comment: Nice use of `saturating_add`

pubky-homeserver/src/persistence/lmdb/tables/users.rs
SeverinAlexB: I tested this case and for me, loading a smaller into a extend struct didn't work. I had to make a migrations.

See [this file](https://github.com/pubky/pubky-core/pull/110/files#diff-cef70e82486aaf65911ee2a1c3a3fb19e62d5172e49e41f35098c376afb543c7) from my moderation PR if you need a template.

SeverinAlexB (changes requested):
Mostly naming concerns. Obviously the `unwrap_or(0)` needs more attention. Otherwise I think this PR is good to go.

pubky-homeserver/src/core/routes/tenants/write.rs
```diff
@@ -39,19 +71,65 @@ pub async fn put(
     path: OriginalUri,
     body: Body,
 ) -> Result<impl IntoResponse> {
-    let public_key = pubky.public_key().clone();
+    let pk = pubky.public_key().clone();
+    let full_path = path.0.path();
+    let existing = existing_len(&state, &pk, full_path)?;
+    let quota = state.user_quota_bytes;
+    let used = state.db.get_user_data_usage(&pk)?;
```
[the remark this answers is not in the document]
SHAcollision: Renamed to:
`pk` -> `public_key`
`quota` -> `quota_bytes`
`current` -> `used_bytes`
`existing` -> `existing_bytes`
`incoming` -> `incoming_bytes`

pubky-homeserver/src/core/routes/tenants/write.rs
```diff
@@ -1,36 +1,68 @@
...
+/// Bytes already stored at `path` for this user (0 if none).
+fn existing_len(state: &AppState, pk: &pkarr::PublicKey, path: &str) -> Result<u64> {
+    let rtxn = state.db.env.read_txn()?;
+    Ok(state
+        .db
+        .get_entry(&rtxn, pk, path)?
+        .map(|e| e.content_length() as u64)
+        .unwrap_or(0))
+    // read‑only txns auto‑abort on drop, so no explicit commit needed
+}
```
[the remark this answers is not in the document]
SHAcollision: done

pubky-homeserver/src/persistence/lmdb/tables/users.rs
```diff
@@ -80,6 +80,42 @@ pub enum UserQueryError {
...
+    /// Updates a user's data usage by a signed `delta` (bytes).
+    /// Increases or decreases the stored `used_bytes` count for `public_key`.
+    /// Negative results are clamped to zero to prevent underflow.
+    pub fn update_data_usage(&mut self, public_key: &PublicKey, delta: i64) -> anyhow::Result<()> {
+        let mut wtxn = self.env.write_txn()?;
+        let mut user = self
+            .tables
+            .users
+            .get(&wtxn, public_key)?
+            .unwrap_or(User::default());
```
SeverinAlexB, line 92: If the pubky does not belong to one of our users, should a new user be created? It should throw an error, no?

SeverinAlexB (approved):
One last thing. Otherwise it's good to go 🚀

pubky-homeserver/src/persistence/lmdb/tables/users.rs
SHAcollision: Yes, correct. Silently creating a brand-new User::default() on a typo’d or unknown key can mask bugs. Replacing it for.
```rust
        .ok_or_else(|| anyhow::anyhow!("no user found for public key {:?}", public_key))?;
```

PR #109, opened 2025-04-23
feat(homeserver): Admin delete entry + ban user endpoints

SeverinAlexB:
Adds the following endpoints to the admin interface:
- `Delete entry` - Deletes a specific file. Adds a delete event
- `Disable user` - Disables a user. Prevents them from sign in and to read/write any files from this specific user.
- `Enable user` - Enables a disabled user.

I had to suppress an inner urge to refactor the core web server and lmdb. So most of this code is structured as good as it gets in consideration of the surrounding environment.


SeverinAlexB:
Closing this PR in favour of #110 to avoid merge conflicts.

PR #112, opened 2025-04-25
feat(homeserver): Set default icann_domain to localhost

SeverinAlexB:
For easier local testing, I set the default `icann_domain` from previously `None` to `Some("localhost")`.

FYI @catch-21


catch-21 (approved):

SHAcollision (approved):
Looks good!

PR #114, opened 2025-04-29
feat(homeserver): nginx auth request

SeverinAlexB:
Implements a session validation and user pubkey extraction endpoint compatible with the nginx `auth_request` module to the admin server.

With this endpoint, nginx is capable of assigning requests to user pubkeys and can therefore rate limit requests based on users.

Nginx usage example:

```nginx
location /protected/ {
    auth_request http://admin-homeserver:6288/nginx_auth_request;
    # Set $user_id from the x-user-id header returned by the homeserver
    auth_request_set $user_id $upstream_http_x_user_id;
    # Use $user_id for rate limiting
    limit_req_zone $user_id zone=user_limit:10m rate=1r/s;
    limit_req zone=user_limit burst=5 nodelay;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,7 @@ dependencies = [
  "axum",
  "axum-core",
  "bytes",
+ "cookie",
  "futures-util",
  "headers",
  "http",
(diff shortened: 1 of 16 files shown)


SeverinAlexB:
Doesn't work because in nginx `auth_request` is evaluated only after the `limit_req`. Therefore one can't use any auth info for rate limiting.

Closing

PR #117, opened 2025-05-01
fix (homeserver): disable relays when republishing user keys

SeverinAlexB:
The relay will rate limit the republishing. To avoid this, we disable them.


SHAcollision (approved):
Looks good!

PR #118, opened 2025-05-05
feat(homeserver): add info endpoint

SHAcollision:
Simple admin info endpoint on `/info` that returns:

```
    num_users: u64,
    num_disabled_users: u64,
    total_disk_used_mb: f64,
    num_signup_codes: u64,
    num_unused_signup_codes: u64,
```


SeverinAlexB (approved):
Let's go

PR #120, opened 2025-05-05
fix(client): signup-publish-retry-flexible

### LINKED ISSUE #111: [client] feat: retry publish on any error while signing up after successful user creation on the homeserver.
There is a small chance the pkarr publish fails when performing `signup()` . This means the `signup_token` gets consumed by the homeserver, but the whole process fails just after a successful user creation because the publishing via dht/relay fails. Ideally this failure is recoverable so it should be re-attempted a few times before throwing a client error. Maybe at least give it a 2-3 tries before fully failing, or a 1 minute total timeout.

SHAcollision:
Fixes #111 without harcoding values

### DIFF
--- a/pubky-client/src/native/internal/pkarr.rs
+++ b/pubky-client/src/native/internal/pkarr.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use pkarr::{
     dns::rdata::{RData, SVCB},
-    errors::{PublishError, QueryError},
+    errors::QueryError,
     Keypair, SignedPacket, Timestamp,
 };
 use std::convert::TryInto;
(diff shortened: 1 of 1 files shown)


SHAcollision:
@SeverinAlexB ready to review.
I ran 4 threads of user signups to the homeserver, each signing up 10K users. Two of them with this PR retrial mechanism and 2 without it. Below are preliminary results of the number of total errors (it's still running but we have enough data). This fix totally eliminates `Signup Error: Publish query timed out with no responses` and `Signup Error: Publishing SignedPacket to Mainline failed`
```
=== Summary by group ===
Group: 3 retrials 1s backoff
  Signup Error: Signup failed: error sending request for url                     2
  Signup Error: Publish query timed out with no responses                        0
  Signup Error: Publishing SignedPacket to Mainline failed                       0
  Signup Error: Publishing SignedPacket to Mainline failed code                  0
  Successful signup                                                           5834
  Profile PUT error: error sending request for url                             151

Group: no retrial
  Signup Error: Signup failed: error sending request for url                     2
  Signup Error: Publish query timed out with no responses                       10
  Signup Error: Publishing SignedPacket to Mainline failed                      12
  Signup Error: Publishing SignedPacket to Mainline failed code                  0
  Successful signup                                                           5729
  Profile PUT error: error sending request for url                             156
```

SeverinAlexB (approved):
Code looks good.

The PR title says it's a feature. This is a fix though, no? I don't see any API changes.

The only thing I am discontent with is the fact that this is a patch for an issue that should be solved in pkarr itself and not in pubky-core. Obviously, we should still merge this PR but in the future, we should start making pkarr more reliable instead of patching issues.

SHAcollision:
> This is a fix though, no? I don't see any API changes.

Correct. It's a fix. There were API changes at first, but were reverted.

> The only thing I am discontent with is the fact that this is a patch for an issue that should be solved in pkarr itself and not in pubky-core.

Indeed. This is something we should be aiming to fix at the Pkarr level, more in depth knowledge is needed. We would also benefit A LOT if we could have a `pkarr_client.publish_reliable(... num_nodes=n)` that ensures the packet has been seeded to at least `n` nodes.

PR #121, opened 2025-05-07
feat: JWT Session Tokens

SeverinAlexB:
- Adds a long-lived JWT token as second means of session authentication to the homeserver.
- Uses a static name `auth_token` for the session cookie the hs can find the user session without the `pubky-host`.
- Compatible with the legacy session cookie.
- Refactors the authentication layers.

This PR only implements a long lived session token. It does not implement a JWT access token system. This should be done at a later point.


SeverinAlexB:
This PR doesn't support the pubky-client multi user requirement. To add a statically name cookie, breaking changes in the pubky-client are required. This should be done with a `/v2` path in the homeserver to guarantee backwards compatibility.

This PR is therefore being postponed for after BTCPrague.

PR #125, opened 2025-05-13
Fix(homeserver): session expired on browser restart

SHAcollision:
It's likely this fixes https://github.com/pubky/pubky-app/issues/1458


SeverinAlexB (approved):
LGTM

PR #127, opened 2025-05-13
Fix: Rate limit - can not extract user limit key & memory leak

SeverinAlexB:
- Fixes the rate limit layer not being able to find the user key
- Fixes memory leak
- Added e2e tests


e2e/src/tests/rate_limiting.rs
```diff
@@ -0,0 +1,132 @@
...
+
+#[tokio::test]
+async fn test_limit_upload() {
+    let mut testnet = Testnet::new().await.unwrap();
+    let client = testnet.pubky_client_builder().build().unwrap();
+
+    let mut config = ConfigToml::test();
+    config.drive.rate_limits = vec![
+        PathLimit::new(
+            GlobPattern::new("/pub/**/*"),
```
SHAcollision, line 100: NIT: the glob pattern `"/pub/**/*"` duplicates **; use `"/pub/**"` to match any-depth paths.

pubky-homeserver/src/core/layers/rate_limiter/layer.rs
```diff
@@ -259,31 +275,31 @@ where
...
-                        limit.0.path.0,
-                        limit.0.method.0,
+                        limit.limit.path.0,
+                        limit.limit.method.0,
                         e
                     );
                     return Box::pin(async move {
                         Ok(Error::new(
-                            StatusCode::BAD_REQUEST,
+                            StatusCode::INTERNAL_SERVER_ERROR,
```
SHAcollision, line 284: Now returns 500 on key extraction failure. Does this contradicts clients’ expectation of `400 Bad Request` for malformed request?

pubky-homeserver/src/core/layers/rate_limiter/layer.rs
```diff
@@ -80,24 +80,37 @@ impl<S> Layer<S> for RateLimiterLayer {
...
         let quota: Quota = path_limit.clone().into();
         let limiter = Arc::new(RateLimiter::keyed(quota));
-        Self(path_limit, limiter)
+
+        // Forget keys that are not used anymore. This is to prevent memory leaks.
+        let limiter_clone = limiter.clone();
+        tokio::spawn(async move {
+            loop {
+                tokio::time::sleep(Duration::from_secs(10)).await;
+                limiter_clone.retain_recent();
```
SHAcollision, line 98: NIT, edge case: spawning a `retain_recent()` task every 10 s per limit introduces an unbounded number of background tasks and could leak; we could use a single periodic cleanup or rely on governor’s built-in eviction.

I asked ChatGPT about governor's built-in eviction, not sure if this is accurate or if you've already deem it as insuficient:
> Governor gives you two complementary, lazy eviction mechanisms:
> 1. Idle‐timeout eviction
> When you build a keyed limiter you can call
>
> ```rust
> use governor::state::keyed::DefaultKeyedStateStore;
> use governor::{Quota, RateLimiter};
> use std::time::Duration;
>
> // allow 10 events/minute per key, but evict any key not seen for 30 minutes
> let lim = RateLimiter::keyed(DefaultKeyedStateStore::new())
>    .with_quota(Quota::per_minute(nonzero!(10)))
>    .with_idle_timeout(Duration::from_secs(30 * 60))
>    .build();
> ```
> Now, if you haven’t touched a particular key for 30 minutes, the next time you go to use it governor will notice that its last‐access timestamp is past the timeout and will drop (evict) that entry before proceeding.
>
> 2. Capacity‐based (randomized) eviction
> If you expect a very large number of distinct keys but still want an upper‐bound on memory, you can also give your store a “maximum capacity.” When the total count of live keys exceeds roughly twice that capacity, governor will (again lazily, on the next access) do a quick random‐sampling sweep of entries and evict enough of the oldest ones to get you back under the ceiling.
>
> Both of these evictions happen lazily, as part of your normal calls to .until_ready(...) or .check_key(...), so there’s no background thread. The cost is just a little extra work—checking timestamps or sampling a few entries—whenever you hit one of those triggers.
>
> **Why this matters**
> - No more memory leaks: without eviction, a “keyed” limiter on untrusted input (e.g. IP addresses) can balloon forever.
> - Amortized constant‐time: you pay only O(1) overhead per request, even for eviction, because governor never does a full scan—idle eviction checks one timestamp, and capacity eviction samples only a handful of entries.
> - Fully configurable: you choose your own idle timeout and capacity, or omit either (or both) if you don’t need that eviction policy.
> - So, governor’s “built-in eviction” is simply these two store‐side policies—idle timeouts and capacity‐triggers—wired into your limiter without any extra threads or cron jobs.

pubky-homeserver/src/data_directory/quota_config/glob_pattern.rs
```diff
@@ -86,4 +86,36 @@ mod tests {
...
+        let glob_pattern = GlobPattern::from_str("/events/").unwrap();
+        assert!(glob_pattern.is_match("/events/"));
+    }
+
+    #[test]
+    fn test_glob_pattern3() {
+        let glob_pattern = GlobPattern::from_str("/pub/**/*").unwrap();
+        assert!(glob_pattern.is_match("/pub/test.txt"));
+        assert!(glob_pattern.is_match("/pub/test/test.txt"));
+        assert!(glob_pattern.is_match("/pub/"));
```
SHAcollision, line 101: Interesting I would expect the pattern `"/pub/**"` with just `**` could match `/pub/` because `**` can be any number of segments including no segment. But when adding the `/*` I would expect it to require at least one more item and it should not match the bare `/pub/` with nothing more. I would have expected this test to fail.

pubky-homeserver/src/core/layers/rate_limiter/layer.rs
SeverinAlexB: This error should only occur when the PubkyHostLayer is not active before the RateLimitingLayer. So this is more of an programming error than a user inflicted error.

With the overhaul of the auth system, this problem will likely disappear anyway.

pubky-homeserver/src/core/layers/rate_limiter/layer.rs
SeverinAlexB: > .with_idle_timeout(Duration::from_secs(30 * 60))

ChatGPT is halluzinating. This method doesn't exist.

> spawning a retain_recent() task every 10 s per limit introduces an unbounded number of background tasks and could leak (very edge case);

`retain_recent()` is a sync method. So there is not risk of spawning infinite tasks. This background task is created once when the layer is created. So there is no reason to fear unbounded tasks.

> or rely on governor’s built-in eviction.

As far as I know governor doesn't have built-in eviction. I copied the retain_recent() code from `tower-governor` as they recommend doing this. It does make sense to clean up rate limiting keys (ips or user pubkeys) that are not used anymore.

CleanShot 2025-05-14 at 08 46 33@2x


So all in all, I think this is a good solution. I improved the code a little bit but the basic approach is still valid.

pubky-homeserver/src/data_directory/quota_config/glob_pattern.rs
SeverinAlexB: These file matchers have very interesting behaviour :)

pubky-homeserver/src/core/layers/rate_limiter/layer.rs
SHAcollision: Sounds good!

> ChatGPT is halluzinating. This method doesn't exist.

:laughing:

SHAcollision (approved):
LGTM

PR #129, opened 2025-05-14
feat(homeserver): Whitelist for rate limits

SeverinAlexB:
Adds the ability to add a whitelist to rate limits.

Examples:

```toml
[[drive.rate_limits]]
path = "/session"
method = "POST"
quota = "20r/m"
key = "ip"
whitelist = [
    "127.0.0.1"
]

[[drive.rate_limits]]
path = "/pub/**" 
method = "PUT"
quota = "1mb/s"
key = "user"
burst = 10
whitelist = [
    "8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo"
]
```

The whitelist is fully validated on ConfigToml parse. 
- It either is a valid pubkey or a IP address.
- Whitelist element must match the rate limit key type.

Includes an end2end test.

### DIFF
--- a/pubky-homeserver/src/data_directory/quota_config/path_limit.rs
+++ b/pubky-homeserver/src/data_directory/quota_config/path_limit.rs
@@ -73,22 +73,10 @@ impl std::fmt::Display for PathLimit {
             .burst
             .map(|b| format!(" burst {}", b))
             .unwrap_or("".to_string());
-        let whitelist_str = if self.whitelist.is_empty() {
-            "".to_string()
-        } else {
-            format!(
-                " whitelist: {}",
-                self.whitelist
-                    .iter()
-                    .map(|k| k.to_string())
-                    .collect::<Vec<String>>()
-                    .join(", ")
-            )
-        };
         write!(
             f,
-            "{} {}: {}{burst_str} by {}.{whitelist_str}",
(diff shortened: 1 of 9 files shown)


SeverinAlexB:
FYI @catch-21

pubky-homeserver/src/data_directory/quota_config/path_limit.rs
```diff
@@ -33,20 +46,49 @@ impl PathLimit {
...
+        } else {
+            format!(
+                " whitelist: {}",
+                self.whitelist
+                    .iter()
+                    .map(|k| k.to_string())
+                    .collect::<Vec<String>>()
+                    .join(", ")
+            )
+        };
```
SHAcollision, line 87: NIT: IMO it's acceptable to print out debug of `Vec<LimitKey>` in this case
```suggestion
let whitelist_str = format!(" whitelist: {:?}", self.whitelist);
```

Will render like `" whitelist: [Ip(127.0.0.1)]"` instead of `" whitelist: 127.0.0.1"` and empty whitelist will look like `" whitelist: []"`. It's not as pretty, but acceptable for the one liner.

SHAcollision (approved):
LGTM, let's go! :rocket:

PR #135, opened 2025-05-19
Readme docker section

evd0kim:
This is just a minor update as I look around in Pubky repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ example command how to build an image:
 docker build --build-arg TARGETARCH=x86_64 -t pubky:core .
 ```
 
-A command for running it in isolated environment with log output:
+A command for running it in an isolated environment with log output:
 
 ```bash
 docker run -it pubky:core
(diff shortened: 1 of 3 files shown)


SeverinAlexB:
As far as I can see, this requires an additional dependency `podman`. This is higher friction than just providing a docker image directly.

Happy to have a docker image but podman adds unnecessary complexity.

evd0kim:
They are almost idential and normally the difference is negligible. I could correct guide for docker-only commands.

SeverinAlexB:
Docker only sounds good

evd0kim:
Curl suggestions in Readme's were wrong and Chris told me that /admin is not in the path anymore.
I think it could be merged at this point.

SeverinAlexB (approved):
Some grammar NITS, otherwise good to go. Feeding it through AI helps

SeverinAlexB:
For the next PR, please read and comply with https://github.com/pubky/pubky-core/blob/main/CONTRIBUTORS.md

PR #138, opened 2025-05-21
feat(homeserver): logging options

evd0kim:
As @SHAcollision suggested now log level options could be provided via configuration file along with common option for example.

```bash
RUST_LOG="pubky_homeserver=info,tower_http=info" cargo run
```

Environment option has superiority over config options. 
Writing into log file could be added, too. For this reason initialization was moved into HomeserverSuite::start, after reading config file.

### DIFF
--- a/pubky-homeserver/config.sample.toml
+++ b/pubky-homeserver/config.sample.toml
@@ -107,3 +107,7 @@ dht_relay_nodes = ["https://pkarr.pubky.app", "https://pkarr.pubky.org"]
 
 # Default UDP request timeout for the DHT
 dht_request_timeout_ms = 2000
+
+[logging]
+level = "info"
+filters = ["pubky_homeserver=debug", "tower_http=debug"]
--- a/pubky-homeserver/src/data_directory/config_toml.rs
+++ b/pubky-homeserver/src/data_directory/config_toml.rs
@@ -70,6 +70,12 @@ pub struct GeneralToml {
     pub user_storage_quota_mb: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct LoggingToml {
+    pub level: Option<String>,
+    pub filters: Option<Vec<String>>,
+}
+
 /// The overall application configuration, composed of several subsections.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct ConfigToml {
@@ -81,6 +87,8 @@ pub struct ConfigToml {
     pub admin: AdminToml,
     /// Peer‐to‐peer DHT / PKDNS settings (public endpoints, bootstrap, relays).
     pub pkdns: PkdnsToml,
+    /// Logging configuration. Environment variables override config settings
+    pub logging: LoggingToml,
 }
 
 impl Default for ConfigToml {
@@ -104,6 +112,18 @@ impl Default for AdminToml {
 impl Default for PkdnsToml {
     fn default() -> Self {
         ConfigToml::default().pkdns
+    }
+}
+
+impl Default for LoggingToml {
+    fn default() -> Self {
+        Self {
+            level: Some("info".to_string()),
+            filters: Some(vec![
+                "pubky_homeserver=debug".to_string(),
+                "tower_http=debug".to_string(),
+            ]),
+        }
     }
 }
 
--- a/pubky-homeserver/src/homeserver_suite/suite.rs
+++ b/pubky-homeserver/src/homeserver_suite/suite.rs
@@ -5,6 +5,7 @@ use crate::{app_context::AppContext, data_directory::PersistentDataDir};
 use anyhow::Result;
 use pkarr::PublicKey;
 use std::path::PathBuf;
+use tracing_subscriber::EnvFilter;
 
 /// Errors that can occur when building a `HomeserverSuite`.
 #[derive(thiserror::Error, Debug)]
@@ -51,6 +52,30 @@ impl HomeserverSuite {
 
     /// Run a Homeserver
     pub async fn start(context: AppContext) -> Result<Self> {
+        let env_filter = match EnvFilter::try_from_default_env() {
+            Ok(f) => f,
+            Err(_) => {
+                // create from configuration
+                let mut filter = EnvFilter::new("");
+
+                if let Some(level) = &context.config_toml.logging.level {
+                    filter = filter.add_directive(level.parse()?);
+                };
+
+                // Add any specific filters
+                if let Some(filters) = &context.config_toml.logging.filters {
+                    for filter_str in filters {
+                        filter = filter.add_directive(filter_str.parse()?);
+                    }
+                };
+                filter
+            }
+        };
+
+        tracing_subscriber::fmt().with_env_filter(env_filter).init();
+
+        tracing::debug!("Homeserver data dir: {}", context.data_dir.path().display());
+
         let core = HomeserverCore::new(context.clone()).await?;
         let admin_server = AdminServer::start(&context).await?;
 
--- a/pubky-homeserver/src/main.rs
+++ b/pubky-homeserver/src/main.rs
@@ -3,7 +3,6 @@ use std::path::PathBuf;
 use anyhow::Result;
 use clap::Parser;
 use pubky_homeserver::HomeserverSuite;
-use tracing_subscriber::EnvFilter;
 
 fn default_config_dir_path() -> PathBuf {
     dirs::home_dir().unwrap_or_default().join(".pubky")
@@ -30,14 +29,6 @@ struct Cli {
 async fn main() -> Result<()> {
     let args = Cli::parse();
 
-    tracing_subscriber::fmt()
-        .with_env_filter(
-            EnvFilter::try_from_default_env()
-                .unwrap_or_else(|_| EnvFilter::new("pubky_homeserver=debug,tower_http=debug")),
-        )
-        .init();
-
-    tracing::debug!("Using data dir: {}", args.data_dir.display());
     let server = HomeserverSuite::start_with_persistent_data_dir_path(args.data_dir).await?;
 
     tracing::info!(


pubky-homeserver/src/data_directory/config.default.toml
```diff
@@ -19,6 +19,7 @@ icann_domain = "localhost"
 user_keys_republisher_interval = 14400 # 4 hours in seconds
 dht_relay_nodes = ["https://pkarr.pubky.app", "https://pkarr.pubky.org"]
 
+[logging]
```
SHAcollision, line 23: Do we maybe also want the default to be this?
```
level = "info"
filters = ["pubky_homeserver=debug", "tower_http=debug"]
```

evd0kim: I assigned Nones in default version mainly for simplifying tests.

evd0kim:
@SeverinAlexB 
I see test passes with 3s

`assert!(time_taken > Duration::from_secs(3), "Should at least take 6s because uploads are limited to 1kb/s and the sum of the uploads is 6kb");`

and fails with 6s.
Locally it finishes with 6s.

pubky-homeserver/Cargo.toml
```diff
@@ -55,6 +55,8 @@ dyn-clone = "1.0.19"
 reqwest = "0.12.15"
 governor = "0.10.0"
 fast-glob = "0.4.5"
+log = "0.4.25"
+env_logger = "0.11.8"
```
SeverinAlexB, line 59: What are these used for?

pubky-homeserver/config.sample.toml
```diff
@@ -107,3 +107,7 @@ dht_relay_nodes = ["https://pkarr.pubky.app", "https://pkarr.pubky.org"]
 
 # Default UDP request timeout for the DHT
 dht_request_timeout_ms = 2000
+
+[logging]
+level = "info"
+filters = ["pubky_homeserver=debug", "tower_http=debug", "http_relay=debug","pkarr_relay=info","pubky_testnet=debug"]
```
SeverinAlexB, line 113: Add a description here similar to the other values in this file.

`"http_relay=debug","pkarr_relay=info","pubky_testnet=debug"` don't make any sense.

pubky-homeserver/src/data_directory/config_toml.rs
```diff
@@ -70,6 +70,12 @@ pub struct GeneralToml {
     pub user_storage_quota_mb: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct LoggingToml {
+    pub level: Option<String>,
+    pub filters: Option<Vec<String>>,
```
SeverinAlexB, line 76: - These shouldn't be Options. In this config, both should be mandatory. You can set defaults in `config.default.toml` so devs/users don't need to supply them while testing/running the app.
- Validate the values properly. Not every string is a log level. Not every string is a filter. serde is your friend. See other types.
- What happens if both level and filters are supplied? Do they work in combination?

pubky-homeserver/src/data_directory/config.default.toml
SeverinAlexB: Defaults should be set. See my other comment.

pubky-homeserver/src/data_directory/config_toml.rs
```diff
@@ -107,6 +115,12 @@ impl Default for PkdnsToml {
     }
 }
 
+impl Default for LoggingToml {
+    fn default() -> Self {
+        ConfigToml::default().logging
+    }
+}
+
```
SeverinAlexB, line 123: Feels unnecessary. All defaults are defined in `config.default.toml`. Defining defaults here will lead to redundancy and verbose code.

pubky-homeserver/src/homeserver_suite/suite.rs
```diff
@@ -51,6 +52,38 @@ impl HomeserverSuite {
...
+
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(env_filter)
+            .try_init()
+            .map_err(|_| {
+                tracing::debug!(
+                    "Instance {} trace config will be ignored",
+                    &context.keypair.public_key()
+                )
+            });
```
SeverinAlexB, line 83: What happen if I initialize two suites in a row? Doesn't initializing the subscriber twice make the app crash. In general, tracing_subcribers shouldn't be initialized by a lib.

SeverinAlexB (changes requested):
Remove the new logs. They might help you understand the code better but are either on the wrong log level or trash the logs unnecessarily. Learn how to use the system first and how it behaves so you can add sane logs on the appropriate levels later.

pubky-homeserver/src/homeserver_suite/suite.rs
evd0kim: The second one won't be initialized.

I've tried to implement it per instance but I've found that:

* entire codebase uses tracing anyway
* log lib for example also initializes subscriber just makes it in a fancier way

pubky-homeserver/Cargo.toml
evd0kim: This should be removed. Thanks for pointing out.

pubky-homeserver/config.sample.toml
evd0kim: It was copied from other places mainly because it is simple to remove unnecessary modules than figuring out where they are.

evd0kim: > "http_relay=debug","pkarr_relay=info","pubky_testnet=debug" don't make any sense.

I've set them to the info level so by default it won't lead to verbose output.

pubky-homeserver/src/homeserver_suite/suite.rs
SeverinAlexB: Good that it doesn't crash.

> tracing_subcribers shouldn't be initialized by a lib.

This point stands though. The homeserver can be used as a lib and is not necessarily only a executable.
The subscriber should therefore be initialized in the executable part and not in the part that gets exposed as a library. Maybe the lib user doesn't want to initialize the logs at all?

I see the pickle with the current structure of the suite though. I see multiple ways to resolve this. The best one might be to set `config.logging` to be an Option. This way, only when it is set the subscriber will be initialized. If unset the subscriber will not be touched.

pubky-homeserver/config.sample.toml
SeverinAlexB: It's not about the log level. http_relay, pkarr_relay, and pubky_testnet are not part of the homeserver. They are therefore useless.

pubky-homeserver/src/data_directory/config_toml.rs
SeverinAlexB: > Filters may set more a verbose level specifically for module.

Do they work too when you set a more restrictive level?

> I've never seen RUST_LOG validating values. And certainly it is not possible for modules.

Use the EnvFilter default parser.

```rust
tracing_subscriber::filter::EnvFilter::builder()
            .parse("pubky_homeserver=debug,tower_http=debug")
```

evd0kim: > Validate the values properly. Not every string is a log level. Not every string is a filter. serde is your friend. See other types.

Generally these values are being validated later upon subscriber initialization

```rust
                if let Some(level) = &context.config_toml.logging.level {
                    filter = filter.add_directive(level.parse()?);
                };

                // Add any specific filters
                if let Some(filters) = &context.config_toml.logging.filters {
                    for filter_str in filters {
                        filter = filter.add_directive(filter_str.parse()?);
                    }
                };
```      

Should I write serialization/deserialization functions because tracing:Level and LevelFilter are not supporting traits?

SeverinAlexB: > Generally these values are being validated later upon subscriber initialization

You want to push validation as much as possible to the ConfigToml. The logging params are not the best example for this but it helps simplify the code. Failing early is good.

> Should I write serialization/deserialization functions because tracing:Level and LevelFilter are not supporting traits?

Checkout the `data_directory/domain.rs` file. It shows a good example on how to wrap a type (tracing::Level in this case) and validate/serialize/deserialize it.

pubky-homeserver/src/data_directory/config_toml.rs
SHAcollision: This is in fact the mechanism to use the `config.default.toml` values as default for a `LoggingToml` . There is no way around this.

pubky-homeserver/src/homeserver_suite/suite.rs
SHAcollision: > The homeserver can be used as a lib and is not necessarily only a executable

This is relevant as homeserver is used as a lib within `pubky-testnet` and then used on test cases of many other apps (e.g. pubky-nexus) . It's important we are not overwriting the logging filters when users of the homeserver lib run the tests.

pubky-homeserver/config.sample.toml
evd0kim: > They are therefore useless.

No they are not. Since tracing is globally scoped they may actually be useful and override previous settings.

pubky-homeserver/src/homeserver_suite/suite.rs
evd0kim: >  It's important we are not overwriting the logging filters when users of the homeserver lib run the tests.

They may be overriden in the moste reliable setup when init() onvoked in start().

Here is why: possible alternative implementation with log library for example also invokes init() just in more indirect way. when doing the same without library but with tracing "default" settings only without subscriber initialization, these settings set up race conditions with whatever settings provided in external subscriber initialization. Moreover in this situation of somehow homeserver log settings take over global external defaults its thread gets killed even if the guard gets into suite instance.

```
pub struct HomeserverSuite {
    context: AppContext,
    #[allow(dead_code)] // Keep this alive. When dropped, the homeserver will stop.
    core: HomeserverCore,
    #[allow(dead_code)] // Keep this alive. When dropped, the admin server will stop.
    admin_server: AdminServer,
    log_guard: DefaultGuard, // - for example
}
```

I see general problem that application level settings (where to write logs for application) conflicts with whole setup when library has its own config. Simplest possible option here is to make 2 configs or leave try_init() in subscriber intialized from the suite. There is a third way probably. I got this idea on the weeked in I try it rn.

pubky-homeserver/config.sample.toml
evd0kim: Because global DEBUG setup will set same option for mainline and it will lead to verbosive logs where it is not expected.  I would actually combine global default DEBUG with limiting INFO options for libraries other than homeserver.

SeverinAlexB: I am not questioning that module level style filters should be set.

> http_relay, pkarr_relay, and pubky_testnet are not part of the homeserver.

Read carefully.

evd0kim:
So here is the example output of homeserver with own tracing layer 94984cf935d16be08f596a802a05318341b6497b:

2025-05-26T21:36:18.536196Z INFO  pubky_homeserver<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Homeserver HTTP listening on http://127.0.0.1:6286
2025-05-26T21:36:18.536270Z INFO  pubky_homeserver<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Homeserver Pubky TLS listening on https://fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo
2025-05-26T21:36:18.536287Z INFO  pubky_homeserver<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Homeserver Pubky TLS listening on https://127.0.0.1:6287
2025-05-26T21:36:18.536302Z INFO  pubky_homeserver<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Admin server listening on http://127.0.0.1:6288
2025-05-26T21:36:18.536318Z INFO  pubky_homeserver<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Press Ctrl+C to stop the Homeserver
2025-05-26T21:37:18.537268Z DEBUG pubky_homeserver::core::user_keys_republisher<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Republishing user keys...
2025-05-26T21:37:31.919121Z WARN  pubky_homeserver::core::user_keys_republisher<fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo>: Republished 6 user keys within 13.4s. 5 success, 1 missing, 0 failed.

The instance of homeserver has an identifier which is the public key fnmeg9oyskzeadmypirdyzfhjk71o3e9f16uaonf7gaoo6pmekwo.

The advantages of this approach:

tracing subscriber should be intialised in main() as it was the case before adding new layer
log outputs are strictly related to homeserver module
log level does not interfere with external modules in the runtime
all optional features could be implemented: file output, more granular log level specification and arbitrary formatting of log lines. probably some other features of tracing library are possible too

Initially, I intended to provide an example with public key labels for more sophisticated cases of two homeservers running along each other but apparently I would need more time for that.

SeverinAlexB:
- I don't have anything against tracing layers but it feels like it just overcomplicates things. The whole `SuiteTraceLayer` could be replaced by a simple `pubky_homeserver=info` line, couldn't it?

- We don't need the homeserver pubkey in the logs. This clutters the logs unnecessary. The use case of running two homeservers in a single binary is very rare, mostly in tests.

- We don't need any custom formatting.

- You now have a single log level only and correct me if I am wrong, you hide all other logs (pkarr, mainline, tower_http, and so on). We still need them optionally. The approach of having a main log level and module defined levels were just fine before.

> all optional features could be implemented: file output, more granular log level specification and arbitrary formatting of log lines. probably some other features of tracing library are possible too

- Make sure to complete this simple task only and not anything more.

- `Option<LoggingToml>` would be a simple solution.

evd0kim:
@SeverinAlexB 

The main idea was also about getting homeserver closer to how tower manipulates logs which is reasonable for library.

But I get your point.

pubky-homeserver/src/homeserver_suite/suite.rs
evd0kim: >  It's important we are not overwriting the logging filters when users of the homeserver lib run the tests.

I won't tell for everybody else, but i definitely needed some more logs to understand where homeserver is.
Motivated user/developer will likely eventually put println! but there will be some friction.

As suggested I removed all additional debug and error logs I've introduced.

pubky-homeserver/config.sample.toml
evd0kim: > http_relay, pkarr_relay, and pubky_testnet 

Removed from default and sample configs.

Right now the options are

```
level = "info"
filters = ["pubky_homeserver=debug"]
```
"info" suppresses every module in the runtime other than homeserver and the target filter elevates verbosity back for homeserver specifically.

pubky-homeserver/src/data_directory/config.default.toml
evd0kim: Defaults are set. See my other comment.

pubky-homeserver/config.sample.toml
```diff
@@ -111,3 +111,12 @@ dht_relay_nodes = ["https://pkarr.pubky.app", "https://pkarr.pubky.org"]
...
 # Default UDP request timeout for the DHT
 dht_request_timeout_ms = 2000
+
+[logging]
+# Global log level - sets the default minimum level for all log messages
+# Options: "trace", "debug", "info", "warn", "error"
+# Default: "info" - shows informational messages and above
+level = "info"
+# Per module log levels. Global log level allows to suppress modules which are external to homeserver. Per module spec
+# has a priority over global log levels
```
SeverinAlexB, line 121: > Global log level allows to suppress modules which are external to homeserver.

Can you elaborate here? I have trouble understand what this means

pubky-homeserver/src/data_directory/config_toml.rs
```diff
@@ -81,6 +92,8 @@ pub struct ConfigToml {
     pub admin: AdminToml,
     /// Peer‐to‐peer DHT / PKDNS settings (public endpoints, bootstrap, relays).
     pub pkdns: PkdnsToml,
+    /// Logging configuration. Environment variables override config settings
+    pub logging: Option<LoggingToml>,
```
SeverinAlexB, line 96: Add a clear description that setting this option will initialize the tracing_subscriber which is non-common for libs to do.

pubky-homeserver/src/homeserver_suite/suite.rs
```diff
@@ -51,6 +52,34 @@ impl HomeserverSuite {
...
+                if let Some(ref config) = context.config_toml.logging {
+                    filter = filter.add_directive(config.level.to_owned().into());
+                    // Add any specific filters
+                    for filter_str in &config.filters {
+                        filter = filter.add_directive(filter_str.to_owned().into());
+                    }
+                }
+                filter
+            }
+        };
```
SeverinAlexB, line 69: Not easy to read code. Too much nesting. Refactor

pubky-homeserver/src/data_directory/log_level.rs
```diff
@@ -0,0 +1,110 @@
...
+            let _level: LevelFilter = s
+                .parse()
+                .map_err(|_| anyhow::anyhow!("invalid log level directive: {}", s))?;
+        };
+        let parsed = s
+            .parse()
+            .map_err(|_| anyhow::anyhow!("invalid log level directive: {}", s))?;
+        Ok(Self(parsed))
+    }
+}
```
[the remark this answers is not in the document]
evd0kim: Yes. No problem. It will be 2 wrapped types.

pubky-homeserver/src/main.rs
```diff
@@ -30,14 +29,6 @@ struct Cli {
 async fn main() -> Result<()> {
     let args = Cli::parse();
 
-    tracing_subscriber::fmt()
-        .with_env_filter(
-            EnvFilter::try_from_default_env()
-                .unwrap_or_else(|_| EnvFilter::new("pubky_homeserver=debug,tower_http=debug")),
```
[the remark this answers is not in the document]
evd0kim: It does not make sense.
This way Subscriber will be initialized in app runtime and homeserver_suite setup will be ignored. 
Please, decide what you want to see eventually.

pubky-homeserver/config.sample.toml
evd0kim: If you put `debug` in config, the log output will be filled with `mainline` module which is in runtime, too.
With `info` and `homeserver=debug` the default configuration allows to see only homeserver debug outputs.

pubky-homeserver/src/main.rs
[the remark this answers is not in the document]
evd0kim: Just tried it and it doesn't work. Once set default subscriber setup shadows other settings.

pubky-homeserver/src/homeserver_suite/suite.rs
```diff
@@ -51,6 +52,34 @@ impl HomeserverSuite {
...
+            .with_env_filter(env_filter)
+            .try_init()
+            .map_err(|_| {
+                tracing::debug!(
+                    "Instance {} trace config will be ignored",
+                    &context.keypair.public_key()
+                )
+            });
+
+        tracing::debug!("Homeserver data dir: {}", context.data_dir.path().display());
```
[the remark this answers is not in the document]
evd0kim: Done https://github.com/pubky/pubky-core/pull/138/commits/fe7eabdda3a7df87e5ce9c5138e7316dee426383

pubky-homeserver/src/main.rs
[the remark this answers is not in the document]
SeverinAlexB: What is the problem of adding tower_http to the `config.default.toml`?

```
[logging]
level = "info"
filters = ["pubky_homeserver=debug", "tower_http=debug"]
```

pubky-homeserver/config.sample.toml
SeverinAlexB: Just tested. mainline and pkarr_republisher are too noisy anyway even on INFO level. I suggest we set `"mainline=warn", "pkarr_republisher=warn"` as default filters.

pubky-homeserver/src/data_directory/log_level.rs
```diff
@@ -0,0 +1,164 @@
+use std::fmt::{self, Display};
+use std::str::FromStr;
+
+use serde::{Deserialize, Serialize};
+use tracing_subscriber::filter::{Directive, LevelFilter};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct LogLevel(pub LevelFilter);
```
SeverinAlexB, line 8: Isn't `LevelFilter` the wrong type here? Shouldn't it be `Level`? `LevelFilter` can also be None (see type definition).

pubky-homeserver/src/homeserver_suite/suite.rs
```diff
@@ -51,6 +52,30 @@ impl HomeserverSuite {
...
+            let _ = tracing_subscriber::fmt()
+                .with_env_filter(env_filter)
+                .try_init()
+                .map_err(|_| {
+                    tracing::debug!(
+                        "Instance {} trace config will be ignored",
+                        &context.keypair.public_key()
+                    )
+                });
+        }
```
SeverinAlexB, line 75: Setting the level doesn't work. I set my config to 

```
[logging]
level = "warn"
filters = []
```

And I still get INFO logs

evd0kim: Logging filters are not getting merged correctly into default config.

evd0kim: Merging will work in case when default config doesn't contain logging or filters are empty.
```
[logging]
level = "warn"
filters = []
```

pubky-homeserver/config.sample.toml
SeverinAlexB: If it's important then we should add it. My problem is I can't decypher what it means. What are "modules which are external to the homeserver". Why do you describe the global log level here in a description of the module level filters?

Do you see what I mean?

pubky-homeserver/src/homeserver_suite/suite.rs
evd0kim: I have no idea. From what I've seen so far I have no options via serde.

I tried this workaround and it worked. I think it should be left as is if you want merging feature. Generally, my opinion is  merging configuration is an obfuscated feature (rather than just plainly creating it and putting into workdir when it does not exist). What we see is a mere consequece of complexity that we observe first before any users appear.

SeverinAlexB (changes requested):
Still not working for me. When running the homeserver with this config, I still get debug logs.

```toml
[logging]
level = "info"
module_levels = []
```

Cargo.toml
```diff
@@ -33,3 +33,6 @@ pubky = { version = "0.5.0-rc.0", path = "pubky-client" }
 pubky-homeserver = { version = "0.5.0-rc.0", path = "pubky-homeserver" }
 pubky-testnet = { version = "0.5.0-rc.0", path = "pubky-testnet" }
 pkarr-republisher = { version = "0.5.0-rc.0", path = "pkarr-republisher" }
+
+[patch.crates-io]
+serde-toml-merge = { git = "https://github.com/evd0kim/serde-toml-merge", commit="4e4f1923b382c623e09c2efb6288c7835c6a9f4e" }
```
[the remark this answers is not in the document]
evd0kim: > Why use a patch and not replace serde-toml-merge directly in the homeserver dependencies?

I like patches. They do inform about the existence of origin right away. More patches could appear in the workspace.

I moved declaration into homeserver module since I don't see any difference especially if my PR gets into library and they will update crate.

evd0kim:
@SeverinAlexB please have a look.
I also added test for merged config. It checks if arrays were replaced in default config upon merge with externally provided (could be user config).

SeverinAlexB (changes requested):
Good, this works now. One comment about the test, otherwise we are good to go with merging.

pubky-homeserver/src/data_directory/config_toml.rs
```diff
@@ -244,5 +271,19 @@ mod tests {
         assert_eq!(parsed.general.signup_mode, SignupMode::Open);
         // Other fields that were not set (left empty) should still match the default.
         assert_eq!(parsed.admin, ConfigToml::default().admin);
+        assert_eq!(parsed.logging, ConfigToml::default().logging);
+    }
+
+    #[test]
+    fn test_merged_config() {
+        let merged = ConfigToml::from_str(TEST_CONFIG).unwrap();
```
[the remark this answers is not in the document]
evd0kim: I've started with just logging section.
I get your point I will optimize config content.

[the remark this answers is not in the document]
evd0kim: It is not just 4 lines.
Actually only 2 of them were not "important" and it follows pretty much from the source code. If you did really mean that the config could be almost empty, then I have bad news, ConfigToml does not behave like that.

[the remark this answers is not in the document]
SHAcollision: @evd0kim isn't it possible to substitute the `config.test.toml` with a few inlined configs like this ?https://github.com/pubky/pubky-core/blob/ed81c491e439ec6468046ed9cf12a22494c3af7d/pubky-homeserver/src/data_directory/config_toml.rs#L268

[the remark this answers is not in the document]
evd0kim: @SHAcollision good suggestion thanks.
I was thinking already that I missed something probably.

PR #139, opened 2025-05-21
Feat(client): Set Relays in the JS Client Config

SeverinAlexB:
This PR adds new Pubky Client config parameters including a the option to set custom relays.

Additionally
- Small refactoring of the client structure and code. No breaking changes.
- Exposes `pubky-common` in the client so dev don't need to install `pubky-common` from crates.  No breaking changes.
- Adds a timeout to the wasm test pipeline in case the testnet doesn't start within 3min (see #138).
- Adds `Tsify` and `serde` to simplify JS<>rust object conversion.

In #133, we moved a lot of Cargo.toml variable to the workspace toml. This PR undoes this again for the `pubky-client` member because if you work with vscode the [rust-analyzer really can't cope with the custom vscode settings.json](https://github.com/rust-lang/rust-analyzer/issues/11900#issuecomment-1166638234) so I had to temporarily move pubky-client out of the workspace and work on it independently. I really wish this would not be such a pain.


.github/workflows/pr-check.yml
```diff
@@ -114,10 +114,18 @@ jobs:
...
-            echo "Waiting for testnet homeserver to be ready..."
+            if [ $count -ge $timeout ]; then
+              echo "Timeout: Testnet homeserver did not start within 3 minutes."
+              exit 1
+            fi
+            echo "Waiting for testnet homeserver to be ready... ($count/$timeout)"
             sleep 1
+            count=$((count + 1))
           done
+          echo "Testnet homeserver is ready."
```
SHAcollision, line 128: Good healthcheck

pubky-client/src/lib.rs
```diff
@@ -19,20 +19,22 @@ macro_rules! cross_debug {
...
 
 #[cfg(wasm_browser)]
 pub use native::Client as NativeClient;
 #[cfg(wasm_browser)]
-pub use wasm::Client;
+pub use wasm::constructor::Client;
 
 // Re-exports
 pub use pkarr::{Keypair, PublicKey};
+pub use pubky_common;
```
SHAcollision, line 37: So, we stand like this
1. We can't unpublish `pubky-common` . Ideally we would do this and call it a day.
2. It will be rarely used by downstream apps. So far, there is no users of `pubky-common`, hence, there won't be any user of this re-export as well for the time being.
3. Re-exporting an unused dependency inflates this crate’s public API; could increased compile time and dependency graph size for every downstream build; Larger generated documentation and search index, obscuring our client API; Unnecessary coupling.

I would suggest we either 1) gate it behind a feature flag or 2) remove this re-export, given that if anyone really needs `pubky-common` it is there on crates.io anyway.

pubky-client/src/native/api/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod auth;
+#[cfg(not(wasm_browser))]
+pub mod http;
+pub mod public;
```
SHAcollision, line 4: nice

pubky-client/src/shared/sleep.rs
```diff
@@ -0,0 +1,14 @@
...
+#[cfg(wasm_browser)]
+use gloo_timers::future::sleep as inner_sleep;
+
+use std::time::Duration;
+
+/// Sleep for the given duration.
+/// Works on both native and wasm.
+pub async fn sleep(duration: Duration) {
+    inner_sleep(duration).await;
+}
```
SHAcollision, line 14: nice

pubky-client/src/wasm/api/auth.rs
```diff
@@ -4,12 +4,15 @@ use url::Url;
 
 use pubky_common::capabilities::Capabilities;
 
-use crate::wasm::wrappers::{
-    keys::{Keypair, PublicKey},
-    session::Session,
+use crate::wasm::{
+    js_result::JsResult,
```
SHAcollision, line 8: Yes, finally.

pubky-client/src/wasm/constructor.rs
```diff
@@ -0,0 +1,122 @@
+use std::{num::NonZeroU64, time::Duration};
```
SHAcollision, line 1: nice `pubky-client/src/wasm/constructor.rs`

pubky-client/src/wasm/wrappers/keys.rs
```diff
@@ -13,26 +15,18 @@ impl Keypair {
...
-        let mut bytes = [0; 32];
-        secret_key.copy_to(&mut bytes);
-
-        Ok(Self(pkarr::Keypair::from_secret_key(&bytes)))
+    pub fn from_secret_key(secret_key: Vec<u8>) -> Result<Keypair, JsValue> {
+        let secret_len = secret_key.len();
+        let secret: [u8; 32] = secret_key
+            .try_into()
+            .map_err(|_| format!("Expected secret_key to be 32 bytes, got {}", secret_len))?;
+        Ok(Self(pkarr::Keypair::from_secret_key(&secret)))
```
SHAcollision, line 23: nice

pubky-client/src/wasm/constructor.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Tsify, Serialize, Deserialize, Debug)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+#[serde(rename_all = "camelCase")]
+pub struct PubkyClientConfig {
+    /// Configuration on how to access pkarr packets on the mainline DHT.
+    pub(crate) pkarr: Option<PkarrConfig>,
+    /// The maximum age of a record in seconds.
+    /// If the user pkarr record is older than this, it will be automatically refreshed.
+    pub(crate) user_max_record_age: Option<NonZeroU64>,
+}
```
SHAcollision, line 122: NIT: let's move these two to the top of these file before the `impl Client {` maybe even higher up?

SHAcollision (commented):
Looks good!! The refactor is clean and needed. Only left some minor comments

pubky-client/src/lib.rs
SeverinAlexB: > 1. We can't unpublish pubky-common . Ideally we would do this and call it a day. 

Yes

> 2. It will be rarely used by downstream apps. So far, there is no users of pubky-common, hence, there won't be any user of this re-export as well for the time being.

`pub use pubky_common::recovery_file;` was exposed already but I don't know if anybody used this. Maybe Ring? Anyway, the file size impact of pubky_common is bigger than just pubky_common::recovery_file so your point stands.

> 3. Re-exporting an unused dependency inflates this crate’s public API; increased compile time and dependency graph size for every downstream build; larger generated documentation and search index, obscuring our client API; unnecessary coupling.

Yes

> I would suggest we either 1) gate it behind a feature flag or 2) remove this re-export, given that if anyone really needs pubky-common it will always be there on crates.io despite us not wanting that...

I removed the reexport of the module. We can always add it back in case there is a request for this in the future.

SHAcollision (approved):
LGTM! :rocket:

PR #141, opened 2025-05-22
chore: bump pkarr to 3.7.2

SHAcollision:
Bumping pkarr to 3.7.2 fixes the early termination of Dht publish/resolve when all relays return early.


SeverinAlexB:
https://github.com/pubky/pkarr/compare/main...Nuhvi:pkarr:main

SeverinAlexB (approved):

PR #147, opened 2025-06-04
chore(client): Refactoring Stage 1

evd0kim:
This PR concerns re-structuring Pubky-client with minimal (as much as possible) rewrites. 
For the reviewer's convenience I decided to split refactoring in stages but it is not determined yet how many of them I will need.
The stage ends if I see that the next step will introduce breaking changes in the whole pubky-core repository or WASM client.


evd0kim:
Follow-up to our meeting @SeverinAlexB  and @SHAcollision.

As I said, what's is done should be just enough. Current implementation generally follows OOP practices and there is nothing to add besides changing locations. However, few changes may be suggested.

I've added PubkyUrl trait basically for more cleaner implementation of reqwest::Client trait inside http.rs.

```Rust
    pub fn request<U: IntoUrl + PubkyUrl>(&self, method: Method, url: U) -> RequestBuilder {
        if url.is_pubky_url() {
            let url = url.to_homeserver_url().expect("URL should be valid");
            return self.http.request(method, url);
        } else if url.is_icann_url() {
            // TODO: remove icann_http when we can control reqwest connection
            // and or create a tls config per connection.
            return self.icann_http.request(method, url);
        }
        self.http.request(method, url)
    }
```

This type allows validation of pubky URLs where it is expected. It is quite possible that http and icann_http could be unified in the single client but I haven't looked into it yet. 

Another idea concerns refactoring session-related / authentication code and putting it along with other implementations of http requests. It may be a good idea to re-use common http practices already presenting in reqwest library. This may be a breaking change. I will take care about backward-compatibility if you like this idea.

SeverinAlexB:
What's the purpose of `PubkyUrl`?

evd0kim:
> What's the purpose of `PubkyUrl`?

Currently reqwest::Client implementation contains URL - related conditional logic

```
    pub fn request<U: IntoUrl>(&self, method: Method, url: U) -> RequestBuilder {
        let url = url.as_str();

        if url.starts_with("pubky://") {
            // Rewrite pubky:// urls to https://_pubky.
            let url = format!("https://_pubky.{}", url.split_at(8).1);

            return self.http.request(method, url);
        } else if url.starts_with("https://") && PublicKey::try_from(url).is_err() {
            // TODO: remove icann_http when we can control reqwest connection
            // and or create a tls config per connection.
            return self.icann_http.request(method, url);
        }

        self.http.request(method, url)
    }
 ```
 
because standard reqwest::URL can't be used with pubky links. It is being suggested to add PubkyUrl trait so it does validation of the provided links and provides a convenient interface for such request functions when business logic requires to identify pubky links or extract keys.

**Comment by SeverinAlexB** [comment]
Is this validation done in the Homeserver client or the pubky enabled reqwest client?

**Comment by evd0kim** [comment]
> Is this validation done in the Homeserver client or the pubky enabled reqwest client?

I am sorry, I don't follow. If you talk about code snippets I provided here they are all related to pubky_core::Client exclusively. This conditional logic allows to route request between different clients, plain http to Homeserver with URI protocol substitution, plain http to homeserver icann domain and just regular non-pubky request.

**Comment by evd0kim** [comment]
> So without understanding the whole stack, you will just copy/paste the hacks that have been done previously.

Specific "hacks" may be removed when they are localized which is done in PubkyUrl. My intent is to improve legacy code step by step. This one new type may be used in other parts of the Pubky Core by the way. 

This 

> pubky:// is pubky tls. This doesn't work in browsers though

means that Url type won't be allowing pubky uri strings which is a matter of fact. 

> I don't think the reqwest library should do protocol changes (pubky:// to https:// and so on). This breaks with the expectation on how request should behave.
> reqwest library shouldn't do a domain change either (_pubky). Homeserver client could though.

I am not sure I understood this correctly but if you mean reqwest::Client trait implementation, the Pubky client selects appropriate client in a specific way because it the reqwest library basically consists of traits.

If you do mean that you don't like whole reqwest::Client implementation done just because we have a specific URL format then yes, we probably could use one standard http/-s client with some *Resolver* that will help to map pubky urls onto standard urls via pkarr.

> The whole use of https when browser and pubky tls when native is currently very implicitly implemented. We need to change that because otherwise nobody understands how it works.

This seems to be closer to the matter. My understanding was that the whole point of interchangeable protocols was about conforming to common standards and concepts. I consider it as an advantage. 

> You will very likely need to make breaking changes here.

May be. Since the chosen approach is to preserve OOP way of doing things in Rust, I think breaking changes are not needed. It may be the case for sign in-s / up-s but for now I expect that authentication and session management could be re-engineered without changing interfaces and very likely they are already obey common ways to do things due to reliance on reqwest library.

> Can you create a new branch called v1 and point this PR to it? 

Sure.

**Comment by SeverinAlexB** [comment]
Again, what is `PubkyUrl` exactly validating/doing?

It's not doing all of this, doesn't it?
```
   pub fn request<U: IntoUrl + PubkyUrl>(&self, method: Method, url: U) -> RequestBuilder {
        if url.is_pubky_url() {
            let url = url.to_homeserver_url().expect("URL should be valid");
            return self.http.request(method, url);
        } else if url.is_icann_url() {
            // TODO: remove icann_http when we can control reqwest connection
            // and or create a tls config per connection.
            return self.icann_http.request(method, url);
        }
        self.http.request(method, url)
    }
 ```

SeverinAlexB:
What is the problem of it being sealed?

I am not against traits in genera. I just don't see why we need to complicate our api with an additional `PubkyUrl` trait.

evd0kim:
Hey, an intermediate report that sums up efforts on specification of Pubky/PKarr URLs.

1) I decided to switch back to trait-based approach. It introduces a top-level specification (available interfaces) and dependency injection into URL traits. URL allows defining clean rules and automatically enforce Pubky/PKarr compilance with RFCs.

2) Pubky/Pkarr URI/URLs are a subset of RFC. Rules are enforced only when user converts URL using `to_pkarr_url()` method:

```Rust
    fn is_pubky_uri(&self) -> bool {
        let domain = self.domain().unwrap_or("").to_string();
        if let Ok(pk) = PublicKey::try_from(domain.clone()) {
            // strict pkarr rule added on top
            self.scheme() == "pubky" && domain.ends_with(pk.to_string().as_str())
        } else {
            false
        }
    }

    fn is_icann_url(&self) -> bool {
        !self.is_pkarr_domain()
    }

    fn is_pkarr_domain(&self) -> bool {
        let domain = self.domain().unwrap_or("");
        if let Ok(pk) = PublicKey::try_from(domain) {
            // strict pkarr rule
            domain.ends_with(pk.to_string().as_str())
        } else {
            false
        }
    }
```

3) That allows cleaner reqwest::http implementation

```Rust
    pub fn request<U: IntoUrl + IntoPubkyUrl>(&self, method: Method, url: U) -> RequestBuilder {
        if url.is_icann_url() {
            // TODO: remove icann_http when we can control reqwest connection
            // and or create a tls config per connection.
            return self.icann_http.request(method, url);
        }
        let url = url.to_pkarr_url().expect("Https Pkarr URL");
        self.http.request(method, url)
    }
```    
which is the purpose of refactoring.

4) also new trait is privided with test vectors which include checking against 

* scheme rule: use pubky://
* pkarr rule: pubky:// <> [https://_pubky.<pk>](http://example.com/)

evd0kim:
I would like to bring this up here @SeverinAlexB @SHAcollision, not just in Slack. There is one rule, that is implicitly applied, the <pk> must be at the top level domain.
I suggest changing it, basically to allow any domain to be at the top level, which will enable TLS support for regular browsers. This way we could provide a better UX than in Tor which requires stand alone browser. 

Quite likely it interferes with DNS resolution in Pkarr library but I think we may decide first when setting URL spec and later implement this feature in Pkarr.

PR #145, opened 2025-05-30
feat(homeserver): Flexible Files Backend (GoogleBucket, LocalFilesystem, InMemory)

SeverinAlexB:
This PR removes the files safed in LMDB and saves them in one of the three available [OpenDal](https://github.com/apache/opendal) services:

- Local Filesystem, defaults to ~/.pubky/data/files. This is set by default if nothing else is configured.
- Google Storage Bucket. Provides unlimited storage space.
- InMemory - Used for testing

The storage is configured in the config.toml:

```toml
# This is the default, storing the data in ~/.pubky/data/files
[storage]
type = "file_system"

# Or store your data in a Google Bucket
[storage]
type = "google_bucket"
credential = "/path/to/my_service_account.json"

# In Memory is only for testing but supported in the config too
[storage]
type = "in_memory"
```

## Migration process

The migration runs in the background as soon as the server is started. I tested it with the prod database successfully putting the data to the file system (4min) and on a second run to a Google Bucket (18min).
The homeserver runs as normal during the migration process. There will be no service interruption.

## Streams

I rewrote the server so all data is streamed to the storage. This keeps the memory footprint of the server low.

## After the migration

As soon as we migrated staging & production, we can remove all LMDB file related code from the database. This will reduce the code bloat significantly.

## Additional

- Cleans up and unifies the HttpError and HttpResult structs.
- Small refactor of EntryPath/WebdavPath structs to clearly indicate if `/pub` is necessary or not.

PR #153, opened 2025-06-13
feat(homeserver): Full Admin Dav API

SeverinAlexB:
Adds a full webdav api to the admin server at http://127.0.0.1:6288/dav/.
The server is protected by basic authentication with the username `admin` and the password set in the config.toml.
The API can be used with any webdav client like [Cyberduck](https://cyberduck.io/) including modern browsers.

Browser Example
CleanShot 2025-06-13 at 11 04 26@2x

The browser will automatically pop up an authentication dialog when navigating to http://127.0.0.1:6288/dav/

## Webdav Downsides

While webdav is great and convenient to use with widespread support, it does not include list pagination. Folders with a big number of files/folders will take a long time to load. Especially the root directory with all the user public keys and the directory will all the PubkyApp posts will see this issue.

## Main Homeserver API

I am not 100% sure yet if we can use this webdav library for the main homeserver API. We have a custom pagination implementation in the main api that is important for the nexus sync. I am not saying it's not possible but there will be lots of tinkering involved and we might see breaking changes.

This PR is based on #145 which needs to be merged first.

TODO: User quota update on upload/download.
Edit: Disabled PUT/DELETE because we don't update LMDB with this WebDav API (yet?)


SHAcollision:
> We have a custom pagination implementation in the main api that is important for the nexus sync.

The `.list()` method of the client is not used. However, Nexus and the moderation service do use the `/events` endpoint of the homeserver, that lists the actions PUT/DEL that homeserver tenants have performed chronologically.

I believe no application at the moment uses `.list()` at the moment (maybe the hackweek ones), though we have always envisioned listing w/ pagination being an important feature to develop lightweight apps that do not depend on an indexer.

PR #166, opened 2025-06-25
chore(homeserver): include serde_toml_merge from git to local code

SeverinAlexB:
We can't publish a crate with a git dependency on crates.io. This PR removes this dependency from #138 and adds the code to the homeserver project.


SHAcollision (approved):
Looks good!

PR #167, opened 2025-06-26
fix(hs): Return 401 instead of a 500 when the signup token is invalid

SeverinAlexB:
The homeserver returns a HTTP `500` Internal Server Error when the signup token is invalid but required.

This PR turns it into a `401` Unauthorized + improved test.

FYI @catch-21


SeverinAlexB:
FYI @MiguelMedeiros @flaviomoceri. The http response code will change. Not sure if you need to change anything on the fronted.

SHAcollision (approved):
LGTM :rocket:

PR #169, opened 2025-06-27
chore(hs): Improved session auth failed reasons and logs

SeverinAlexB:
We get random logouts with a 404 Not Found error when calling /session. This PR improves the http response message in this case + adds logs.

I had to fix the rust version for the pipeline to 1.87. The new rust 1.88 adds new clippy warnings that I didn't want to fix in this PR.

FYI: @catch-21

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -14,7 +14,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.87
           components: rustfmt
           override: true
       - name: Check formatting
@@ -28,7 +28,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.87
           components: clippy
           override: true
       - name: Cache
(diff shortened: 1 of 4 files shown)


pubky-homeserver/src/core/layers/authz.rs
```diff
@@ -121,30 +122,61 @@ fn authorize(
...
+            tracing::warn!(
+                "No session found in the database for session secret: {}, pubky: {}",
+                session_secret,
+                public_key
+            );
+            return Err(HttpError::unauthorized_with_message(
+                "No session found for session secret",
+            ));
+        }
+    };
```
SHAcollision, line 150: Looks good.

.github/workflows/pr-check.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.87
```
SHAcollision, line 17: Happens every ~6 weeks :)

SeverinAlexB: Yeahh. We need a better solution for this. It's really annoying when the pipeline just suddenly shows you new clippy errors you have never seen.

PR #174, opened 2025-07-02
fix(hs): Cookie Samesite Strict

SeverinAlexB:
Until now, the cookie was sent to any location with SameSite::None. This is insecure. The cookie should only be sent to the homeserver itself.

This PR fixes this.

This might also resolve issues related parallel usage of `beta.pubky.app` and `pubky.app`.

FYI: @catch-21


SHAcollision (approved):
Looks good.

PR #179, opened 2025-07-11
fix(hs): SVCB priority

SeverinAlexB:
Turns out that SVCB priority of `0` has a [special meaning](https://kb.isc.org/docs/svcb-and-https-resource-records-what-are-they) similar to a cname. This PR changes the priority to `1` to comply with the standard.


SHAcollision (approved):
Looks good

PR #181, opened 2025-07-12
refactor: separate native wasm into different crates

SHAcollision:
In order to more profoundly refactor the `pubky-client` into a "transport" struct (aka "PubkyClient") and an Agent struct that encapsulates the methods, keypair and session of a single user, we have to do a ton of mental work due to our `/native` / `/wasm` crate configuration. It is obvious that the high level of interlacing between the rust native crate and the JS binding makes maintainability tough.

In this PR I attempt to extract the `wasm` into its own crate `pubky-client/bingins/js`. Therefore getting rid of `/native` and `/wasm` clients in the main `pubky-client` crate.

- While the 2 crates are now separated, there is still a lot of add-hoc `#[cfg(target_arch = "wasm32")]` conditional compiling on the rust crate that ideally can be simplified away. I'll keep trying to extract as much logic as possible into the `pubky-wasm` crate.
- ~The "basic fetch" test is not working on browser. Need to spend further time understanding what exactly I broke during this refactor.~ I realize now the `# basic fetch` test has never worked in browser and this is the reason our pr check workflow that runs the wasm tests only runs `node-js`. This test was not broken by this refactor.


pubky-client/bindings/js/pkg/pubky_wasm.d.ts
```diff
@@ -0,0 +1,197 @@
...
+export function createRecoveryFile(keypair: Keypair, passphrase: string): Uint8Array;
+/**
+ * Create a recovery file of the `keypair`, containing the secret key encrypted
+ * using the `passphrase`.
+ */
+export function decryptRecoveryFile(recovery_file: Uint8Array, passphrase: string): Keypair;
+/**
+ * Pkarr Config
+ */
+export interface PkarrConfig {
```
SeverinAlexB, line 17: The `pkg` folder should not be checked in to git. It should be in gitignore.

SeverinAlexB (changes requested):
I haven't double check every line of code but I assume you mostly moved things around and not changed any logic. Looks all good to me except for the pkg dir.

After that, we can merge IMO

pubky-client/bindings/js/pkg/pubky_wasm.d.ts
SeverinAlexB: You are correct

SeverinAlexB (approved):

PR #188, opened 2025-07-23
fix(js-client): optional client config fields on ts interface

### LINKED ISSUE #144: [JS-Client] PubkyClientConfig ts types don't allow undefined.
In typescript, PubkyClientConfig is defined like this

```ts
export interface PubkyClientConfig {
    /**
     * Configuration on how to access pkarr packets on the mainline DHT.
     */
    pkarr: PkarrConfig | null;
    /**
     * The maximum age of a record in seconds.
     * If the user pkarr record is older than this, it will be automatically refreshed.
     */
    userMaxRecordAge: NonZeroU64 | null;
}
```

The optional values are nullable. The problem with this approach is that you have to set the options specifically to null instead of just omitting them.

Example:

```ts
// Current state
  const client = new Client({
    pkarr: null,
    userMaxRecordAge: 1000
  });
```

```ts
// Desired state
  const client = new Client({
    userMaxRecordAge: 1000
  });
```

This needs to be fixed.

SHAcollision:
Fixes #144

Client config interface fields can now be left `undefined` 
```ts
/**
 * Pubky Client Config
 */
export interface PubkyClientConfig {
    /**
     * Configuration on how to access pkarr packets on the mainline DHT.
     */
    pkarr?: PkarrConfig;
    /**
     * The maximum age of a record in seconds.
     * If the user pkarr record is older than this, it will be automatically refreshed.
     */
    userMaxRecordAge?: NonZeroU64;
}
```

```ts
/**
 * Pkarr Config
 */
export interface PkarrConfig {
    /**
     * The list of relays to access the DHT with.
     */
    relays?: string[];
    /**
     * The timeout for DHT requests in milliseconds.
     * Default is 2000ms.
     */
    requestTimeout?: NonZeroU64;
}
```

Thank you @SeverinAlexB for figuring this out. Possibly useful as well on our `pubky-app-specs` wasm bindings.

### DIFF
--- a/pubky-client/bindings/js/src/constructor.rs
+++ b/pubky-client/bindings/js/src/constructor.rs
@@ -18,9 +18,11 @@ static TESTNET_RELAYS: [&str; 1] = ["http://localhost:15411/"];
 #[serde(rename_all = "camelCase")]
 pub struct PkarrConfig {
     /// The list of relays to access the DHT with.
+    #[tsify(optional)]
     pub(crate) relays: Option<Vec<String>>,
     /// The timeout for DHT requests in milliseconds.
     /// Default is 2000ms.
+    #[tsify(optional)]
     pub(crate) request_timeout: Option<NonZeroU64>,
 }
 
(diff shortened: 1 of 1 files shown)


SeverinAlexB (approved):
🎉

PR #190, opened 2025-07-25
Consolidate, bump `pkarr` version

ok300:
This PR consolidates the `pkarr` version used across the member crates. It defines it just once, in the workspace root, then re-uses that version.

---

`default-features` was set to `false` in the root `Cargo.toml` because cargo doesn't allow member crates to override that setting.

Consolidating the version would otherwise result in a compilation error

```
Caused by:
  error inheriting `pkarr` from workspace root manifest's `workspace.dependencies.pkarr`

Caused by:
  `default-features = false` cannot override workspace's `default-features`
```

because `pubky-client/bindings/js/Cargo.toml` was trying to set it to `false` when the flag is implied to be `true` in the workspace root.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2597,9 +2597,9 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "pkarr"
-version = "3.8.0"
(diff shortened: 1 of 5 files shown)


SHAcollision (approved):
LGTM asking @SeverinAlexB for review as well.

PR #194, opened 2025-07-28
chore(deps): bump opendal to 0.54

SHAcollision:
Attempting to fix  https://github.com/pubky/pubky-nexus/actions/runs/16568273872/job/46853557170?pr=513

For some reason direct dependency on `opendal` and transient dependencies on `opendal` get into conflict when we import and use `pubky-testnet` .

In this PR I bump version of both `opendal` and `dav-server-opendalfs` .

Opendal 0.54 has a breaking change that I have sorted by "simply removing it" which helped me confirm compilation is OK and test suites of downstream crates now pass https://github.com/pubky/pubky-nexus/actions/runs/16569835925/job/46858827162?pr=513

 However, "simply removing it" might have altered the behavior of the homeserver. Invoking @SeverinAlexB for a more in depth check of this `opendal` breaking change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -861,9 +861,9 @@ dependencies = [
 
 [[package]]
 name = "dav-server-opendalfs"
-version = "0.6.1"
+version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a962d50f6fbb5ef2330de6c916b9892d1951480139769ec1d84594ee9099e161"
+checksum = "5305e1e00d258d119aae1e1fbdee3a704cf66b632b4f1a6d384d65531d84c35c"
 dependencies = [
  "anyhow",
  "bytes",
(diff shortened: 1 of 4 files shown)


ok300:
Indeed, the blocking methods and inner types were removed in https://github.com/apache/opendal/pull/6216/files#diff-78e3ff3bf3003b7455646a7677752acf3a585308cc6a660db2713460f0d14630 on May 23rd (just 3 days after the v0.5.3 release, so it makes sense this change is present in v0.5.4).

It's a big PR, so it's hard to see this change in the diff. To see it, you have to navigate to `core/src/raw/layer.rs` and removing the blocking elements is the only change in that file (and in `LayeredAccess`).

Then anyone who implements `LayeredAccess` would from this point on need to remove these blocking methods and inner types too, so what this PR did by removing them makes sense to me 👍

SeverinAlexB (approved):
> However, "simply removing it" might have altered the behavior of the homeserver.

It shouldn't. This is actually a good breaking change. We never used the blocking types but I was forced to implement them anyway.

@ok300 Thx for the explanation.

PR #195, opened 2025-07-28
refactor(client): WIP decouple native/wasm client HTTP implementations

SHAcollision:
# WIP: do not review yet

The primary goal of this work is to refactor the pubky-client to fully decouple native and wasm logic. Currently both use `reqwest`-like API, however wasm must use the browser's `fetch()` API, so what we currently have is a great mess. The previous architecture was tightly coupled to reqwest and contained platform-specific logic.

This WIP PR implements a new, decoupled architecture for the native client. In short, a `pubky-client` now has an `http` struct that implements the `HttpClient` trait. We create two of these: one using `reqwest` (in the `pubky-client` crate) and another one using `web_sys` (in the `bindings/js` crate). 

### This is overall a pretty painful PR

**not to be merged into `main` but into a new branch `pubky-client-future` were we will be adding the following breaking changes, this is just the first big one.**

# Changes
- [x] Introduce `HttpClient` Trait: A new `http::HttpClient` trait defines the abstract boundary for making HTTP requests.
- [x] Create `NativeHttpClient`: A concrete `NativeHttpClient` struct implements the trait using `reqwest`, encapsulating all native-specific networking logic.
- [x] Make Client Generic: The `pubky::Client` is now generic (`Client<H: HttpClient>`) and holds an instance of the abstract `HttpClient`.
- [x] Refactor `ClientBuilder` to `ClientConfig`: The builder has been renamed to `ClientConfig` and its role has been focused to only configure platform-agnostic parts of the client, like `pkarr` settings.
- [x] Provide `NativeClient` Wrapper: A NativeClient type alias and `impl` block provide a simple, ergonomic API for native users, hiding the underlying generic complexity.
- [x] Update All API Methods: All methods in `src/api/` have been updated to use the abstract `self.http.request()` method, removing all direct dependencies on `reqwest::Response` and `RequestBuilder`.
- [x] Fix `CookieJar` Session Handling: The custom logic for handling Pkarr session cookies has been correctly integrated into the `CookieStore` trait implementation, fixing a functional regression from the refactor.
- [x] Implement WasmHttpClient: Create the WASM-specific implementation of the HttpClient trait.
- [x] Refactor pubky-wasm Bindings: Update the entire bindings/js crate to use the new architecture.
- [ ] Integration Testing: fix all integration tests on the to ensure there are no functional regressions.


SHAcollision:
Closing on favor of #219

PR #197, opened 2025-07-30
Decouple `key_republisher`

ok300:
This PR is step 1 of 2 towards making `key_republish.rs` re-usable by other services, like the Nexus API.

It does the following:
- renames `key_republish.rs` to `key_republish_generic.rs`
- de-couple the `KeyRepublisher`
  - rename `HomeserverKeyRepublisher` -> `KeyRepublisher`
  - introduce `KeyRepublisherContext`, which contains only what the republisher needs, to avoid coupling with the homeserver-specific `AppContext
  - introduce `derive_key_publisher_context`, which creates the `KeyRepublisherContext` from the `AppContext`
  - remove any reference to Homeserver from `key_republish_generic` comments
  - updated tests to remove any dependency on `AppContext`, so that they strictly refer to testing a generic key republisher

Step 2 will be to move `key_republish_generic.rs` out of the homeserver into `pkarr` and update the HS and Nexus API to reference `KeyRepublisher` from `pkarr`.


ok300:
As discussed, the logic is still being refined and fine-tuned. It shouldn't be treated as standard at this time.

PR #211, opened 2025-08-05
feat(client): structured error handling for native and WASM bindings

### LINKED ISSUE #65: pubky: better errors
Replace `anyhow::Error` with `thiserror::Error` everywhere, and add `wasm_bindgen` compatible wrappers for these errors to help JS devs.

SHAcollision:
Fixes #65 Does not break existing API, but adds to it.

- [ ] Export key bytes as `Uint8Array`: Returning `Vec<u8>` in Rust makes wasm-bindgen convert it to a normal JavaScript Array, which is slower, uses more memory, and breaks APIs needing a TypedArray. Returning `Uint8Array` avoids copying, keeps data in a contiguous buffer, and works directly with browser APIs like Web Crypto and Streams.

- [ ] Migrate from the opaque `anyhow::Error` to a structured, custom `Error` enum defined with `thiserror`. This change significantly improves the library's public API by providing typed, specific error variants.

- [ ] Bindings error handling improved provide a structured and type-safe developer experience. Instead of throwing simple strings, the WASM module now throws a consistent error object with a machine-readable name (e.g., `PkarrError`) and a descriptive message, enabling robust programmatic handling in catch blocks. IN addition, full TypeScript definitions for all possible error types are now automatically generated via `tsify`, providing users with type-safety and autocompletion.

- [ ] Use `js_sys::Array` to return a native JS Array instead of forcing wasm-bindgen to convert a Rust `Vec<String>` on every call. It eliminates an extra copy and delivers a proper JS object for JS consumers.


pubky-client/bindings/js/src/js_error.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Converts a simple string slice error message into a `JsError`.
+impl From<&str> for JsError {
+    fn from(err: &str) -> Self {
+        Self {
+            name: "Error".to_string(),
+            message: err.to_string(),
+        }
+    }
+}
```
SeverinAlexB, line 100: Should this exist? Or should they be manually converted to a JsError? The `name` property here sounds meaningless.

pubky-client/bindings/js/src/js_error.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    // Variant from `impl From<JsValue>`
+    JavaScriptError,
+
+    // Variant from `impl From<PublicKeyError>`
+    PublicKeyError,
+
+    // Variant from `impl From<&str>`
+    Error,
+}
```
SeverinAlexB, line 154: Every time I see such error enums I wonder if we can merge certain variants to simplify it. A developer who uses the client will have a hard knowing which errors to catch, will likely be overwhelmed by the numbers of errors.

For example:
- `HttpStatus` & `HttpError` sound like they can be merged.
- `InvalidUrlStructure` & `UrlError` & `UrlParseError` sound very similar.
- `Error` doesn't really have any meaning when an outside dev looks at it. Maybe I am missing something.
- `AuthError` & `AuthRequestExpired` sound very similar too.

I could continue with this list but I am waiting on your input first. Maybe I am missing the meaning of this enum.

The errors obviously can have a different description. For example:

```js
{
  name: "AuthError",
  description: "Auth request expired"
}
```

The JS and Rust approach might differ here. In Rust, you can summerize errors in enums. Think of 

```rust
enum UrlError {
  InvalidUrlStructure,
  UrlParseError
}
```

Js is more limited of course.

pubky-client/bindings/js/src/wrappers/session.rs
```diff
@@ -17,11 +17,11 @@ impl Session {
...
-            .capabilities()
-            .iter()
-            .map(|c| c.to_string())
-            .collect()
+    pub fn capabilities(&self) -> js_sys::Array {
+        let arr = js_sys::Array::new();
+        for cap in self.0.capabilities() {
+            arr.push(&wasm_bindgen::JsValue::from_str(&cap.to_string()));
+        }
+        arr
```
SeverinAlexB, line 25: Unfortunately, with `js_sys::Array`, you are loosing the ts array type. It generates `capabilities(): Array<any>;`. Is this somehow fixable?

pubky-client/src/errors.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    #[error("Cryptography error: {0}")]
+    Crypto(String),
+
+    #[error("Authentication error: {0}")]
+    Auth(String),
+
+    #[error("The provided auth request is no longer valid.")]
+    AuthRequestExpired,
+}
```
SeverinAlexB, line 39: Same here similar to the PubkyErrorName enum. Can we simplify the enum by merging certain errors?

- `VerificationFailed` & `Crypto` & `Auth` & `AuthRequestExpired`  all happen in the auth section. Can they be merged?
- Can `InvalidUrlStructure` be directly turned into a `url::ParseError`?

Maybe we can merge/siimplify others too?

SeverinAlexB (changes requested):
Very nice work. I added comments regarding error simplification and ts types.

pubky-client/bindings/js/src/wrappers/session.rs
SHAcollision: True, does not seem to be possible to keep the `String[]` . Will revert this change.

pubky-client/bindings/js/src/js_error.rs
SHAcollision: I've look for all the cases where we were using this blanked Error from a `&str` or an expect and handled them one by one. Then removed this conversion impl.

pubky-client/bindings/js/src/js_error.rs
SHAcollision: I have combined using Error composition several errors in native rust, then simplified `PubkyErrorName` to just 6 variant.s

pubky-client/src/errors.rs
SHAcollision: Used composition so we have only 6 high level `PubkyError`s while each of them keeps the full width of all variants.

SHAcollision (commented):
Thanks a lot for the review @SeverinAlexB tried to tackle those comments on the last 3 commits.

pubky-client/src/errors.rs
```diff
@@ -0,0 +1,117 @@
...
+    Pkarr(#[from] PkarrError),
+
+    #[error("URL error: {0}")]
+    Url(#[from] UrlError),
+
+    #[error("Authentication error: {0}")]
+    Authentication(#[from] AuthError),
+
+    #[error("Homeserver not found for the given public key")]
+    HomeserverNotFound,
```
SeverinAlexB, line 89: Is this possibly a DNS error too? Maybe we can turn this into a pkarr error? Maybe `PkarrError::NotFound` or so?

pubky-client/src/errors.rs
```diff
@@ -0,0 +1,117 @@
+use thiserror::Error;
+
+// --- Build-Time Error ---
+// This error can only happen during the client construction phase.
+#[derive(Debug, Error)]
+pub enum BuildError {
+    #[error("Failed to build the Pkarr client: {0}")]
+    Pkarr(#[from] pkarr::errors::BuildError),
+}
```
SeverinAlexB, line 9: Should this error be merged into the PkarrError?

pubky-client/src/api/auth.rs
```diff
@@ -197,21 +207,22 @@ impl Client {
...
-
         let encrypted_token = encrypt(&token.serialize(), &client_secret);
-
         let engine = base64::engine::GeneralPurpose::new(&URL_SAFE, NO_PAD);
 
         let mut callback_url = relay.clone();
-        let mut path_segments = callback_url.path_segments_mut().unwrap();
+        let mut path_segments = callback_url
+            .path_segments_mut()
+            .map_err(|_| UrlError::InvalidStructure("Relay URL cannot be a base".to_string()))?;
```
SeverinAlexB, line 216: This could be a `url::ParseError::RelativeUrlWithoutBase`

pubky-client/src/errors.rs
```diff
@@ -0,0 +1,117 @@
...
+
+// --- Consolidated URL Error ---
+#[derive(Debug, Error)]
+pub enum UrlError {
+    #[error("Failed to parse URL: {0}")]
+    Parse(#[from] url::ParseError),
+
+    #[error("Invalid URL structure: {0}")]
+    InvalidStructure(String),
```
SeverinAlexB, line 45: Can we avoid the `InvalidStructure` error and just use url::ParseError? Most if not all of them originate in set_host and set_port and they seem to be mostly thrown when there is a 'can-not-be-a-base` error.

SHAcollision (commented):
> I am being very picky here because I want the DX to be really good. Let me know if I go overboard.

Yes please, I also think this is key and the back and forth is helping a lot.

Thanks a lot. I think I've tackled all of them. Some I have explained further the reasoning in a comment.

pubky-client/bindings/js/src/js_error.rs
```diff
@@ -0,0 +1,177 @@
...
+/// ```
+#[derive(Tsify, Serialize, Deserialize)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+#[serde(rename_all = "camelCase")]
+pub struct PubkyError {
+    pub name: PubkyErrorName,
+    pub message: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub status_code: Option<u16>,
+}
```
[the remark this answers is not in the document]
SeverinAlexB: Make sense. Two comments:

As far as I know, tsify already brings the serde_json with it. So you don't really need to add it anymore.

I am in favour of having a `data` or `detail` property already even though we only use it for `status_code`. This will future proof the API. I am 100% sure that in the near future, we will add more error information and we will regret not already having a data/detail property.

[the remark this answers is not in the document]
SHAcollision: Implemented as `data: JsValue`

SHAcollision:
> Two small comments, otherwise it looks good to me

Nice. Pushed a fix for those two.

SeverinAlexB (approved):
LGTM

pubky-client/bindings/js/src/js_error.rs
```diff
@@ -65,14 +65,35 @@ pub struct PubkyError {
...
+// --- Constructors for Ergonomics ---
+impl PubkyJsError {
+    /// Creates a new error with a name and message.
+    pub fn new(name: PubkyErrorName, message: String) -> Self {
+        Self {
+            name,
+            message,
+            data: JsValue::UNDEFINED,
+        }
+    }
```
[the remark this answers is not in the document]
SeverinAlexB: This might even be better!

```rust
pub fn new<T: Into<String>>(name: PubkyErrorName,  message: T) {
        Self {
            name,
            message: s.into(),
            data: JsValue::UNDEFINED,
        }
}
```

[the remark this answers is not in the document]
SHAcollision: Good one. While implementing it I found not all Errors have `into()` but they all (as well as String as &str) implement Display, so I opted for this instead as the most powerful solution:
```rust
    pub fn new<T: Display>(name: PubkyErrorName, message: T) -> Self {
        Self {
            name,
            message: message.to_string(),
            data: JsValue::UNDEFINED,
        }
    }
```

Also pushed a refactor of `list()` function so the TS return type becomes `Promise<string[]>` instead of the vague `Promise<Array<any>>`

SeverinAlexB (approved):
Let go!

PR #221, opened 2025-08-25
feat: ephemeral testnet with random keypair

tipogi:
Running multiple ephemeral testnets makes it difficult to test the nexus watcher layer. We need a way to track all homeservers, and having a hardcoded homeserver keypair makes this difficult

### DIFF
--- a/pubky-testnet/src/ephemeral_testnet.rs
+++ b/pubky-testnet/src/ephemeral_testnet.rs
@@ -1,6 +1,5 @@
-use http_relay::HttpRelay;
-
 use crate::Testnet;
+use http_relay::HttpRelay;
 
 /// A simple testnet with random ports assigned for all components.
 ///
@@ -21,9 +20,25 @@ impl EphemeralTestnet {
         };
 
         me.testnet.create_http_relay().await?;
-        me.testnet.create_homeserver_suite().await?;
+        me.testnet.create_homeserver().await?;
 
         Ok(me)
+    }
+
+    /// Run a new simple testnet network with a minimal setup.
+    pub async fn start_minimal() -> anyhow::Result<Self> {
+        let mut me = Self {
+            testnet: Testnet::new().await?,
+        };
+        me.testnet.create_http_relay().await?;
+        Ok(me)
+    }
+
+    /// Create an additional homeserver with a random keypair
+    pub async fn create_random_homeserver(
+        &mut self,
+    ) -> anyhow::Result<&pubky_homeserver::HomeserverSuite> {
+        self.testnet.create_random_homeserver().await
     }
 
     /// Create a new pubky client builder.
@@ -74,5 +89,21 @@ mod test {
         {
             let _ = EphemeralTestnet::start().await.unwrap();
         }
+    }
+
+    #[tokio::test]
+    async fn test_homeserver_with_random_keypair() {
+        let mut network = EphemeralTestnet::start_minimal().await.unwrap();
+        assert!(network.testnet.homeservers.len() == 0);
+
+        let _ = network.create_random_homeserver().await.unwrap();
+        let _ = network.create_random_homeserver().await.unwrap();
+        assert!(network.testnet.homeservers.len() == 2);
+
+        // The two newly created homeservers must have distinct public keys.
+        assert_ne!(
+            network.testnet.homeservers[0].public_key(),
+            network.testnet.homeservers[1].public_key()
+        );
     }
 }
--- a/pubky-testnet/src/static_testnet.rs
+++ b/pubky-testnet/src/static_testnet.rs
@@ -66,6 +66,13 @@ impl StaticTestnet {
         Ok(testnet)
     }
 
+    /// Create an additional homeserver with a random keypair
+    pub async fn create_random_homeserver(
+        &mut self,
+    ) -> anyhow::Result<&pubky_homeserver::HomeserverSuite> {
+        self.testnet.create_random_homeserver().await
+    }
+
     /// Create a new pubky client builder.
     pub fn pubky_client_builder(&self) -> pubky::ClientBuilder {
         self.testnet.pubky_client_builder()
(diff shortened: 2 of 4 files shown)


SHAcollision:
This PR does something a bit different that expected.
- Expected usage: a single testnet network where many homeservers with random (or given) secrets exist.
- On this PR: start a new testnet with a random homeserver keypair.

This PR implies a single Nexus test will need to initiate several testnet DHTs, each with a single random homeserver, instead of a single DHT with several random homeservers. Is that really what we want? Nexus will need to use a different `pubky-client` to talk to each of those homeservers, this is radically different to the behaviour on mainnet where there is only one DHT.

SHAcollision (changes requested):
Need clarification about intent. This PR does something different than I expected in order to be useful in tests.
- Expected usage: a single testnet network where many homeservers with random keypairs exist.
- On this PR: start a new testnet with a **SINGLE** random homeserver keypair.

This PR implies a single Nexus test will need to initiate several testnet DHTs, each with a single random homeserver, instead of a single DHT with several random homeservers. Is that really what we want? Nexus will need to use a different pubky-client to talk to each of those homeservers, this is radically different to the behaviour on mainnet where there is only one DHT.

This is how I would have tackled.
Additions to `testnet.rs`
```rust
impl Testnet {
    /// Creates a homeserver suite using a freshly generated random keypair.
    /// Automatically listens on the configured ports and uses this Testnet's bootstrap nodes and relays.
    pub async fn create_homeserver_suite_random(&mut self) -> Result<&HomeserverSuite> {
        let mock_dir = MockDataDir::new(ConfigToml::test(), Some(Keypair::random()))?;
        self.create_homeserver_suite_with_mock(mock_dir).await
    }
}
```
Additions to `ephemeral_testnet.rs`
```rust
impl EphemeralTestnet {
    /// Create an additional homeserver suite with a random keypair.
    pub async fn create_homeserver_suite_random(
        &mut self,
    ) -> anyhow::Result<&pubky_homeserver::HomeserverSuite> {
        self.testnet
            .create_homeserver_suite_random()
            .await
            .map_err(Into::into)
    }
}
```
And that is all, I think? I haven't tested it. 
The usage would be something like this:
```rust
#[tokio::test]
async fn test_multiple_random_homeservers_unique_keys_ephemeral() {
    // Start EphemeralTestnet (creates one http relay + one initial homeserver).
    let mut testnet = EphemeralTestnet::start().await.unwrap();

    // Create two additional homeservers with a random keypair.
    let hs1 = testnet.create_homeserver_suite_random().await.unwrap();
    let hs2 = testnet.create_homeserver_suite_random().await.unwrap();
    // This testnet now has 3 homeservers. The default hardcoded one and two extra randoms.

    // The two newly created homeservers must have distinct public keys.
    assert_ne!(&hs1.public_key(), &hs2.public_key());
}
```

SHAcollision (changes requested):
There is a few things we can simplify for this PR. IMO the only change needed to also perform as well in `StaticTestnet`, is not been made, and the changes introduced on it, do not seem to be needed, right? Are we using `StaticTestnet` on Nexus tests at all?

pubky-testnet/src/static_testnet.rs
```diff
@@ -27,22 +27,38 @@ pub struct StaticTestnet {
...
+    /// If you want to extend the [StaticTestnet] instance with other homeservers, call
+    /// [StaticTestnet::run_fixed_homeserver] to add new ones with the desired keypair.
+    ///
+    /// ### Arguments
+    ///
+    /// - `config_path`: path to the homeserver config
+    /// - `use_default_kp`: if true, starts with the default homeserver keypair. Otherwise, uses a random keypair.
+    pub async fn start_with_options(
+        config_path: Option<PathBuf>,
+        use_default_kp: bool,
```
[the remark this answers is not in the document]
ok300: Done: 8c9fb652f4d06a3c986cab6c1be5dc6ab2072f87

ok300:
> IMO the only change needed to also perform as well in `StaticTestnet` has not been made

Not true.

The way it was done before (now reverted, as you requested) allowed for:
a) create a test suite with an initial default homeserver (keypair `[0, 32]`)
b) crate a test suite with an initial random homeserver (random keypair)
c) allow adding extra random homeservers to an existing test suite instance

The way it is done now (as per the req. comment above)
- the test suite can add a new random homeserver (which is identical to c above, so the previous code did in fact address it)
- note: it is impossible to achieve b above: start test suite with 1 single random homeserver, instead the constructor always initializes a default one internally

> Are we using `StaticTestnet` on Nexus tests at all?

No, this was added here simply as courtesy.

---

PR is ready for final review.

pubky-testnet/src/ephemeral_testnet.rs
```diff
@@ -21,11 +20,27 @@ impl EphemeralTestnet {
...
+    pub async fn start_minimal() -> anyhow::Result<Self> {
+        let mut me = Self {
+            testnet: Testnet::new().await?,
+        };
+        me.testnet.create_http_relay().await?;
+        Ok(me)
+    }
+
+    /// Create an additional homeserver suite with a random keypair
+    pub async fn create_homeserver_suite_random(
```
SHAcollision, line 38: could be renamed to `create_random_homeserver()` so the names match across testnets. Also consistent with deleting the "suite" part.

pubky-testnet/src/static_testnet.rs
```diff
@@ -66,6 +66,13 @@ impl StaticTestnet {
         Ok(testnet)
     }
 
+    /// Create an additional homeserver suite with a random keypair
+    pub async fn create_random_homeserver_suite(
```
SHAcollision, line 70: > could be renamed to `create_random_homeserver()` so the names match across testnets. Also consistent with deleting the "suite" part.

SHAcollision (approved):
Looks good, just left a NIT comment.

SHAcollision (approved):

PR #239, opened 2025-09-12
Fix Dockerfile cross-compilation

catch-21:
@86667 identified that cross-compilation was not correctly configured in the Dockerfile for ARM->x86 or x86->ARM. This PR addresses building Docker image with all native and cross compilation combinations whist retaining use of host x86 build toolchain.

If we want to also support x86 Mac and Windows host then this can be simplified with less conditions by always downloading and using x86_64-linux-musl-cross (untested).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,16 +16,30 @@ RUN apk add --no-cache \
     build-base \
     curl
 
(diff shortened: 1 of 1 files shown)


86667:
Thanks for sorting this. I think we should use something like `buildx` when we next update this.

PR #254, opened 2025-10-31
Add events-stream endpoint

86667:
Implementation of `/events-stream` endpoint. 

The requirements can be found [here](https://docs.google.com/document/d/188sZTRvfnHjJlD8Q6MVFizKCzUITdUUX6XZLbtigq18/edit?tab=t.d3ta9wcnaetd#heading=h.crxs5hzh7cpk).

## Description

The `/events-stream` endpoint provides Server-Sent Events (SSE) for real-time event streaming. It provides two modes of operation:
1) Batch mode for historical events only
2) Live mode for historical events followed by real-time updates

It supports streaming events for a batch of users simultaneously with per-user cursor tracking, optional path filtering and configurable reverse chronological ordering. 

The single endpoint supports 2 main use cases:
1) Nexus polling for the next batch of events for a list of Users
2) Backup App style syncing of all events for some group of Users with live listening for real-time Events 

## Future improvements

There are a couple things which I'll add after this initial version is deployed:

- Allow client to specify wait time in-between db reads. This is in order to reduce stream overload on slow clients.
- Performance monitoring of real usage: Table read capacity and memory usage

### DIFF
--- a/pubky-homeserver/src/client_server/routes/events.rs
+++ b/pubky-homeserver/src/client_server/routes/events.rs
@@ -22,23 +22,16 @@ use crate::{
     shared::{webdav::WebDavPath, HttpError, HttpResult},
 };
 
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum EventStreamError {
+    #[error("User not found")]
     UserNotFound,
+    #[error("{0}")]
     InvalidParameter(String),
-    DatabaseError(sqlx::Error),
+    #[error("Database error: {0}")]
+    DatabaseError(#[from] sqlx::Error),
+    #[error("Invalid public key: {0}")]
     InvalidPublicKey(String),
-}
-
-impl std::fmt::Display for EventStreamError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            EventStreamError::UserNotFound => write!(f, "User not found"),
-            EventStreamError::InvalidParameter(msg) => write!(f, "{}", msg),
-            EventStreamError::DatabaseError(e) => write!(f, "Database error: {}", e),
-            EventStreamError::InvalidPublicKey(key) => write!(f, "Invalid public key: {}", key),
-        }
-    }
 }
 
 impl From<EventStreamError> for HttpError {
@@ -48,12 +41,6 @@ impl From<EventStreamError> for HttpError {
             EventStreamError::DatabaseError(e) => HttpError::from(e),
             _ => HttpError::bad_request(error.to_string()),
         }
-    }
-}
-
-impl From<sqlx::Error> for EventStreamError {
-    fn from(error: sqlx::Error) -> Self {
-        EventStreamError::DatabaseError(error)
     }
 }
 
--- a/pubky-homeserver/src/persistence/lmdb/sql_migrator/events.rs
+++ b/pubky-homeserver/src/persistence/lmdb/sql_migrator/events.rs
@@ -32,7 +32,6 @@ pub async fn create<'a>(
     let created_at = timestamp_to_sqlx_datetime(timestamp);
     let event_type = match event {
         Event::Put(_) => EventType::Put {
-            // Legacy LMDB events don't have content_hash - use zero hash
             content_hash: Hash::from_bytes([0; 32]),
         },
         Event::Delete(_) => EventType::Delete,
--- a/pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
+++ b/pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
@@ -47,9 +47,9 @@ pub fn build_storage_operator(
             };
             let builder = opendal::services::Fs::default().root(files_dir.as_str());
             opendal::Operator::new(builder)?
+                .layer(events_layer)
                 .layer(user_quota_layer)
                 .layer(entry_layer)
-                .layer(events_layer)
                 .finish()
         }
         #[cfg(feature = "storage-gcs")]
@@ -60,9 +60,9 @@ pub fn build_storage_operator(
             );
             let builder = config.to_builder()?;
             opendal::Operator::new(builder)?
+                .layer(events_layer)
                 .layer(user_quota_layer)
                 .layer(entry_layer)
-                .layer(events_layer)
                 .finish()
         }
         #[cfg(any(feature = "storage-memory", test))]
@@ -70,9 +70,9 @@ pub fn build_storage_operator(
             tracing::info!("Store files in memory");
             let builder = opendal::services::Memory::default();
             opendal::Operator::new(builder)?
+                .layer(events_layer)
                 .layer(user_quota_layer)
                 .layer(entry_layer)
-                .layer(events_layer)
                 .finish()
         }
     };


pubky-homeserver/src/persistence/sql/entities/event.rs
```diff
@@ -105,19 +207,122 @@ impl EventRepository {
...
+            .from_subquery(combined_query, subquery_alias.clone())
+            .column((subquery_alias.clone(), EventIden::Id))
+            .column((subquery_alias.clone(), EventIden::User))
+            .column((subquery_alias.clone(), EventIden::Type))
+            .column((subquery_alias.clone(), EventIden::Path))
+            .column((subquery_alias.clone(), EventIden::CreatedAt))
+            .column((subquery_alias.clone(), EventIden::ContentHash))
+            .column((subquery_alias.clone(), UserIden::PublicKey))
+            .order_by((subquery_alias, EventIden::Id), order)
+            .limit(DEFAULT_LIST_LIMIT as u64)
```
SHAcollision: Nit: just at a glance, not sure if correct. Using here `DEFAULT_LIST_LIMIT` instead of query param limit sanitised by `DEFAULT_MAX_LIST_LIMIT` as in [L346](https://github.com/pubky/pubky-core/pull/254/files#diff-fc2328ca3be562c6ac1d1903a0ac3b4dcf10bf2424721a2c930727bdb5ae133eR346)

SeverinAlexB:
I will take over the review from @dzdidi here

pubky-homeserver/src/core/routes/feed.rs
```diff
@@ -46,3 +68,229 @@ pub async fn feed(
...
+/// event: PUT
+/// data: pubky://user_pubkey/pub/example.txt
+/// data: cursor: 42
+/// data: content_hash: af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262
+/// ```
+pub async fn feed_stream(
+    State(state): State<AppState>,
+    params: EventStreamQueryParams,
+) -> HttpResult<Sse<impl Stream<Item = Result<Event, Infallible>>>> {
+    use crate::constants::MAX_EVENT_STREAM_USERS;
```
SeverinAlexB: https://refactoring.guru/smells/long-method

86667: Ive pulled a couple bits of logic out. The `stream` macro makes it difficult to split up much further without losing readability IMO.

pubky-homeserver/src/persistence/files/events/repository.rs
```diff
@@ -105,19 +207,122 @@ impl EventRepository {
...
+            .from_subquery(combined_query, subquery_alias.clone())
+            .column((subquery_alias.clone(), EventIden::Id))
+            .column((subquery_alias.clone(), EventIden::User))
+            .column((subquery_alias.clone(), EventIden::Type))
+            .column((subquery_alias.clone(), EventIden::Path))
+            .column((subquery_alias.clone(), EventIden::CreatedAt))
+            .column((subquery_alias.clone(), EventIden::ContentHash))
+            .column((subquery_alias.clone(), UserIden::PublicKey))
+            .order_by((subquery_alias, EventIden::Id), order)
+            .limit(DEFAULT_LIST_LIMIT as u64)
```
86667: This is correct - the user provided limit is enforced by the `feed_stream()` API layer function which loops over this `get_by_user_cursors()` with a set limit of 100 items returned per call. Its an arbitrary value to balance size of db querys with speed of adding data to the stream.

pubky-homeserver/src/persistence/sql/migrations/m20251014_events_table_index_and_content_hash.rs
```diff
@@ -0,0 +1,225 @@
...
+        let query = statement.build(PostgresQueryBuilder);
+        sqlx::query(query.as_str()).execute(&mut **tx).await?;
+
+        // Add nullable content_hash column for tracking file content hashes
+        let statement = Table::alter()
+            .table(TABLE)
+            .add_column(ColumnDef::new(EventIden::ContentHash).binary().null())
+            .to_owned();
+        let query = statement.build(PostgresQueryBuilder);
+        sqlx::query(query.as_str()).execute(&mut **tx).await?;
```
SeverinAlexB: Are there any downsides that the event hashes can be null here? The developers using the pubky-sdk therefore is never sure if there is a hash available. There probably should always be a hash available to make the life of the dev using the pubky-sdk easy, no?

Should we copy the hash from the entry table?

pubky-homeserver/src/core/routes/feed.rs
```diff
@@ -2,15 +2,85 @@ use axum::{
...
+    /// data: content_hash: abc123... (optional, only if present)
+    /// ```
+    pub fn to_sse_data(&self) -> String {
+        let mut lines = vec![self.path.clone(), format!("cursor: {}", self.cursor)];
+        if let Some(hash) = &self.content_hash {
+            lines.push(format!("content_hash: {}", hash));
+        }
+        lines.join("\n")
+    }
+}
```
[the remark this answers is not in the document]
86667: Fair. Pushed.

pubky-homeserver/src/persistence/files/events/events_repository.rs
```diff
@@ -0,0 +1,527 @@
...
+                    .replace('\\', "\\\\")
+                    .replace('_', "\\_")
+                    .replace('%', "\\%");
+                let like_pattern = format!("{}%", escaped_prefix);
+                statement = statement
+                    .and_where(
+                        Expr::col((EVENT_TABLE, EventIden::Path))
+                            .like(LikeExpr::new(like_pattern).escape('\\')),
+                    )
+                    .to_owned();
```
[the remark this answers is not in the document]
86667: The first is inserting the escape characters, the second `escape()` fn is inserting into the SQL statement the instruction for which characters are the escape markers

pubky-homeserver/src/persistence/files/events/events_repository.rs
```diff
@@ -0,0 +1,520 @@
...
+            union_queries.push(statement);
+        }
+
+        // Combine all user queries with UNION ALL and wrap in subquery
+        let mut combined_query = union_queries[0].clone();
+        for query in union_queries.iter().skip(1) {
+            combined_query = combined_query
+                .union(sea_query::UnionType::All, query.clone())
+                .to_owned();
+        }
```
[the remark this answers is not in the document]
86667: A conversation with Claude revealed a few very technical reasons that the current approach is most likely most efficient. I dont want to attempt to reproduce the arguments here. 

I ran a test with 5000 events over 50 users (max users per query) and am quite happy with the results:

- 4,202 events/sec average throughput
- 0.24ms per event
- No degradation as more events accumulated

### Batch Timing Breakdown - how long each set of 500 events took:
Batch 1 (1-500):     138ms  →  3,599 events/sec
Batch 2 (501-1000):  134ms  →  3,722 events/sec  
Batch 3 (1001-1500): 113ms  →  4,414 events/sec
Batch 4 (1501-2000): 122ms  →  4,073 events/sec
...continuing with similar performance...
Batch 10 (4501-5000): 113ms  →  4,387 events/sec

### Total streaming time: 1.19 seconds for 5,000 events

Also keep in mind that i plan to add monitoring for this - https://github.com/pubky/pubky-core/blob/open_tel_init/pubky-homeserver/src/core/metrics.rs#L12C5-L12C35

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -1,16 +1,249 @@
...
+impl std::fmt::Display for EventStreamError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            EventStreamError::UserNotFound => write!(f, "User not found"),
+            EventStreamError::InvalidParameter(msg) => write!(f, "{}", msg),
+            EventStreamError::DatabaseError(e) => write!(f, "Database error: {}", e),
+            EventStreamError::InvalidPublicKey(key) => write!(f, "Invalid public key: {}", key),
+        }
+    }
+}
```
SeverinAlexB: NIT: This looks like a perfect case for [thiserror](https://docs.rs/thiserror/latest/thiserror/). Nothing you need to change now but good to know for next time.

pubky-homeserver/src/persistence/lmdb/sql_migrator/events.rs
```diff
@@ -23,34 +20,48 @@ fn pubky_url_into_entry_path(url: &str) -> anyhow::Result<EntryPath> {
...
 ) -> anyhow::Result<()> {
+    use pubky_common::crypto::Hash;
+
     let created_at = timestamp_to_sqlx_datetime(timestamp);
     let event_type = match event {
-        Event::Put(_) => EventType::Put,
+        Event::Put(_) => EventType::Put {
+            // Legacy LMDB events don't have content_hash - use zero hash
+            content_hash: Hash::from_bytes([0; 32]),
+        },
```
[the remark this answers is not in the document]
86667: This was just to satisfy the compiler's type checker. I'll remove the comment as it is indeed a little misleading

PR #264, opened 2025-11-14
fix: feed cursor empty when no file

SeverinAlexB:
Nexus shows an invalid event line. Turns out this is a bug that came with the postgres PR. A cursor was added to the response even if there is no next page. In this case, the cursor was empty like `cursor:`.

```
nexusd       | 2025-11-14T11:43:07.565563Z DEBUG nexus_watcher::service::processor:63: Polling new events from homeserver
nexusd       | 2025-11-14T11:43:07.565627Z DEBUG hyper_util::client::legacy::pool:269: reuse idle connection for ("https", 8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo)
nexusd       | 2025-11-14T11:43:07.566511Z DEBUG hyper_util::client::legacy::pool:395: pooling idle connection for ("https", 8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo)
nexusd       | 2025-11-14T11:43:07.566544Z DEBUG nexus_watcher::service::processor:85: Homeserver response lines ["cursor:"]
nexusd       | 2025-11-14T11:43:07.566551Z  INFO nexus_watcher::service::processor:46: Processing 1 event lines
nexusd       | 2025-11-14T11:43:07.566552Z DEBUG nexus_watcher::events:43: New event: cursor:
nexusd       | 2025-11-14T11:43:07.566554Z ERROR nexus_watcher::service::processor:118: InvalidEventLine: Malformed event line, cursor:
```

This PR fixes this by remove the cursor in case there is not next page.

FYI: @catch-21

### DIFF
--- a/pubky-homeserver/src/core/routes/feed.rs
+++ b/pubky-homeserver/src/core/routes/feed.rs
@@ -34,9 +34,7 @@ pub async fn feed(
         .iter()
         .map(|event| format!("{} pubky://{}", event.event_type, event.path.as_str()))
         .collect::<Vec<String>>();
-    let next_cursor = events
-        .last()
-        .map(|event| event.id.to_string());
+    let next_cursor = events.last().map(|event| event.id.to_string());
 
     if let Some(next_cursor) = next_cursor {
         result.push(format!("cursor: {}", next_cursor));


SHAcollision (dismissed):
Looks good!

I am not sure, but maybe returning no cursor breaks the old contract and Nexus might fail to parse the events still if no `cursor:` line exists at all. The response of this endpoint is still plaintext, not a Json, so the parsing is manual. I think the old behaviour was to always return the index of the last element even if there is no further pages, but I am not sure maybe @ok300 knows with more certainty?

SeverinAlexB:
Some context to how the cursor changed with the new postgres homeserver. This might need updates in Nexus in case nexus strongly validates the cursor.

Ok, looking at the code before postgres: The cursor was a [timestamp number](https://github.com/pubky/pubky-core/blob/325d8138461bd0221968413c6c321785d7cb2e06/pubky-homeserver/src/persistence/lmdb/tables/events.rs#L91).
With the new postgres change: The cursor is the [event id](https://github.com/pubky/pubky-core/blob/main/pubky-homeserver/src/core/routes/feed.rs#L39) aka regular number.
Old cursor are still supported by [this function](https://github.com/pubky/pubky-core/blob/main/pubky-homeserver/src/persistence/sql/entities/event.rs#L83). New feed calls will hand out the new format though.

So if Nexus still expects the timestamp and not just a number, then this needs changing.

ok300 (approved):
LGTM.

> I think the old behaviour was to always return the index of the last element even if there is no further pages

I checked again, this is dynamically handled in `process_event_lines`. For each line:
- if a line starts with `cursor: `, we extract the cursor and save it
- otherwise we try to parse and process an event

As an extra precaution, I opened a Nexus PR to only save cursors that are not empty: https://github.com/pubky/pubky-nexus/pull/618

PR #268, opened 2025-11-19
feat(sdk): added auth signup flow

SeverinAlexB:
- Refactors `PubkyAuthFlow`
- Adds a new deep link to signup to a new homeserver.
- Adds a new new `AuthFlowKind` argument when starting an auth flow. AuthFlowKind allows to choose between the signin and the signup flow.
- Standardizes the deeplinks. Adds parsing methods.
  - Allows `pubkyring` and the `pubkyauth` url schema for deep links. `pubkyauth` is always used for new links.
- Added example

The new QR codes format is

```
pubkyauth://signup?hs={hs_pubkey}&ic={invite_code}&relay={http_relay_url_with_channel_id}&secret={secret_base64}&caps={comma_separated_capabilities}
```

---
<s>This PR adds the PubkyAuthSignupFlow struct to the SDK. This is done in the context of "Invitecodeless Franky" and to provide a simple QR code that Pubky Ring can scan when signing up to a homeserver.

The new QR codes format is

```
pubkyring:///signup?hs={hs_pubkey}&ic={invite_code}&relay={http_relay_url_with_channel_id}&secret={secret_base64}&caps={comma_separated_capabilities}
```

The properties relay, secret, and caps are kept the same as the sign in deep link. This way, the `signer.approve_auth()` method can be reused.

See the `signup_authz` as an example.</s>

### DIFF
--- a/pubky-sdk/bindings/js/src/actors/deep_links/seed_export.rs
+++ b/pubky-sdk/bindings/js/src/actors/deep_links/seed_export.rs
@@ -16,7 +16,7 @@ impl SeedExportDeepLink {
             pubky::deep_links::SeedExportDeepLink::from_str(url).map_err(|e| {
                 PubkyError::new(
                     PubkyErrorName::InvalidInput,
-                    format!("Invalid signup deep link: {}", e),
+                    format!("Invalid seed export deep link: {}", e),
                 )
             })?,
         ))
--- a/pubky-sdk/bindings/js/pkg/README.md
+++ b/pubky-sdk/bindings/js/pkg/README.md
@@ -17,7 +17,7 @@ Module system + TS types: ESM and CommonJS both supported; TypeScript typings ge
 ## Getting Started
 
 ```js
-import { Pubky, PublicKey, Keypair } from "@synonymdev/pubky";
+import { Pubky, PublicKey, Keypair, AuthFlowKind } from "@synonymdev/pubky";
 
 // Initiate a Pubky SDK facade wired for default mainnet Pkarr relays.
 const pubky = new Pubky(); // or: const pubky = Pubky.testnet(); for localhost testnet.
@@ -43,7 +43,7 @@ const addr = `pubky${userPk}/pub/example.com/hello.json`;
 const json = await pubky.publicStorage.getJson(addr); // -> { hello: "world" }
 
 // 5) Authenticate on a 3rd-party app
-const authFlow = pubky.startAuthFlow("/pub/my-cool-app/:rw"); // require permissions to read and write into `my.app`
+const authFlow = pubky.startAuthFlow("/pub/my-cool-app/:rw", AuthFlowKind::signin()); // require permissions to read and write into `my.app`
 renderQr(authFlow.authorizationUrl); // show to user
 const session = await authFlow.awaitApproval();
 ```
@@ -55,7 +55,7 @@ Find here [**ready-to-run examples**](https://github.com/pubky/pubky-core/tree/m
 Use `new Pubky()` to quickly get any flow started:
 
 ```js
-import { Pubky, Keypair } from "@synonymdev/pubky";
+import { Pubky, Keypair, AuthFlowKind } from "@synonymdev/pubky";
 
 // Mainnet (default relays)
 const pubky = new Pubky();
@@ -68,7 +68,7 @@ const pubkyLocal = Pubky.testnet("localhost");
 const signer = pubky.signer(Keypair.random());
 
 // Pubky Auth flow (with capabilities)
-const authFlow = pubky.startAuthFlow("/pub/my-cool-app/:rw");
+const authFlow = pubky.startAuthFlow("/pub/my-cool-app/:rw", AuthFlowKind::signin());
 
 // Public storage (read-only)
 const publicStorage = pubky.publicStorage;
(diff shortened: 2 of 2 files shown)


pubky-sdk/bindings/js/src/pubky.rs
```diff
@@ -54,6 +55,46 @@ impl Pubky {
...
+    /// @throws {PubkyError}
+    /// - `{ name: "InvalidInput" }` for malformed capabilities or bad relay URL
+    /// - `{ name: "RequestError" }` if the flow cannot be started (network/relay)
+    ///
+    /// @example
+    /// const flow = pubky.startAuthFlow("/pub/my-cool-app/:rw");
+    /// renderQr(flow.authorizationUrl);
+    /// const session = await flow.awaitApproval();
+    #[wasm_bindgen(js_name = "startAuthSignupFlow")]
+    pub fn start_auth_signup_flow(
```
SHAcollision: We are missing a `start_auth_signup_flow()` at the entry of the Rust crate like this one. The new actor never gets surfaced on the `Pubky` facade.

pubky-sdk/src/actors/signup_auth_flow.rs
```diff
@@ -0,0 +1,630 @@
...
+
+    /// Set a custom relay for the signup flow.
+    pub fn set_relay(&mut self, relay: Url) {
+        self.relay = relay;
+    }
+
+    /// Convert the `SignupAuthUrl` to a URL.
+    #[allow(
+        clippy::missing_panics_doc,
+        reason = "Should be able to parse the base url"
```
SHAcollision: What part of this function could panic? 🤔

pubky-sdk/bindings/js/src/actors/auth_signup_flow.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Defaults to the default Synonym-hosted relay when omitted.
+    ///
+    /// @returns {AuthFlow}
+    /// A running auth flow. Call `authorizationUrl()` to show the deep link,
+    /// then `awaitApproval()` to receive a `Session`.
+    /// @throws {PubkyError}
+    /// - `{ name: "InvalidInput", message: string }` if any capability entry is invalid
+    ///     or for an invalid relay URL.
+    /// @example
+    /// const flow = AuthFlow.start("/pub/my-cool-app/:rw,/pub/pubky.app/:w");
```
SHAcollision: Copy/pasted docs need to be updated. The `AuthSignupFlow`’s docs still talk about a generic “pubkyauth authorization flow”, omit the required `homeserver_public_key` parameter entirely, and claim the method returns `{AuthFlow}`.

pubky-sdk/src/actors/signup_auth_flow.rs
```diff
@@ -0,0 +1,630 @@
+//! Client <=> Signer signup authing (“pubkysignupauth”) as a single, self-contained flow.
```
SHAcollision: Two other places were some documentation is useful:
- The Rust crate documentation (sourced from `pubky-sdk/README.md`) still only advertises `pubky.start_auth_flow(...)` as the QR/deeplink story. There is no mention of `PubkySignupAuthFlow`, the new builder, or the signup-specific query parameters, so crates.io users will not learn about this feature.
- Simiilarly, the npm README has no reference to `startAuthSignupFlow` or the new deep-link format; it still documents only `startAuthFlow`. npm consumers therefore have no guidance on how to bootstrap a signup QR. `pubky-sdk/bindings/js/pkg/README.md`

SHAcollision (commented):
Nice work @SeverinAlexB !

I left some comments/questions. 

Some high level thoughts here:
- Adding `AuthSignupFlow` as a near-copy of `AuthFlow` creates two public entrypoints that differ mostly in the deep-link builder and the intent (sign-up vs. sign-in), but now have to be kept in lockstep for polling, error handling, logging, and HTTP client usage. Given how similar the behaviors are, we could consider a more unified approach if possible. For example, a single `AuthFlow` with explicit “mode” parameters (e.g., `FlowKind::Signin | FlowKind::Signup { invite_code, homeserver }`) would likely reduce maintenance cost and minimize code drift, while still allowing the specialized deep-link structure needed for signup
- From a developer-experience perspective, one extensible flow type could make the SDK easier to learn: users would discover both sign-in and sign-up in one API surface and share the same builder methods (with optional invite code and homeserver pubkey when doing signup). The current split requires users to learn two actors with largely duplicated method names and lifecycle, and makes it easy to pick the wrong one.

Nits:
- We could consider adding a runnable example to `/examples` that demonstrate how to start the signup flow, set a signup token, or parse the QR parameters.
- We have a `signer.approve_auth()` method that handles the actions needed when an `auth_url` is received/decoded. We could also create a handler for these new URLs (or increase the responsibilities of the existing one if that makes sense).

pubky-sdk/bindings/js/src/actors/deep_links/seed_export.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[wasm_bindgen]
+impl SeedExportDeepLink {
+    #[wasm_bindgen(js_name = "parse")]
+    pub fn try_from(url: &str) -> JsResult<Self> {
+        Ok(Self(
+            pubky::deep_links::SeedExportDeepLink::from_str(url).map_err(|e| {
+                PubkyError::new(
+                    PubkyErrorName::InvalidInput,
+                    format!("Invalid signup deep link: {}", e),
```
SHAcollision:
```suggestion
                    format!("Invalid seed export deep link: {}", e),
```

SHAcollision (approved):
Great work! This is basically ready to merge, I left two minor comments.

PR #271, opened 2025-11-20
Docs/improvements

dzdidi:
Initially was done in scope of https://github.com/pubky/pubky-core/pull/242

Moving to stand alone pr now


SeverinAlexB (commented):
Do we really need the user in this doc? Does it make the docs clearer? Is having an "Authenticator" not good enough already?

PR #273, opened 2025-11-24
chore: set mdbook version explicitly

86667:
Cached mdbook is a different version to the latest which `cargo isntall mdbook` installs. This PR sets consistent behavior between pipeline and local builds.

The motivation for this was to resolve these failures were seeing on merge - https://github.com/pubky/pubky-core/actions/runs/19637265165/job/56231021650


SHAcollision (approved):
LTGM 👍

PR #276, opened 2025-11-27
feat: add pubky keys type

SHAcollision:
- Introduced shared `PublicKey`/`Keypair` wrappers in `pubky-common::crypto::keys` so every crate displays identifiers as `pubky<z32>` by default while still exposing `.z32()` for raw hostnames and registry usage.
- For now we go with `pubky...` as default prefix for the public key identifier. This will be easy to change if the product team decides otherwise.
- Replaced remaining direct `pkarr` key usages across the workspace (SDK, JS bindings, homeserver, testnet) with the shared wrappers to guarantee consistent formatting and serde behavior. The `pkarr-republisher` keeps using `pkarr` keys.
- Removed the re-exported pkarr keys, instead it now re-export pubky keys from `pubky-sdk`, keeping documentation and examples aligned with the new display defaults.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4783,7 +4783,6 @@ dependencies = [
  "indexmap",
  "log",
  "memchr",
- "native-tls",
  "once_cell",
  "percent-encoding",
  "rustls",
--- a/e2e/src/tests/admin.rs
+++ b/e2e/src/tests/admin.rs
@@ -18,7 +18,7 @@ struct InfoResponse {
 #[tokio::test]
 #[pubky_testnet::test]
 async fn admin_info_includes_metadata() {
-    let mut config = ConfigToml::test();
+    let mut config = ConfigToml::default_test_config();
     config.pkdns.public_ip = IpAddr::V4(Ipv4Addr::LOCALHOST);
     config.pkdns.public_pubky_tls_port = Some(9443);
     config.pkdns.icann_domain = Some(Domain::from_str("example.test").unwrap());
--- a/e2e/src/tests/auth.rs
+++ b/e2e/src/tests/auth.rs
@@ -10,12 +10,13 @@ use pubky_testnet::{
 use std::str::FromStr;
 use std::time::Duration;
 
+use super::build_full_testnet;
 use pubky_testnet::pubky::errors::{Error, RequestError};
 
 #[tokio::test]
 #[pubky_testnet::test]
 async fn basic_authn() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let homeserver = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
@@ -111,7 +112,7 @@ async fn disabled_user() {
 #[tokio::test]
 #[pubky_testnet::test]
 async fn authz() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let server = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
@@ -189,7 +190,7 @@ async fn authz() {
 #[tokio::test]
 #[pubky_testnet::test]
 async fn signup_authz() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let server = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
@@ -276,7 +277,7 @@ async fn signup_authz() {
 #[tokio::test]
 #[pubky_testnet::test]
 async fn persist_and_restore_info() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let homeserver = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
@@ -315,7 +316,7 @@ async fn persist_and_restore_info() {
 
 #[tokio::test]
 async fn multiple_users() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let server = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
@@ -339,7 +340,7 @@ async fn multiple_users() {
 #[tokio::test]
 #[pubky_testnet::test]
 async fn authz_timeout_reconnect() {
-    let testnet = EphemeralTestnet::start().await.unwrap();
+    let testnet = build_full_testnet().await;
     let server = testnet.homeserver_app();
     let pubky = testnet.sdk().unwrap();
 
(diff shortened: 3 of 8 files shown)


SHAcollision:
> Nice, I was about to look into a similar PR, glad this is already in place.
> 
> Can you please add a way to get from a `Keypair` to a `PubkyId`?
> 
> For example:
> 
> ```rust
> impl Keypair {
>     fn pubky_id(&self) -> Result<PubkyId, String> {
>         PubkyId::try_from(self.public_key().to_z32().as_str())
>     }
> }
> ```
> 
> which allows
> 
> ```rust
> // Before (verbose)
> let pubky_id = PubkyId::try_from(&keypair.public_key().to_z32())?;
> 
> // After (concise)
> let pubky_id = keypair.pubky_id()?;
> ```

This would introduce a dependency on `pubky-app-specs` here, right? Can we do it without this dependency? I don't think we want that under any circumstance. Instead we likely want a method in `pubky-app-specs` that can `From` a `PubkyId` from a `pubky::PublicKey` ?

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,198 @@
...
+    #[must_use]
+    pub fn random() -> Self {
+        Self(pkarr::Keypair::random())
+    }
+
+    /// Construct a [`Keypair`] from a 32-byte secret key.
+    #[must_use]
+    pub fn from_secret_key(secret: &[u8; 32]) -> Self {
+        Self(pkarr::Keypair::from_secret_key(secret))
+    }
```
SeverinAlexB: Missing secret key as [u8; 32] export.

SHAcollision: Good catch, working on this.

SHAcollision:
Merged main and fixes a myriad of conflicts and problems that rose from working in paralell with the new signup/signin authflow:
- Pubky key wrappers wired through auth flow + deep link signup (hostnames use `.z32()` now) in `auth_flow.rs` and `signup.rs`.
- Session export/import now compiles on native with a non‑WASM import guard in `core.rs`; native `prepare_request` added in native.rs.
- JS bindings aligned to wrapper API in `keys.rs` and pkarr cache seeding uses wrapper conversion in `http.rs`.
- Deterministic SQL ordering for list operations via COLLATE "C" in `repository.rs`.
- Metrics test now uses `z32` host strings in `metrics.rs`; macro doctest example ignored in lib.rs.
- Tightened prefix handling in keys.rs to avoid misparsing raw keys that start with pubky. keys.rs (line 11)
- Switched homeserver URL helpers back to `.z32()` for raw pkarr-style URLs.

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,220 @@
...
+use std::{io, path::Path};
+
+use serde::{Deserialize, Serialize};
+
+type ParseError = <pkarr::PublicKey as TryFrom<String>>::Error;
+
+/// Returns true if the value is in `pubky<z32>` form.
+pub fn is_prefixed_pubky(value: &str) -> bool {
+    matches!(value.strip_prefix("pubky"), Some(stripped) if stripped.len() == 52)
+}
```
SeverinAlexB: NIT: Shouldn't this be a static method that belongs to the `PublicKey` like `PublicKey::is_pubky_prefixed(key: &str)`?

pubky-common/src/crypto/keys.rs
SeverinAlexB: Has this been fixed?

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,220 @@
...
+        Self(pkarr::Keypair::random())
+    }
+
+    /// Export the secret key bytes.
+    #[must_use]
+    pub fn secret_key(&self) -> [u8; 32] {
+        let mut out = [0u8; 32];
+        out.copy_from_slice(self.0.secret_key().as_ref());
+        out
+    }
```
SeverinAlexB: In theory, this is the secret aka seed and not the secret_key aka private key. I know pkarr is already naming it this way but should we fix the naming here at least?

There are obviously other methods that would need to be renamed.

CleanShot 2026-01-09 at 10 05 39@2x

SeverinAlexB (changes requested):
Few comments but I think most of it is good.

Did you ever close the prod or staging database and tested it with the data there to find issues?

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,220 @@
...
+impl From<Keypair> for pkarr::Keypair {
+    fn from(value: Keypair) -> Self {
+        value.0
+    }
+}
+
+/// Wrapper around [`pkarr::PublicKey`] that renders with the `pubky` prefix.
+#[derive(Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct PublicKey(pkarr::PublicKey);
```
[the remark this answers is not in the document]
SHAcollision: This is indeed intended as Claude guessed: this is "storage/transport format" vs "display format". Transport (e.g. query params, `pubky-host`, etc), serialization/deserialization, database storage and QRs strings (e.g. `hs=<z32`) will all still use `z32` keys. 2 reasons: 1. backwards compatibility** with existing payloads/DB data and 2. avoid redundant bytes in transport/storage where the `pubky` prefix adds no semantic value.

I'll take the following action as suggested by claude: document explicitly that serde/transport/DB use raw z32 while `Display` is prefixed for human-facing identifiers.

pubky-common/src/crypto/keys.rs
SHAcollision: Agree, will move into `PublicKey::is_pubky_prefixed`

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+fn parse_public_key(value: &str) -> Result<pkarr::PublicKey, ParseError> {
+    let raw = if is_prefixed_pubky(value) {
+        value.strip_prefix("pubky").unwrap_or(value)
+    } else {
+        value
+    };
+    pkarr::PublicKey::try_from(raw.to_string())
+}
```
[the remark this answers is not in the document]
SHAcollision: Yes, this will help with clarity.

pubky-common/src/crypto/keys.rs
SeverinAlexB: Can we make it the `secret()`? This would still adhere to the pkarr naming to some extend and secret is basically the same as a seed. What do you think?

pubky-common/src/crypto/keys.rs
```diff
@@ -0,0 +1,220 @@
...
+    #[must_use]
+    pub fn into_inner(self) -> pkarr::PublicKey {
+        self.0
+    }
+
+    /// Return the raw z-base32 representation without the `pubky` prefix.
+    #[must_use]
+    pub fn z32(&self) -> String {
+        self.0.to_string()
+    }
```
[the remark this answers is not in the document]
SeverinAlexB: Up to you. I haven't thought deeply about this topic, just seeing this from a reviewers POV

pubky-common/src/crypto/keys.rs
SHAcollision: Sure, I will re-rename into `secret()` makes sense :ok_hand:

SeverinAlexB (approved):
Let's go!

PR #277, opened 2025-11-28
feat: Make admin server optional

86667:
This change requires `admin` and `metrics` server to be enabled/disabled explicitly in the config. In doing so it adds the ability to disabled admin server. 
These changes are backwards compatible with old configs.


SeverinAlexB (approved):
Let's go!

PR #278, opened 2025-11-29
Builder pattern for EphemeralTestnet allowing for custom configurations

86667:
Here we implement a builder for `EphemeralTestnet` for the purpose of allowing a homeserver config to be specified. 
This patterns allows for further customisation possibilities, too. 

Right now with this we:
- Enable Metrics server only where necessary
- Enable Admin server only where necessary

### DIFF
--- a/pubky-testnet/src/ephemeral_testnet.rs
+++ b/pubky-testnet/src/ephemeral_testnet.rs
@@ -7,8 +7,8 @@ use pubky_homeserver::{ConfigToml, ConnectionString, HomeserverApp, MockDataDir}
 ///
 /// Components included:
 /// - A local DHT with bootstrapping nodes.
-/// - An HTTP relay.
 /// - A homeserver (default pubkey: `8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo`).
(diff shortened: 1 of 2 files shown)


pubky-testnet/src/ephemeral_testnet.rs
```diff
@@ -14,60 +14,106 @@ pub struct EphemeralTestnet {
...
-            testnet: Testnet::new_with_custom_postgres(postgres_connection_string).await?,
+            testnet: Testnet::new().await?,
         };
         me.testnet.create_http_relay().await?;
         Ok(me)
     }
 
-    /// Create an additional homeserver with a random keypair
+    /// Create an additional homeserver with a random keypair.
     pub async fn create_random_homeserver(
```
SHAcollision: Homeserver creation helpers disable admin/metrics by default. This changes behaviour for callers that previously might be expecting the extra homeservers to expose admin endpoints out of the box; if that change is intentional it should be documented and the helper name/comment updated, otherwise the defaults should be restored or gated by parameters. I can't recall if the homeserver admin server is ever used or not in any of the Nexus test cases. Maybe it is on the `pubky-moderation` tests?

PR #284, opened 2025-12-01
fix(sdk): null rust pointer error on authflow

SHAcollision:
- First added some MRE (as tests) for panicking the wasm runtime from the JS caller in several where it could panic on the AuthFlow.

- Guarded the JS AuthFlow wrapper with shared state and client-state errors so repeat awaits or polling after completion no longer trigger WASM null pointer panics.

- Unrelated by needed: Documented the Rust and JS pattern for reusing a shared/global Pubky facade instead of recreating it per request.

After merging the lastes from `main` figured out a small bug as well:

- Capped HTTP relay polling to three consecutive failures so auth flows surface request errors instead of retrying forever when the relay is unreachable.

### DIFF
--- a/pubky-sdk/bindings/js/src/pubky.rs
+++ b/pubky-sdk/bindings/js/src/pubky.rs
@@ -16,6 +16,10 @@ pub struct Pubky(pub(crate) pubky::Pubky);
 impl Pubky {
     /// Create a Pubky facade wired for **mainnet** defaults (public relays).
     ///
+    /// Prefer to instantiate only once and use trough your application a single shared `Pubky`
+    /// instead of constructing one per request. This avoids reinitializing transports and keeps
+    /// the same client available for repeated usage.
+    ///
     /// @returns {Pubky}
     /// A new facade instance. Use this to create signers, start auth flows, etc.
     ///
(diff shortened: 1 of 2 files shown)


pubky-sdk/src/actors/auth/http_relay_link_channel.rs
```diff
@@ -131,11 +133,15 @@ impl HttpRelayLinkChannel {
                 Err(e) => match e {
                     PollError::Timeout => {}
```
86667: nit: Why is timeout not a failure? I feel this could be tidied a bit so that we need to track only `attempt`

86667 (dismissed):
Not a nice bit of code that `try_poll_once()` fn 😢 

I dont know a cleaner solution though, looks good apart from nits

pubky-sdk/src/actors/auth/http_relay_link_channel.rs
SHAcollision: We expect the http-relay channel long poll to timeout always in 600 seconds (10 minutes). The behaviour should be to simply re-create the channel and keep polling for 10 more minutes forever, so downstream users do not need to worry about the QR code expiring if the auth is not approved right away.

86667 (approved):

PR #293, opened 2025-12-29
feat: events-stream sdk

86667:
SDK builder API for using `/events-stream`.

### DIFF
--- a/pubky-sdk/src/actors/event_stream.rs
+++ b/pubky-sdk/src/actors/event_stream.rs
@@ -37,9 +37,9 @@ use std::fmt::Display;
 use std::pin::Pin;
 use std::str::FromStr;
 
+use crate::PublicKey;
 use eventsource_stream::Eventsource;
 use futures_util::{Stream, StreamExt};
-use pkarr::PublicKey;
 use reqwest::Method;
 use url::Url;
 
@@ -253,15 +253,15 @@ impl EventStreamBuilder {
     /// Constructs a URL like:
     /// `https://{homeserver}/events-stream?user=pk1&user=pk2:cursor&limit=100&live=true&path=/pub/`
     fn build_request_url(&self, homeserver: &PublicKey) -> Result<Url> {
-        let mut url = Url::parse(&format!("https://{homeserver}/events-stream"))?;
+        let mut url = Url::parse(&format!("https://{}/events-stream", homeserver.z32()))?;
 
         {
             let mut query = url.query_pairs_mut();
             for (user, cursor) in &self.users {
                 if let Some(c) = cursor {
-                    query.append_pair("user", &format!("{user}:{c}"));
+                    query.append_pair("user", &format!("{}:{c}", user.z32()));
                 } else {
-                    query.append_pair("user", &user.to_string());
+                    query.append_pair("user", &user.z32());
                 }
             }
             if let Some(limit) = self.limit {


pubky-sdk/src/actors/event_stream.rs
```diff
@@ -0,0 +1,446 @@
...
+    Pkdns, PubkyHttpClient, cross_log,
+    errors::{Error, RequestError, Result},
+};
+
+/// Cursor for pagination in event queries.
+///
+/// The cursor represents the ID of an event and is used for pagination.
+/// It can be parsed from a string representation of an integer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct EventCursor(i64);
```
SHAcollision: What's the type used on te homeserver side for cursors, is it also `i64` ? I wonder if there is some reason to restrict to only positive values `u64`

pubky-sdk/bindings/js/src/actors/event_stream.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// @param {PublicKey} user - User public key
+    /// @param {number | null} cursor - Optional cursor position (event ID) to start from
+    /// @returns {EventStreamBuilder} - Builder for chaining
+    /// @throws {Error} - If trying to add more than 50 users
+    #[wasm_bindgen(js_name = "addUser")]
+    pub fn add_user(
+        self,
+        user: &PublicKey,
+        cursor: Option<f64>,
```
SHAcollision: I wonder, can we enforce `Option<i64>` here? The JS binding accepts coverts using `as` without validation. This silently truncates floats and can overflow/lose precision for very large values, which is very possible for monotonic event IDs.

If `Option<i64>` is not it's possible to accept `string | bigint` (or enforce integer checks on number input) and validate that the cursor is a safe integer before casting. Consider aligning with the Rust API by accepting string input in JS and parsing to `i64` with explicit error handling.

pubky-sdk/bindings/js/src/pubky.rs
```diff
@@ -171,4 +168,36 @@ impl Pubky {
...
+    ///
+    /// @returns {EventStreamBuilder} A builder to add users and configure the event stream
+    ///
+    /// @example
+    /// ```typescript
+    /// const user1 = PublicKey.from("o1gg96ewuojmopcjbz8895478wdtxtzzuxnfjjz8o8e77csa1ngo");
+    /// const user2 = PublicKey.from("pxnu33x7jtpx9ar1ytsi4yxbp6a5o36gwhffs8zoxmbuptici1jy");
+    /// const stream = await pubky.eventStream()
+    ///   .addUser(user1, null)
+    ///   .addUser(user2, 100)
```
SHAcollision: Nit: cursor is `@param {string | null}`
```suggestion
    ///   .addUser(user2, "100")
```

pubky-homeserver/src/persistence/files/events/events_entity.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct EventEntity {
-    pub id: i64,
+    pub id: u64,
```
SHAcollision: very nit: codex brought this up. Unclear to me if this really is an issue.

> The homeserver repository now uses `u64` for `EventCursor` and compares it directly in SQL queries. Postgres doesn’t have unsigned integer types; in practice IDs are still in the `i64` range, so it should be fine. However, it’s worth verifying that `sea_query` + `sqlx` correctly bind `u64` to BIGINT without unexpected type coercions.
> **Recommendation:** Sanity-check the SQL binding for `u64` (or consider converting to `i64` at the query boundary) if any runtime issues appear.

SHAcollision (approved):
Nice work. LGTM! 🚀  I leave some minor nits, but they are non blocking IMO.

pubky-homeserver/src/persistence/files/events/events_entity.rs
86667: Ive gone with documenting the assumption that we wont overflow i64 for a single incrementing counter. Runtime check felt like a waste and is unsolvable without db migration anyway. This could only go wrong if cursor changed from being single incrementing.

ok300: True! Thanks.

PR #295, opened 2026-01-08
Fix: apply C collation to entry ordering and cursors for deterministic ordering

SHAcollision:
We figured out there are differences on local vs CI testing because the ordering of deep lists is different. 

## The exact reason for this failure of tests, sometimes:
- The e2e test `tests::storage::list_deep` expects paths under `/pub/example.com/` to be returned in byte-wise ASCII order (a.txt → b.txt → c.txt → cc-nested/z.txt → d.txt). The homeserver’s SQL query orders rows only with `ORDER BY entries.path ASC`, so the ordering is delegated to PostgreSQL’s default collation.
- On clusters initialized with an ICU-backed default collation (e.g., Ubuntu 24’s PostgreSQL 17+ defaults), punctuation is largely ignored when sorting. Under that collation, `"/pub/example.com/cc-nested/z.txt"` is considered closer to `"/pub/example.com/c…"` than `"/pub/example.com/c.txt"`, making `cc-nested` appear before `c.txt`. GitHub CI uses a libc/`C`-style collation where punctuation is honored, so the CI ordering matches the test expectation.

## So the solution this PR implements is:
- Apply `COLLATE "C"` to entry list ordering and cursor comparisons for deterministic, byte-wise pagination across locales.
- Additionally, I also fixed the docs for `pubky-testnet` postgres connection string.

## However, there could be some caveats, that I think are fine for now:
- Using `COLLATE "C"` in query expressions can reduce index usage on text columns and may impact query performance.
- Only entry listings are covered by the fix on this PR; but I guess there could still be other string-based ordering still depends on database defaults?

## Going forward
We should consider enforcing LC_COLLATE='C'/LC_CTYPE='C' at database creation (test and real) to avoid query-level overrides. But I assume this is a change we might not be able to apply to existing prod databases?
We could evaluate index strategy if performance regressions shows up. I guess we won't notice anything.

### DIFF
--- a/pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
+++ b/pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
@@ -348,17 +348,11 @@ impl EntryRepository {
 
         if reverse {
             statement = statement
-                .order_by_expr(
-                    Expr::cust("entries.path COLLATE \"C\""),
-                    Order::Desc,
-                )
+                .order_by_expr(Expr::cust("entries.path COLLATE \"C\""), Order::Desc)
                 .to_owned();
         } else {
             statement = statement
-                .order_by_expr(
-                    Expr::cust("entries.path COLLATE \"C\""),
-                    Order::Asc,
-                )
+                .order_by_expr(Expr::cust("entries.path COLLATE \"C\""), Order::Asc)
                 .to_owned();
         }
 


pubky-testnet/README.md
```diff
@@ -15,16 +16,17 @@ docker run --name postgres \
   -e POSTGRES_DB=pubky_homeserver \
   -p 127.0.0.1:5432:5432 \
   -d postgres:18-alpine
+```
 
 # Run the testnet binary (all resources ephemeral). The environment variable must point to the postgres admin database.
-TEST_PUBKY_CONNECTION_STRING='postgres://postgres:postgres@localhost:5432/postgres' cargo run -p pubky-testnet
 
+TEST_PUBKY_CONNECTION_STRING='postgres://postgres:postgres@localhost:5432/postgres?pubky-test=true' cargo run -p pubky-testnet
```
SHAcollision: Autoformatting was applied to this `.md` file on this PR, but the only change that matters is the addition of `?pubky-test=true` on the connection string.

pubky-testnet/README.md
```diff
@@ -15,16 +16,17 @@ docker run --name postgres \
   -e POSTGRES_DB=pubky_homeserver \
   -p 127.0.0.1:5432:5432 \
   -d postgres:18-alpine
+```
```
SHAcollision: This snippet was missing the closure.

SeverinAlexB (dismissed):
Let's go

PR #297, opened 2026-01-13
Re-export `pkarr`

ok300:
This PR updates the `pubky` crate to re-export its `pkarr` dependency.

The reason for this is that `pubky` is a higher-level abstraction over `pkarr`, but callers might still want to work with `pkarr` objects and methods. This PR would make dev experience better in that case. For example, it would allow Nexus to drop `pkarr` as a direct dependency: https://github.com/pubky/pubky-nexus/pull/684


ok300:
I see what you mean.

However I don't think that necessarily holds. Re-exports are a courtesy. Callers can opt out of breaking `pubky::pkarr` changes by using `pkarr` directly, but can't do so for breaking `pubky` changes. Major releases should therfore reflect breaking changes in `pubky`, as the caller has to accept them, but not necessarily those of dependencies (even if re-exported), since callers are not forced to accept them.

PR #300, opened 2026-01-21
chore: DRY up auth approval for signup deeplinks

SHAcollision:
Low priority
This PR changes:
- Parse `pubkyauth` URLs once and route signup vs relay auth from a single path.
- Decrease complexity of some functions and remove the clippy `allow` for high cognitive complexity.
- Reuse shared token-posting helper for relay flows (signup / signin).
- Explicitly support direct signup deeplinks (no relay/secret) without duplicating logic.
- Updated canonical ordering of the `params` from mandatory first (`hs`) to optional last (`st`...) and the test expectation strings.
- The `caps` param is now not enforced. By `PubkyAuth` standard an `AuthToken` with no capabilities is used to "authenticate" (prove identity). When some capabilities are requested, then it is used to "authorize" (give access to that scope). Previously `caps` were always enforce to exist.
- The parsing logic now enforces relay and secret pairing: if one is present without the other, parsing fails.

### DIFF
--- a/e2e/src/tests/auth.rs
+++ b/e2e/src/tests/auth.rs
@@ -221,7 +221,10 @@ async fn signup_authz() {
     };
     assert_eq!(signup_deep_link.capabilities(), &caps);
     assert_eq!(
-        signup_deep_link.relay().as_str(),
+        signup_deep_link
+            .relay()
+            .expect("signup deep link should include relay")
+            .as_str(),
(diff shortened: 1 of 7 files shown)


SeverinAlexB:
@MCarlomagno This is an older PR that wants to introduce a signup deeplink. The SDK changed a lot since then. deep links have seen a full refactor. Can you rewrite it so we finally get it into the codebase?

PR #304, opened 2026-02-07
fix: Fix crate naming issue preventing `dependabot` runs

ok300:
The dependabot scheduled jobs were failing[^1] with:

> Dependabot can't resolve your Rust dependency files
> ...
> the binary target name examples is forbidden, it conflicts with cargo's build directory names

The package in `examples/rust/Cargo.toml` had `name = "examples"`, which conflicts with Cargo's reserved build directory names (`examples`, `target`, `tests`, `benches`, etc.). Dependabot's Cargo version enforces this restriction, causing it to fail when parsing the workspace manifest.

This PR addresses this issue by renaming the package from `examples` to `pubky-core-examples` in `examples/rust/Cargo.toml`.

[^1]: See job logs at https://github.com/pubky/pubky-core/network/updates

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -150,7 +150,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Cargo check examples
         run: |
-          cargo check -p examples --all-features --bins
+          cargo check -p pubky-core-examples --all-features --bins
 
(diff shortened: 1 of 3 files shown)


SeverinAlexB (approved):
Fingers crossed this doesn't break anything else (it shouldn't really). Let's try :)

PR #305, opened 2026-02-07
chore: Group `dependabot` PRs

ok300:
This PR configures `dependabot` to group all update PRs into one, instead of a separate PR per dependency update.

Ideally this should be merged before #304.


SeverinAlexB (approved):
Thank you sir!

PR #315, opened 2026-02-16
Bump `pkarr`, `mainline`, `pkarr-relay`

ok300:


ok300:
> Changes look OK to me. Only one quesiton, it seems now we have reqwest 0.12 and 0.13 on the tree, should we tackle that or just open an issue for later?

Done, `reqwest` dependency is consolidated in workspace root and bumped to `0.13`

pkarr-republisher/src/multi_republisher.rs
```diff
@@ -217,7 +217,7 @@ mod tests {
         let mut settings = RepublisherSettings::default();
         settings
             .pkarr_client(pkarr_client)
-            .min_sufficient_node_publish_count(NonZeroU8::new(3).unwrap());
+            .min_sufficient_node_publish_count(NonZeroU8::new(4).unwrap());
```
SeverinAlexB: Comment to document our discussion on the call:

It looks like there is a bug either in pkarr/mainline or the republisher. If the testnet only has 3 nodes, it can't be published to 4 nodes.

@ok300 Can you add a TODO comment to one of tests so devs will remember?

SeverinAlexB (approved):
lgtm

PR #316, opened 2026-02-17
feat: Replace Http-Relay with new /inbox endpoint

SeverinAlexB:
## Replace the http-relay

This PR adds the new [pubky-http-relay](https://github.com/pubky/pubky-http-relay/) repo and removes the existing http-relay member. 

- Removed the in-repo http-relay crate and replaced it with the new [pubky-http-relay](https://github.com/pubky/pubky-http-relay/) dependency, reducing maintenance burden. `pubky-http-relay` has an improved `/inbox` endpoint that fixes https://github.com/pubky/pubky-ring/issues/254. 

## Support the new /inbox in the SDK

- Added `HttpRelayInboxChannel` which replaces the `HttpRelayLinkChannel`. These structs have duplicated code on purpose so we can remove `HttpRelayLinkChannel` in a future version without any trouble.
- Unified auth subscription behind an EncryptedAuthChannel enum that auto-dispatches between /inbox and /link channels. If the relay url ends with `/link`, it will use the old approach, if it uses anything else, it will use the new `/inbox` approach.
- Updated default relay URL from link-based to inbox-based (DEFAULT_HTTP_RELAY_INBOX)

## MISC

- Set Rust Minimum Supported Rust Version (MSRV) to 1.89 and added CI check for it. This makes sure that we and our dependencies are compatible with the set version.

## TODO
- [x] Wait until the new relay is deployed to `https://httprelay.pubky.app` because the new inbox relay is used by default from now on.

### DIFF
--- a/pubky-testnet/src/static_testnet.rs
+++ b/pubky-testnet/src/static_testnet.rs
@@ -179,6 +179,7 @@ impl StaticTestnet {
     async fn run_fixed_http_relay(&mut self) -> anyhow::Result<()> {
         let relay = HttpRelay::builder()
             .http_port(15412) // Random available port
+            .cors_allow_all(true)
             .run()
             .await?;
         self.testnet.http_relays.push(relay);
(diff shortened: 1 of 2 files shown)


SHAcollision (dismissed):
The PR looks great, great job. Approving as is. Left small nits, if you wan to tackle those, I'll re-approve.

I understand that by using new [pubky/http-relay](https://github.com/pubky/http-relay) on `pubky-testnet` the `e2e` tests and `pubky-sdk/bindings/js/pkg/tests` running now are using the `/inbox` always, right?

PR #325, opened 2026-02-25
feat: set-version.sh and publish-libs.sh to be per-crate only

86667:
I want to be able to publish a single one of this repo's crates alone. 
=> Each module should have a version number independent of the other workspace members.
=> Publishing and releasing should be per-crate.

Therefore these changes allow for individual modules to be published separately. 
Ive removed the ability to publish multiple at the same time in order to reduce complexity of the scripts and make it clear that there probably isnt much of a use case for doing so anyway.


SeverinAlexB (approved):
Havent't tested the new sed but I assume it works. Let's go

PR #357, opened 2026-03-26
feat!(hs): Remove LMDB

SeverinAlexB:
Fully removes any LMDB code. We introduced postgres with version v0.6.0. We had enough time to update. This PR now removes any LMDB support.


86667 (approved):
Very good, very good.

A couple pre-existing issues found by AI which may be worth fixing now whilst we're here in `migration.rs`. Otherwise good to.

#### Migrator::run_migration silently swallows errors (Pre-existing)

At [migrator.rs:87-96](vscode-webview://1fq2qhl3t8ljrt2r5akosl3aq26vet4tagge6uv6tl307br8890j/pubky-homeserver/src/persistence/sql/migrator.rs#L87-L96), when a migration fails, the error is logged and the transaction rolled back, but Ok(()) is returned. This means run_migrations won't propagate the failure — it will appear as if migrations ran successfully.
#### Test uses SQLite query against Postgres (Pre-existing)

At [migrator.rs:274](vscode-webview://1fq2qhl3t8ljrt2r5akosl3aq26vet4tagge6uv6tl307br8890j/pubky-homeserver/src/persistence/sql/migrator.rs#L274), the rollback test queries sqlite_master which won't work against Postgres. This test likely passes only because the fetch_one fails (which is the expected assertion), but for the wrong reason.

86667 (approved):

Issue #185, opened 2025-07-16
Add configurable data dir to config for pubky-testnet


catch-21:
It's not possible to choose the data dir for `pubky-testnet`, it always uses an ephemeral tmp dir. It would be useful to point it at a persistent location for some local testing. 

That is, we should be able to:
- Specify a file for the the testnet to store its data
- Write its data contents to file (rather than the existing forced `StorageConfigToml::InMemory`)
- Stop the testnet and restart with the same path, with the same data persisting

PastaGringo:
Did it on my docker compose and it worked :

```
  volumes:
      - ./homeserver/config/homeserver.config.toml:/root/.pubky/config.toml
      - ./homeserver/data:/root/.pubky
```

86667:
@techraed I have updated the issue description, removing the ask for the homeserver's data path being configurable. I think that @PastaGringo's solution there is enough.

PR #361, opened 2026-03-28
feat(testnet): add possibility to run  `pubky-testnet` using persistent file storage

techraed:
Resolves issue #185 

Changes summary:
~~1. homeserver CLI now accepts 3 path args: to data dir, to secret file and to config file. The resolution algo:
    - If path is defined, then the value is used
    - If path is not define, then data dir path is used as a base path
    - If data dir is not defined, then home directory is used (`~/.pubky`)
    The CLI behavior didn't change if no params are provided or only if data dir param is provided - that way changes are backward compatible.
2. `PersistentDataDir` is renamed to `HomeserverPaths`, and the latter one stores all three resolved paths. Consequently, the module `data_directory` is also renamed to `homeserver_config`.
3. Due to changes semantics described upper, `DataDir` is now renamed to `SetupSource` (also module is renamed). The trait is no longer a `Clone` implementor, as homeserver app covers the trait object with `Arc`.
4.  `MockDataDir` is renamed to `MockSetupSource`. The module is also renamed. The type now can be instantiated with a path to data dir.
5. `EphermalTestnet` can now define a data dir in its' builder. If data dir was defined for an ephermal testnet, then a file system will be used by open-dal. Otherwise testnet storage remains in-memory.~~

**UPD**: the original issue was changed. See further comments to understand the context and changes summary.


86667:
Hey @techraed thanks for your work! 

Nicely done, I have a couple comments around making changes as simple as possible in open codebases.

There are a lot of changes here. Im not sure why the cli has changed, and the refactor and directory rename makes it difficult to understand what is happening and is technically a breaking change. 

Please separate behavioral changes from refactors either in different PRs or at least different commits, and seriously consider whether a rename of a common type is necessary. It may be that the renaming/refactor is absolutely necessary for the desired behaviour to work, if so please justify it.

> I have a question regarding pubky-testnet having defined a path to data dir. What's the use case for that?

Hm, if the second option is more difficult then I agree lets keep it for a separate piece of work. Achieving only 1. is still progress in the right direction.

techraed:
> making changes as simple as possible in open codebases.

> refactor and directory rename makes it difficult to understand

Yeah, you right about that. Sorry, forgot that for open codebases it's usually preferred small and only relative changes.

Anyway, let me try to justify these changes.

## The CLI
The original issue states the following: *"It's not possible to choose the data dir for the homeserver"*. And the same issue is for the `pubky-testnet`.

Frankly speaking, the `master` version of the cli __allows__ you defining the path to some other file storage. But there are several issues.

### Problem-1 Config and the data path are the same
The problem is that under the path there __must__ the config and the secrets file. Therefore if you have some other files location, you must transfer there your config and secrets file, because the setup code for the homeserver has a strong dependency of the config and secrets files from the data directory (must be in the same directory).

### Problem-2 Low UX mitigations
Alright, you could mitigate that just by the following changes to storage config toml:
```rust
pub enum StorageConfigToml {
    /// Files are stored on the local file system.
    FileSystem { path: PathBuf }, // <------------- Added here the path

    // other variants here
}
```
That would allow not touching the CLI, but the UX would be very low — basically, you would need to create a directory, inside the directory you **must** define a config file, which **must** define the path.

### Solution
So what's the best and ergonomic way (imho) to define a custom data dir without having to think of config and secretes files being close to data dir? Change the CLI!

```shell
# Calling like this would create everything inside the ~/.pubky
./target/release/homeserver

# Remains the same behavior
./target/release/after_this_pr_homeserver

# Calling like this would create everything inside the /etc/some/path/to/data
./target/release/homeserver --data-dir /etc/some/path/to/data

# Remains the same behavior
./target/release/after_this_pr_homeserver --data-dir /var/lib/homeserver/data

# Now config and data are in the separate dirs. Secrets will be created in the data dir.
./target/release/after_this_pr_homeserver --data-dir ~/data --config /etc/configs/homeserver_config.toml

# Now everything is accessible from their own paths
./target/release/after_this_pr_homeserver --data-dir /var/lib/homeserver/data --config /etc/configs/homeserver_config.toml --secret-key /etc/secrets/homeserver_secrets
```

## Renaming because of the CLI changes
So the CLI changes bring the separation between different setup sources. So we can't say now the path is to persistent data directory. It's now a path to different configs. That's now a directory is renamed: `data_directory` —> `homeserver_configs`. Same renaming for `data_dir.rs` module and the crate `DataDir`: `data_dir.rs` —> `setup_source.rs` and `trait DataDir` —> `SetupSource`.

## `pubky-testnet` related changes
The original issue states *"pubky-testnet always uses an emphemeral tmp dir. It can be useful to point it at a persistent location for some local testing."*.

To do that the original `MockDataDir`
```rust
pub struct MockDataDir {
    pub(crate) temp_dir: std::sync::Arc<tempfile::TempDir>,
    pub config_toml: super::ConfigToml,
    pub keypair: pubky_common::crypto::Keypair,
}
```
is changed to
```rust
pub struct MockSetupSource {
    root: MockDataDir, // THE CHANGE !
    pub config_toml: super::ConfigToml,
    pub keypair: pubky_common::crypto::Keypair,
}

enum MockDataDir {
    Temp(std::sync::Arc<tempfile::TempDir>),
    Persistent(PathBuf),
}
```

## Renaming because of the `MockDataDir` changes
That's related to `DataDir` trait name changes to `SetupSource`.

# Final comment
@86667 You are right about the size (scope) of changes, these renamings trigger moving files and review becomes complicated. But the changed nature of original entities made me bring all the changes (feature and refactoring) together. Hope that made everything clear for you.

techraed:
@86667 shall I remain the same the PR? or divide changes related to logic and renamings (different commits)?

86667:
Hey @techraed 

> Anyway, let me try to justify these changes.

I understand that there is some way to justify 41 files changes and breaking the interface between modules. The question is whether its worth the pain. Ideally, we would keep changes as minimal as possible so that we can quickly test and confirm all is well, then iterate quickly. 

> @86667 shall I remain the same the PR? or divide changes related to logic and renamings (different commits)?

This would be great, if we could have a minimal PR which solves a single problem.

This being said, I do agree that the task here is fairly under-specified and the development of it is itself research into what is best. I think that maybe we could achieve what we need (data dir specified so that it can be reused later) by the more simple change of allowing the path of  `data` to be specified but keeping `config.toml` and `secret` in `HOME/.pubky`. What do you think?

That way, we could later add an optional var to `config.toml` which points to the data location and then not have to worry about the `--data-dir` flag each time we run. I think the `DataDir` trait may do most of thee heavy lifting for us here, though i may be incorrect.

Looking forward to seeing what you think, thanks again for getting invovled!

86667:
I re-considered what were doing here and decided to update the original issue. The actual need is in the testnet only, the homeserver I think we can get away with not changing.

See this issue comment - https://github.com/pubky/pubky-core/issues/185#issuecomment-4198862187

techraed:
Sure, that's a right thing to do. After the https://github.com/pubky/pubky-core/pull/361#issuecomment-4197916303 I thought in the same direction.

Will do

techraed:
hey @86667
Now it's ready for the review.

Only focused on providing an ability to use persistent data storage (i.e., file system) for tests. Basic scenario which was described in the https://github.com/pubky/pubky-core/issues/185#issuecomment-4198862187 is implemented in here - https://github.com/pubky/pubky-core/pull/361/changes#diff-87ec3cd0feb2ec7271c55b5a28635507f0a79f26433b869d2977b872c07cf6a9R922.

Summary of changes:
1. `MockDataDir` can be instantiated either with a temp file storage or a persistent one by providing a path to it.
2. Test DB lifecycle is now opt-in for cleanup. `EphemeralTestnetBuilder::new().drop_db_on_cleanup(false)` - this now sets a special param in the query part of the connection string (like `postgres://localhost:5432/postgres?pubky-test-persist`), that signals to `TestDbDropper` not to drop a DB, so a test DB can survive across testnet restarts. That's vital for test database (created via `pubky_homeserver::persistence::sql::SqlDb::create_test_database`) if we opt file storage to persist, because if database is dropped, then all the metadata for the file storage is lost and we couldn't use the existing file storage via a homeserver from a new testnet.
3. An ability to obtain a db connection string from the running homeserver is enabled for test databases. The returned value also includes the target DB name as a query param (like `postgres://localhost:5432/postgres?pubky-test-db-name=<some_name>`), so a second testnet can reconnect to that DB directly instead of creating a new one. The connection string can be provided into `EphemeralTestnetBuilder::new().postgres(connection_with_db_name_param)`.

pubky-homeserver/src/data_directory/mock_data_dir.rs
```diff
@@ -52,11 +99,25 @@ impl Default for MockDataDir {
...
+        }
     }
 
     fn ensure_data_dir_exists_and_is_writable(&self) -> anyhow::Result<()> {
-        Ok(()) // Always ok because this is validated by the tempfile crate.
+        if let MockDataDirKind::Persistent(path) = &self.root {
+            std::fs::create_dir_all(path)?;
+
+            // Check if we can write to the data directory
+            let test_file_path = path.join("test_write_f2d560932f9b437fa9ef430ba436d611"); // random file name to not conflict with anything
```
86667: This could conflict with itself. Generate a value or include timestamp maybe?

e2e/src/tests/storage.rs
```diff
@@ -916,3 +916,74 @@ async fn write_same_path_separate_users() {
...
+    let homeserver = testnet.homeserver_app();
+
+    // Sign in with the same user keypair to get a session on the new testnet.
+    // The new testnet has a fresh DHT, so we first publish the PKDNS record
+    // pointing this user to the homeserver, then signin (which resolves via DHT).
+    let keypair = Keypair::from_secret(&[1; 32]);
+    let signer = pubky.signer(keypair);
+    signer
+        .pkdns()
+        .publish_homeserver_force(Some(&homeserver.public_key()))
```
86667, line 978: I would expect the homeserver to automatically republish the record to the new DHT. Does this not happen? Or is this so that the test doesnt have to wait?

pubky-homeserver/src/data_directory/mock_data_dir.rs
```diff
@@ -1,37 +1,68 @@
...
+    /// verify persistence).
+    pub fn new_persistent_data_dir(
+        data_dir: PathBuf,
+        config_toml: super::ConfigToml,
+        keypair: Option<pubky_common::crypto::Keypair>,
+    ) -> anyhow::Result<Self> {
+        let keypair = keypair.unwrap_or_else(pubky_common::crypto::Keypair::random);
+        std::fs::create_dir_all(&data_dir)?;
+
+        debug_assert!(
```
86667: Why use debug here? Feels like something we would always want to check for?

pubky-homeserver/src/data_directory/mock_data_dir.rs
```diff
@@ -52,11 +99,25 @@ impl Default for MockDataDir {
...
     fn path(&self) -> &Path {
-        self.temp_dir.path()
+        match &self.root {
+            MockDataDirKind::Temp(temp_dir) => temp_dir.path(),
+            MockDataDirKind::Persistent(path) => path.as_path(),
+        }
     }
 
     fn ensure_data_dir_exists_and_is_writable(&self) -> anyhow::Result<()> {
-        Ok(()) // Always ok because this is validated by the tempfile crate.
```
86667, line 59:
```
// Always ok because this is validated by the tempfile crate.
```
This comment is still useful for the `MockDataDirKind::Temp` case i think. Maybe `match` on `MockDataDirKind` and put the comment in the `Temp` branch

pubky-homeserver/src/persistence/sql/connection_string.rs
```diff
@@ -42,18 +42,91 @@ impl ConnectionString {
 
 #[cfg(any(test, feature = "testing"))]
 impl ConnectionString {
+    const TEST_FLAG_KEY: &'static str = "pubky-test";
+    const TEST_DB_NAME_KEY: &'static str = "pubky-test-db-name";
+    const TEST_PERSIST_KEY: &'static str = "pubky-test-persist";
+    const DEFAULT_CONNECTION_STRING: &'static str =
+        "postgres://localhost:5432/postgres?pubky-test=true";
```
86667, line 59: Why have you changed the default connection string? This will break devs with an authenticated local setup (which is likely everyone)

pubky-homeserver/src/data_directory/mock_data_dir.rs
```diff
@@ -1,37 +1,68 @@
...
+
+        Ok(Self {
+            root: MockDataDirKind::Persistent(data_dir),
+            config_toml,
+            keypair,
+        })
+    }
+
+    /// Returns true if the root of this [`MockDataDir`] is a temporary directory.
+    pub fn is_temp(&self) -> bool {
```
86667: Did you intend to use this? If no, remove

pubky-homeserver/src/data_directory/mock_data_dir.rs
```diff
@@ -40,8 +71,24 @@ impl MockDataDir {
...
+
         Self::new(config, Some(keypair)).expect("failed to create MockDataDir")
     }
+
+    /// Creates a [`MockDataDir`] with a config and keypair for testing, backed by a real directory.
+    ///
+    /// Same as [`MockDataDir::test()`] but with a real directory that is not cleaned up on drop.
+    /// Use this for integration tests that need to verify persistence across process restarts.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn test_persistent_data_dir(data_dir: PathBuf) -> Self {
```
86667: Unused, was this intended for use? The integration test uses `.with_data_dir()`

pubky-testnet/src/ephemeral_testnet.rs
```diff
@@ -187,10 +214,22 @@ impl EphemeralTestnetBuilder {
...
+            config.general.database_url.remove_persist_param();
+        }
+
         let keypair = self
             .homeserver_keypair
             .unwrap_or_else(|| Keypair::from_secret(&[0; 32]));
-        let mock_dir = MockDataDir::new(config, Some(keypair))?;
+        let mock_dir = if let Some(data_dir) = self.data_dir {
+            config.storage = StorageConfigToml::FileSystem;
+            MockDataDir::new_persistent_data_dir(data_dir, config, Some(keypair))?
```
86667, line 228: This all feels a bit sloppy. We have `data_dir`, `config.storage` and `MockDataDirKind` all determining whether were using memory or file system, and acting on that info in different places.

86667: We should figure out a way to have a single decision point with all `MockDataDirKind` logic in one place. The memory/file system decision is spread across three layers, they should ideally be in-sync at construction time

pubky-homeserver/src/persistence/sql/sql_db.rs
```diff
@@ -50,40 +50,73 @@ impl SqlDb {
...
     fn drop(&mut self) {
-        // Drop the database after the test.
+        // Drop the database after the test if dropper switch is on,
+        // i.e. self.should_drop is `true`.
         // This works in combination with the pubky_test macro.
-        let _ = pubky_test_utils::register_db_to_drop(
-            self.db_name.clone(),
-            self.connection_string.clone(),
-        );
+        if self.should_drop {
```
86667: Very fishy. `Drop` is for dropping. 

` db_dropper: Option<std::sync::Arc<TestDbDropper>>,` is optional for this reason, just dont set it. Here youre setting the dropper and passing in a flag telling it to carry out its only purpose or not. This is sloppy and how code bases get into spaghetti mess.

If you need the `TestDbDropper` metadata then maybe consider pulling it into a separate `TestDbInfo` struct, and only creating the dropper when cleanup is actually wanted.

86667 (changes requested):
I may be wrong here, but it seems like the quickest path to get the functionality working is implemented and pushed without much thought around the code base's abstractions, understand-ability and ease to build upon.

With AI, code is cheap. But so also is review, research, design discussion, etc. Therefore as code cost decreases the expected code quality increases. 

Happy to review once more, but i don't have the resources to continue a back and forth for issues which a simple AI "review this code" prompt can find. Hope you understand.

pubky-homeserver/src/data_directory/mock_data_dir.rs
techraed: done in [abf19b8](https://github.com/pubky/pubky-core/pull/361/commits/abf19b83ce036ef23a1181c48b524ccba88dba37)

uses now unique names for files generated with uuid

pubky-homeserver/src/data_directory/mock_data_dir.rs
techraed: done in [abf19b8](https://github.com/pubky/pubky-core/pull/361/commits/abf19b83ce036ef23a1181c48b524ccba88dba37)

pubky-homeserver/src/data_directory/mock_data_dir.rs
techraed: done in [e9f90b1](https://github.com/pubky/pubky-core/pull/361/commits/e9f90b154952605e7134b4c1d87acf6bf3a73fcc)

Just a convenient public API for public type

pubky-homeserver/src/data_directory/mock_data_dir.rs
techraed: done in [e9f90b1](https://github.com/pubky/pubky-core/pull/361/commits/e9f90b154952605e7134b4c1d87acf6bf3a73fcc)

That's a public type with public API and user is given several ways to create a mock data dir with persistence, but ok to remove to avoid defocusing

pubky-homeserver/src/persistence/sql/connection_string.rs
techraed: Well... The string with `postgres` role and password was **_never_** used.

1. That's actually funny, because the original string `postgres://postgres:postgres@localhost:5432/postgres?pubky-test=true` has that attribute `pubky-test=true`, which triggers this check https://github.com/pubky/pubky-core/blob/dfd06dfb8e54917778cce53990800c3684f7dce7/pubky-homeserver/src/app_context.rs#L183-L189 to always result in creating the connection with that string - https://github.com/pubky/pubky-core/blob/dfd06dfb8e54917778cce53990800c3684f7dce7/pubky-homeserver/src/persistence/sql/sql_db.rs#L85 . So I just removed ambiguity
2. I think it's too strict and opinionated to say that everyone has `authenticated local setup`. Users mostly connect to postgres using unix socket with env vars set for `pg_user` and `pg_password`. Not everyone has `postgres` role inside the `postgres` db. Most of times the admin role has the same name as the user name of the OS. For docker cases it can be `posgtres` user and pass, but as I said, they are usually set as env vars, so simple `localhost` is fine. Especially for testing (as it's all under the `cfg(test)` and `testing` feature).

pubky-testnet/src/ephemeral_testnet.rs
techraed: > We should figure out a way to have a single decision point with all `MockDataDirKind` logic in one place. The memory/file system decision is spread across three layers, they should ideally be in-sync at construction time

A 100% right suggestion. That's really how it must be done. Totally agree. However, there's, imo, the elephant in the room - the current codebase uses that style of mutating configs a lot:
- `ConfigToml::default_test_config` - https://github.com/techraed/pubky-core/blob/e6f3d63369d2896f2bf3a40e001d25feb503a965/pubky-homeserver/src/data_directory/config_toml.rs#L210-L224
- `Testnet::create_homesever` - https://github.com/techraed/pubky-core/blob/e6f3d63369d2896f2bf3a40e001d25feb503a965/pubky-testnet/src/testnet.rs#L99-L105
- **_Especially_**, `Testnet::create_homeserver_app_with_mock` - https://github.com/techraed/pubky-core/blob/e6f3d63369d2896f2bf3a40e001d25feb503a965/pubky-testnet/src/testnet.rs#L124-L132
- `EphemeralTestnetBuilder::create_random_homeserver_with_config()` - https://github.com/techraed/pubky-core/blob/e6f3d63369d2896f2bf3a40e001d25feb503a965/pubky-testnet/src/ephemeral_testnet.rs#L317-L320
- and other places.

The solution is to use a builder pattern for the config, which is quite common for configs set-up.

The current implementation is done that way, because fixing the config set-up is completely other issue, out of the scope. As you suggested earlier, changes must be kept as small as possible.

pubky-homeserver/src/data_directory/mock_data_dir.rs
techraed: That's related to this - https://github.com/pubky/pubky-core/pull/361#discussion_r3141945416. Ideally, the function must not receive the config with `InMemory` storage for files. That's an invariant checked by assertion.

However, really better check that with error returning. Done in [393887e](https://github.com/pubky/pubky-core/pull/361/commits/393887ee98b5b498379a2cc8d7c1fe0e07836c1e)

pubky-homeserver/src/persistence/sql/sql_db.rs
techraed: You are right. I wanted to implement something like a defuser inside the drop. I just wanted to create as less as possible new types and abstractions.
Anyway, here's the proper version done - https://github.com/pubky/pubky-core/pull/361/changes/64de682ecdf7b1a1639038c6c7f9504f3ac02d89

e2e/src/tests/storage.rs
techraed: I may be wrong here, but the homeserver can't mint a new user record (the old one is gone with dropped DHT). That is why the test must explicitly call `publish_homeserver_force` with the homeserver public key before signin.

Regarding other `sign*` methods:
- `signup` fails for existing user
- `signin` and `signin_blocking` both first try to establish session by resolving `pubky://<key>/session`, which is pkdns endpoint, but on a fresh dht it doesn't exist

techraed:
Thanks for the review, @86667 .

Most of your comments were accepted and changes were implemented. As you can see, they are mostly for minor changes, that doesn't break abstractions or affect understandability (_except for https://github.com/pubky/pubky-core/pull/361#discussion_r3125008279 - that was a bad decision, I agree. In my own defense, I'd like to say that I tried to touch/create as less as possible abstractions/changes_).

Two comments, however, aren't resolved:
- https://github.com/pubky/pubky-core/pull/361#discussion_r3141920481
- https://github.com/pubky/pubky-core/pull/361#discussion_r3141945416
I hope you get into it, because they point to problems that were either outside of the scope of the issue or just removed ambiguity. I hope after getting into it, you'll change your opinion on changes quality and compromise decisions in the PR.

Best regards!

86667:
Superseded by - https://github.com/pubky/pubky-core/pull/472

Issue #313, opened 2026-02-16
[feature] Add signup token expiry date


SeverinAlexB:
Signup tokens are current valid forever. Tokens should expire though to reduce the chance of brute forcing deprecated tokens.

86667:
[Here](https://github.com/pubky/pubky-core/pull/363/changes#diff-e56136417dc64817d7e01ac8263c717f3da644eb44899e197ef55bb019e413feR36) we allow params to be passed in with a signup token generation request. This should build upon that.

andrei-21:
What exactly do you mean by `deprecated`?

Do I understand correctly that signup codes are created either after a human check or manually? If so, then there are not that many valid codes in a huge space (`256**7 = 72057594037927936`). It is not feasible to brute-force on a server (not locally). If you can check 1000 codes per second, it will take a million years to check a half of them.

SeverinAlexB:
You are correct. Closing

SeverinAlexB:
FYI: We only have 6000 unused signup tokes in the prod database so far.

PR #363, opened 2026-04-01
feat(homeserver): Per user rate limits

86667:
This PR implement per-user rate limits, configurable at signup token generation and modifiable via admin API.

The limits are enforced per user key regardless of whether the request is authenticated.

## features
- Per-user resource limits type and storage: storage quota, max sessions, read bandwidth, write bandwidth, read burst, write burst
- Admin API (`GET/PATCH /users/{pubkey}/quota`) to manage per-user limits at runtime
- Single rate limit middleware extended
- Session count check at `/session`


pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
+# Rate Limits & Per-User Resource Quotas
+
+The homeserver enforces two independent layers of protection:
+
+1. **Global path-based rate limits** — Speed caps (`5mb/s`) throttle individual streams via backpressure; request-count limits (`10r/m`) gate access to sensitive endpoints. Applied per-request, keyed by IP or user.
+2. **Per-user resource quotas** — Bandwidth budgets (`500mb/d`) cap total bytes that can be read from or written to a user's stored resources per time window. Once exhausted, further requests against that user's data are rejected until the window resets.
```
SeverinAlexB: I don't think hard caps are the way to go (hard caps and throttles are rather similar from a configuration POV). This is going to be a horrible user experience. Think about it:
You sign up to a phone data plan. It's not communicated how much data you will have (similar to Bitkit). You try out Pubky with other apps, do some things that are not expected. Everything stops working. You have no idea why.

If we do something like this, we need a clear communication strategy. I suggest we don't do anything like this. We just throttle at `Xmb/s` with good bursts.

pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+{rate}{unit}/{time_unit}
+```
+
+- **rate**: A positive integer.
+- **unit**: `r` (requests), `kb`, `mb`, or `gb`.
+- **time_unit**: `s` (second), `m` (minute), `h` (hour), `d` (day).
+
+Examples: `10r/m`, `1mb/s`, `100gb/d`.
```
SeverinAlexB: These config instructions should life in the config.sample.toml. We will 100% forget to update this doc so I suggest to keep it lean.

pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
...
+---
+
+## Per-User Resource Quotas
+
+Four independent quotas enforced per user's resources (identified by the tenant pubkey in the request hostname):
+
+| Quota | Config field | Purpose |
+|-------|-------------|-----------|
+| Storage quota | `storage_quota_mb` | Maximum total bytes a user can store. |
+| Max sessions | `max_sessions` | Maximum concurrent authenticated sessions. |
```
SeverinAlexB: What's this about?

pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
...
+
+```toml
+[quotas]
+storage_quota_mb = 500        # 500 MB quota. Omit for unlimited.
+max_sessions = 10             # 10 concurrent sessions. Omit for unlimited.
+rate_read = "500mb/d"         # 500 MB read per day. Omit for unlimited.
+rate_write = "100mb/h"        # 100 MB write per hour. Omit for unlimited.
+```
+
+After the one-time database migration, changing these values in config.toml only affects users created afterwards. Existing users keep whatever was written to their DB row.
```
SeverinAlexB: Hmm, do we need this config? Shouldn't just every new user created without a signup-code config by default have no limits?

pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
...
+#### Deploy-time defaults (config.toml)
+
+Set in the `[quotas]` section. These apply to **newly created users** and were backfilled onto existing users during the initial migration:
+
+```toml
+[quotas]
+storage_quota_mb = 500        # 500 MB quota. Omit for unlimited.
+max_sessions = 10             # 10 concurrent sessions. Omit for unlimited.
+rate_read = "500mb/d"         # 500 MB read per day. Omit for unlimited.
+rate_write = "100mb/h"        # 100 MB write per hour. Omit for unlimited.
```
SeverinAlexB: No bursts?

pubky-homeserver/docs/rate-limits.md
```diff
@@ -0,0 +1,109 @@
...
+```json
+{
+  "storage_quota_mb": 500,
+  "max_sessions": 10,
+  "rate_read": "500mb/d",
+  "rate_write": null
+}
+```
+
+To make a user fully unlimited, PUT with all fields set to `null`.
```
SeverinAlexB: This should likely be moved to actual endpoint as a doc string. Maybe we also need an openapi.yaml similar to the homeserver?

SeverinAlexB (changes requested):
No full review yet. Mostly general questions about the approach taken.

pubky-homeserver/docs/rate-limits.md
86667: removed

86667: https://github.com/pubky/pubky-core/issues/369

pubky-homeserver/docs/rate-limits.md
86667: Moved over to throttling :+1:

pubky-homeserver/docs/rate-limits.md
86667: Bursts added

86667:
@SeverinAlexB Ive made some changes:

## Rate limit override semantics

Moved to a per-user rate limit override of the existing path-based rate limit. Rather than applying both, we override the per-path limit with the per-user one. The per-path then acting as a global catch-all rate limit which can be overridden up or down for a particular user.

## All quotas to have same override behaviour

Made all quota configs (`user_storage_quota`, `max_sessions`, `read_rate` and `write_rate`) being overrides.  The default for `user_storage_quota` is the config var, for `max_sessions` is unlimited and for the read/write rates is the per-path rates set in config. 

This i think is the easiest behavioru to understand - else we end up with some quota configs being overrides (like `read_rate` and `write_rate` which must override per-path rates), and others being explicitly set at user creation time (`user_storage_quota`). I preferred the code with this design, its easier to understand and easier to build on. Also less to store and no migrations needed. The simple cases of no limits, or a single limit config for all users is much easier to configure also.

## Admin API PUT -> PATCH

PATCH semantics for setting a user's quota.

pubky-homeserver/docs/rate-limits.md
86667: Added

pubky-homeserver/src/client_server/routes/tenants/write.rs
```diff
@@ -43,16 +40,6 @@ pub async fn put(
...
-    // Check if the size hint exceeds the quota so we can fail early
-    let content_size_hint = body.size_hint().exact();
-    fail_if_size_hint_bigger_than_user_quota(
-        content_size_hint,
-        state.user_quota_bytes,
-        &entry_path,
-        &mut state.sql_db.pool().into(),
-    )
-    .await?;
-
```
86667, line 55: Removed this because:
- Keep storage quota logic in a single place (`persistance/files/opendal`)
- An attacker using this as a DoS could simply remove their `Content-Length` header to avoid this check anyway
- The purpose of the rate limiters are to protect the server from heavy upload/download

pubky-homeserver/src/data_directory/storage_config/storage_config_toml.rs
```diff
@@ -15,3 +15,22 @@ pub enum StorageConfigToml {
...
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct StorageToml {
+    /// Which backend to use (file_system, google_bucket, in_memory).
+    #[serde(flatten)]
+    pub backend: StorageConfigToml,
+
+    /// Default per-user storage quota in MB.
+    /// Omit for unlimited. `0` means zero storage (not unlimited).
+    /// Takes precedence over the deprecated `[general].user_storage_quota_mb`.
+    pub default_quota_mb: Option<u64>,
```
86667, line 35: I did want to avoid breaking the existing `user_storage_quota_mb`, but it would mean keep its `0=unlimited` semantics alongside the other quota's `0=none` semnatics, too confusing. Also with the refactoring in this PR the var can more correctly live in storage config rather than general.

pubky-testnet/src/testnet.rs
```diff
@@ -128,7 +128,7 @@ impl Testnet {
         if !self.dht_relay_urls().is_empty() {
             mock_dir.config_toml.pkdns.dht_relay_nodes = Some(self.dht_relay_urls().to_vec());
         }
-        mock_dir.config_toml.storage = StorageConfigToml::InMemory;
+        mock_dir.config_toml.storage.backend = StorageConfigToml::InMemory;
```
86667, line 131: I think this break is worth is for the benefit of the new `StorageConfig` wrapper. This imo is the direction we should move to split configs by their modules.

e2e/src/tests/admin.rs
```diff
@@ -75,3 +77,440 @@ async fn admin_info_includes_metadata() {
...
+        elapsed_b
+    );
+}
+
+/// Test that per-user max_sessions limit is enforced during signin.
+/// Set max_sessions=2 via admin API, create 2 sessions, verify the 3rd signin
+/// is rejected with 429, then sign out one session and verify a new signin succeeds.
+#[tokio::test]
+#[pubky_testnet::test]
+async fn max_sessions_enforced_via_admin_api() {
```
SeverinAlexB: I am not sure we should limit sessions. What is the thinking behind it? Aren't storage, read/write limits enough?

pubky-homeserver/src/admin_server/routes/generate_signup_token.rs
```diff
@@ -1,15 +1,44 @@
...
+        &mut state.sql_db.pool().into(),
+    )
+    .await?;
     Ok((StatusCode::OK, code.id.0))
 }
+
+/// GET /generate_signup_token — create a token with all-Default limits.
+///
+/// All fields start as `Default` — resolved at enforcement time from system config.
+pub async fn generate_signup_token(State(state): State<AppState>) -> HttpResult<impl IntoResponse> {
```
SeverinAlexB, line 25: To be REST compliant, a `GET` operation should not create a token. To create a new token `POST` should be used.

Isn't this endpoint just a duplicate of `generate_signup_token_with_limits`?

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+    UserRepository::get(&pubkey, &mut state.sql_db.pool().into())
+        .await
+        .map_err(map_user_not_found)
+}
+
+/// GET /users/{pubkey}/quota — return the user's effective limits.
+pub async fn get_user_quota(
+    State(state): State<AppState>,
+    Path(pubkey_str): Path<String>,
+) -> HttpResult<impl IntoResponse> {
```
SeverinAlexB, line 30: Hmmm, can't you do something like this?

```suggestion
pub async fn get_user_quota(
    State(state): State<AppState>,
    Path(pubkey): Path<PublicKey>,
) -> HttpResult<impl IntoResponse> {
```

This is the correct way to do it in axum/tower. I see that you use this pattern in other endpoints as well.

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+/// All fields follow the same semantics:
+/// - absent → keep existing value
+/// - `null` → reset to Default (use system default)
+/// - `"unlimited"` → Unlimited (no limit)
+/// - value → explicit custom limit
+pub async fn patch_user_quota(
+    State(state): State<AppState>,
+    Path(pubkey_str): Path<String>,
+    Json(patch): Json<UserQuotaPatch>,
+) -> HttpResult<impl IntoResponse> {
```
SeverinAlexB, line 50: `PATCH` -> goood

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+            "rate_read": null,
+            "rate_write": null
+        }"#;
+        let config: UserQuota = serde_json::from_str(json).unwrap();
+        // null maps to Default for all fields
+        assert_eq!(config.storage_quota_mb, QuotaOverride::Default);
+        assert_eq!(config.max_sessions, QuotaOverride::Default);
+        assert_eq!(config.rate_read, QuotaOverride::Default);
+        assert_eq!(config.rate_write, QuotaOverride::Default);
+    }
```
SeverinAlexB: I feel like these tests should be moved to the file where the UserQuota struct lives. It has nothing really to do with the route. This is true for more tests in this file.

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+        let url = format!("/users/{}/quota", pubkey.z32());
+
+        // GET defaults (all fields Default)
+        let response = server
+            .get(&url)
+            .add_header("X-Admin-Password", "test")
+            .expect_success()
+            .await;
+        response.assert_status_ok();
+        // Default user should have empty JSON (all fields Default → omitted)
```
SeverinAlexB, line 110: This is interesting. I see why we return an empty json but that also makes the API inconvenient for checking the current user quota.

To provide a good DX, what do you think about this?
- `GET /users/{}/quota` returns the full quota every time (merge default with overrides)
- `GET /users/{}/quota?overrides-only=1` returns the overrides only

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+            "rate_read": "rubbish"
+        });
+        let response = server
+            .patch(&url)
+            .add_header("X-Admin-Password", "test")
+            .content_type("application/json")
+            .bytes(serde_json::to_vec(&body).unwrap().into())
+            .expect_failure()
+            .await;
+        response.assert_status(axum::http::StatusCode::UNPROCESSABLE_ENTITY);
```
SeverinAlexB, line 195: This should be a `401 BAD REQUEST` validation error as far as I know?

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,467 @@
...
+            .get(&url)
+            .add_header("X-Admin-Password", "test")
+            .expect_success()
+            .await;
+        let json: serde_json::Value = response.json();
+        // Unlimited → present as "unlimited"
+        assert_eq!(
+            json["rate_read"], "unlimited",
+            "Unlimited rate field should be 'unlimited' in JSON"
+        );
```
SeverinAlexB, line 255: Idea: Instead of using the term `unlimited` which would always require custom parsing, why not set the value to `i32::MAX` or `i64::MAX`? This would be easier to handle and still be technically unlimited.

Not sure if this works out as smoothly in this case but it's a pattern that is used quite often.

pubky-homeserver/src/admin_server/app_state.rs
```diff
@@ -19,6 +22,8 @@ pub(crate) struct AppState {
     pub(crate) admin_password: String,
     pub(crate) inner_dav_handler: DavHandler,
     pub(crate) metadata: AdminMetadata,
+    /// Shared cache for resolved user limits
+    pub(crate) user_quota_cache: UserQuotaCache,
```
SeverinAlexB, line 23: Hmmm, I like the cache but I am wondering how we will handle cache invalidation when we eventually horizontally scale the homeserver.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -40,13 +85,14 @@ use super::extract_ip::extract_ip;
...
-    /// Create a new rate limiter layer with the given quota.
-    ///
-    /// If quota is None, rate limiting is disabled.
-    pub fn new(limits: Vec<PathLimit>) -> Self {
+    /// Create a new rate limiter layer with the given path limits and per-user
+    /// quota resolution dependencies.
+    pub fn new(limits: Vec<PathLimit>, cache: UserQuotaCache, sql_db: SqlDb) -> Self {
         if limits.is_empty() {
             tracing::info!("Rate limiting is disabled.");
         } else {
```
SeverinAlexB, line 92: Is this log still true? Rate limiting is not really disabled anymore when no config limits are provided?

```suggestion
if limits.is_empty() {
tracing::info!("General rate limiting is disabled.");
} else {
```

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -70,9 +135,39 @@ impl<S> Layer<S> for RateLimiterLayer {
...
+                        break;
+                    };
+                    pool.retain(|_, limiter| {
+                        limiter.retain_recent();
+                        limiter.shrink_to_fit();
+                        !limiter.is_empty()
+                    });
+                }
+            });
+        }
```
SeverinAlexB: NIT: Conflicting inner name

```suggestion
        for pool in [user_read_limiters.clone(), user_write_limiters.clone()] {
            let pool_weak = Arc::downgrade(&pool);
            tokio::spawn(async move {
                let mut interval =
                    tokio::time::interval(Duration::from_secs(CLEANUP_INTERVAL_SECS));
                interval.tick().await; // skip first immediate tick
                loop {
                    interval.tick().await;
                    let Some(inner_pool) = pool_weak.upgrade() else {
                        break;
                    };
                    inner_pool.retain(|_, limiter| {
                        limiter.retain_recent();
                        limiter.shrink_to_fit();
                        !limiter.is_empty()
                    });
                }
            });
        }
```

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -56,7 +102,26 @@ impl RateLimiterLayer {
                 .collect::<Vec<String>>();
             tracing::info!("Rate limits configured: {}", limits_str.join(", "));
         }
-        Self { limits }
+
+        // Spawn a periodic cleanup task for the user quota cache.
+        let cache_weak = Arc::downgrade(&cache);
+        tokio::spawn(async move {
```
SeverinAlexB: NIT: Not sure if necessary but do we need to store the task handle so we can shut it down when the homeserver goes out of scope?

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -70,9 +135,39 @@ impl<S> Layer<S> for RateLimiterLayer {
...
 
+        let user_read_limiters: Arc<DashMap<SpeedLimitKey, Arc<KeyedRateLimiter>>> =
+            Arc::new(DashMap::new());
+        let user_write_limiters: Arc<DashMap<SpeedLimitKey, Arc<KeyedRateLimiter>>> =
+            Arc::new(DashMap::new());
+
+        // Spawn periodic cleanup for per-user limiter pools.
+        for pool in [user_read_limiters.clone(), user_write_limiters.clone()] {
+            let pool_weak = Arc::downgrade(&pool);
+            tokio::spawn(async move {
```
SeverinAlexB: NIT: Same here, do we need to store the handle to shut it down if the homeserver goes out of scope aka cleanup?

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -70,9 +135,39 @@ impl<S> Layer<S> for RateLimiterLayer {
             .map(|path| LimitTuple::new(path.clone()))
             .collect();
 
+        let user_read_limiters: Arc<DashMap<SpeedLimitKey, Arc<KeyedRateLimiter>>> =
+            Arc::new(DashMap::new());
+        let user_write_limiters: Arc<DashMap<SpeedLimitKey, Arc<KeyedRateLimiter>>> =
+            Arc::new(DashMap::new());
+
+        // Spawn periodic cleanup for per-user limiter pools.
+        for pool in [user_read_limiters.clone(), user_write_limiters.clone()] {
```
SeverinAlexB: `Arc<DashMap<SpeedLimitKey, Arc<KeyedRateLimiter>>>` could benefit from it's own struct. Then you can do `limiters.spawn_cleanup()`, `limiters.clean_expired()`,  `limiters.get_or_create()`. This is cleaner than the proposed loose code. This also makes it easier to store the tokio task handle.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -133,18 +228,105 @@ impl LimitTuple {
...
+///
+/// ## Cache capacity behaviour
+///
+/// The cache is bounded by [`MAX_CACHED_USER_QUOTAS`]. When a miss
+/// occurs at capacity, expired entries are evicted first. If the cache is
+/// still full after eviction, 10% of entries are evicted (arbitrary order)
+/// to make room in bulk and avoid per-request churn.
+async fn resolve_limits(
+    pubkey: &PublicKey,
+    cache: &UserQuotaCache,
```
SeverinAlexB: Hmmm same here. Why not make UserQuotaCache into a proper struct that holds the SqlDb. This way, you can ask `cache.resolve_limits(pubkey)` or `cache.spawn_cleanup_task()`. UserQuotaCache always sticks to the SqlDb which fulfils the [Data Clumps](https://refactoring.guru/smells/data-clumps) smell.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -262,87 +439,169 @@ where
...
             };
 
-            if limit.limit.is_whitelisted(&key) {
-                continue;
+            // Collect (key, limiter) pairs for download throttling in phase 5.
+            let mut download_throttlers: Vec<(LimitKey, Arc<KeyedRateLimiter>)> = Vec::new();
+
+            // ── Phase 1: Request-count limits ──
+            for limit in &path_limits_owned {
+                if limit.limit.quota.rate_unit != RateUnit::Request {
```
SeverinAlexB: This method is getting huge and would benefit from a refactor

https://refactoring.guru/smells/comments
https://refactoring.guru/smells/long-method

pubky-homeserver/src/client_server/routes/tenants/write.rs
SeverinAlexB: Hmmm, this was not intended to be a fool-proof check but to make the upload fail early. For example: I have a quota of 1GB and I upload 2 GBs, it should fail instantly and not only after 15min after I uploaded 1GB. Does this make sense?

Happy if you want to move this to another file but it did serve a purpose.

pubky-homeserver/src/client_server/routes/auth.rs
```diff
@@ -33,6 +33,32 @@ use tower_cookies::{
...
+    tx.commit().await?;
+
+    // Populate cache so the rate limiter sees the new user immediately
+    // (evicts any negative cache entry from pre-signup lookups).
+    state
+        .user_quota_cache
+        .insert(public_key.clone(), CachedUserQuota::new(user.quota()));
+
+    Ok(user)
+}
```
SeverinAlexB, line 60: Feels like this needs to go to a `UserService` struct or so. This should not be part of a route aka presentation layer.

pubky-homeserver/src/data_directory/config_toml.rs
```diff
@@ -85,6 +87,14 @@ pub struct AdminToml {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
 pub struct GeneralToml {
     pub signup_mode: SignupMode,
+    /// Deprecated: use `[storage].default_quota_mb` instead.
+    /// Kept for backwards compatibility: `0` means unlimited.
+    /// Ignored when `[storage].default_quota_mb` is set.
+    #[serde(default)]
```
SeverinAlexB: Config defaults are defined in `config.default.toml`, not with serde.

pubky-homeserver/src/persistence/sql/entities/signup_code.rs
```diff
@@ -191,6 +228,32 @@ pub struct SignupCodeEntity {
...
+    /// Per-user max sessions. `None` = Default (resolved from system config at enforcement time).
+    pub quota_max_sessions: Option<i32>,
+    /// Per-user read rate limit. `None` = Default (resolved from system config at enforcement time).
+    pub quota_rate_read: Option<String>,
+    /// Per-user write rate limit. `None` = Default (resolved from system config at enforcement time).
+    pub quota_rate_write: Option<String>,
+    /// Per-user read rate burst override. `None` = default (burst = rate).
+    pub quota_rate_read_burst: Option<i32>,
+    /// Per-user write rate burst override. `None` = default (burst = rate).
+    pub quota_rate_write_burst: Option<i32>,
```
SeverinAlexB, line 236: These values should probably be `u32`/`u64` as negative rate limits do not make any sense?

pubky-homeserver/src/persistence/files/user_quota_layer.rs
```diff
@@ -173,39 +206,39 @@ impl<R: oio::Write, A: Access> oio::Write for WriterWrapper<R, A> {
...
+        // Lock the user row to serialize concurrent writes and prevent quota bypass.
+        let mut user: UserEntity =
+            UserRepository::get_for_update(self.entry_path.pubkey(), &mut (&mut tx).into())
+                .await
+                .map_err(|e| {
+                    opendal::Error::new(
+                        opendal::ErrorKind::Unexpected,
+                        format!("Failed to get user {}: {}", self.entry_path.pubkey(), e),
+                    )
+                })?;
```
SeverinAlexB, line 224: Why do you lock if you have a transaction already? I recommend reading up on Postgres locks that are applied automatically without you actively invoking them. I am not 100% sure but I think this should be covered by the PG row log already.

https://claude.ai/share/a4cd0754-d8c8-486f-a00f-f391a111d97b

pubky-homeserver/src/persistence/sql/entities/user.rs
```diff
@@ -166,6 +210,78 @@ impl UserRepository {
...
+    }
+
+    /// Atomically apply a partial quota update to a user.
+    ///
+    /// Opens a transaction, locks the row with `FOR UPDATE`, merges the patch
+    /// into the current quota, and writes the result back.
+    pub async fn patch_quota(
+        public_key: &PublicKey,
+        patch: &UserQuotaPatch,
+        pool: &sqlx::PgPool,
```
SeverinAlexB, line 264: Should be a `executor: &mut UnifiedExecutor<'a>,`?

pubky-homeserver/src/persistence/sql/entities/user.rs
```diff
@@ -203,6 +346,34 @@ pub struct UserEntity {
...
+    /// Per-user max sessions. `None` = Default (resolved from system config at enforcement time).
+    pub quota_max_sessions: Option<i32>,
+    /// Per-user read rate limit. `None` = Default (resolved from system config at enforcement time).
+    pub quota_rate_read: Option<String>,
+    /// Per-user write rate limit. `None` = Default (resolved from system config at enforcement time).
+    pub quota_rate_write: Option<String>,
+    /// Per-user read rate burst override. `None` = default (burst = rate).
+    pub quota_rate_read_burst: Option<i32>,
+    /// Per-user write rate burst override. `None` = default (burst = rate).
+    pub quota_rate_write_burst: Option<i32>,
```
SeverinAlexB, line 350: Should be unsigned?

pubky-homeserver/src/persistence/user_quota.rs
```diff
@@ -0,0 +1,1044 @@
...
+pub struct CachedUserQuota {
+    /// The resolved limit configuration, or `None` for a negative (user-not-found) entry.
+    pub config: Option<UserQuota>,
+    cached_at: Instant,
+    ttl: Duration,
+}
+
+impl CachedUserQuota {
+    /// Wrap a resolved config with a fresh timestamp.
+    pub fn new(config: UserQuota) -> Self {
```
SeverinAlexB, line 187: In theory, a cache is not part of persistence. It should be in a service layer

pubky-homeserver/src/persistence/user_quota.rs
```diff
@@ -0,0 +1,1044 @@
+use std::sync::Arc;
```
SeverinAlexB, line 1: This file should be moved to `src/persistence/sql` as it persists to postgres

pubky-homeserver/src/persistence/user_quota.rs
```diff
@@ -0,0 +1,1044 @@
...
+
+    /// Validate that rate fields survive a Display → FromStr roundtrip,
+    /// and that burst fields are only set when a corresponding rate is present.
+    pub fn validate_rate_roundtrips(&self) -> Result<(), String> {
+        validate_rate_value("rate_read", &self.rate_read)?;
+        validate_rate_value("rate_write", &self.rate_write)?;
+        validate_burst("rate_read_burst", self.rate_read_burst, &self.rate_read)?;
+        validate_burst("rate_write_burst", self.rate_write_burst, &self.rate_write)?;
+        Ok(())
+    }
```
SeverinAlexB, line 445: This doc string and method name is confusing. It should survive a Display -> FromStr? Why? What is the purpose behind this? Is this a `is_valid()` method or what is this doing?

SeverinAlexB (changes requested):
Solid work. I mostly have clean code concerns. I will need to do a second round later because it's so much code.

pubky-homeserver/src/admin_server/routes/generate_signup_token.rs
86667: This is the existing endpoint which i havent changed.  The new one is a POST. 
I suggest we consider deprecating this GET endpoint as part of [this work](https://github.com/pubky/pubky-core/issues/313) which will touch these endpoint also?

e2e/src/tests/admin.rs
86667: The thinking was that for per-ip rate limits a user could get around speed limits with many sessions from different ips and write. Now that we're per-key, I think youre right. removed.

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: My bad. fixed

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: I guess you mean `400 BAD REQUEST` and not `401 UNAUTHORIZED`.

I do think that 422 is more accurate that 400 here.

400 - https://http.cat/status/400
- malformed request **syntax**

422 - https://http.cat/status/422
- the syntax of the request entity is correct, but it was unable to process the contained instructions

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: This follows the pattern of background task being tied to a shared resource lifetime via taking a `Weak` refernce to the `Arc` and checking if the `Arc` is dropped each loop via `cache_weak.upgrade()`. 

I quite like it. No need for cancellation tokens, implementing `Drop` or an `.abort()` from outside of the task.

86667: Added a comment to make it clearer what its doing

86667: Happy to discuss this further if you have thoughts - its an interesting one.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: Nice idea

86667: Nice idea - fixed.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: done

pubky-homeserver/src/client_server/routes/auth.rs
86667: We followed the pattern of the file here, lets leave as is in order to keep this PR changes down. Ive marked to be done in an upcoming refactor which I'll do once this and your auth work are in

pubky-homeserver/src/admin_server/app_state.rs
86667: True, there is a 5 min TTL, perhaps its enough to wait for that to expire? We could also reduce it maybe to 2 mins. Removing the cache invalidation would simplify the code too.

What do you think, is the TTL enough? How much damage could someone really do it 2 mins

pubky-homeserver/src/client_server/routes/auth.rs
86667: Forget that. A `UserService` was useful for resolving a few other problems too:
- How to pass around `default_storage_quota` cleanly
- Having a single owner of checking storage quota logic
- Not accessing `UserRepository` and `UserEntity` in business logic

Implemented and used throughout. Definately further cleanup which can be done on top of this at some point too.

pubky-homeserver/src/persistence/sql/entities/user.rs
86667: This is the user entity which matches the postgres types. At this level, `-1` means unlimited for int types, the domain level types are unsigned: `QuotaOverride<u64>`

pubky-homeserver/src/persistence/sql/entities/user.rs
86667: Ended up removing this fn and instead use existing `UserRepository::get_for_update` and `UserRepository::set_quota`

pubky-homeserver/src/persistence/sql/user_quota.rs
```diff
@@ -0,0 +1,1044 @@
...
+
+    /// Validate that rate fields survive a Display → FromStr roundtrip,
+    /// and that burst fields are only set when a corresponding rate is present.
+    pub fn validate_rate_roundtrips(&self) -> Result<(), String> {
+        validate_rate_value("rate_read", &self.rate_read)?;
+        validate_rate_value("rate_write", &self.rate_write)?;
+        validate_burst("rate_read_burst", self.rate_read_burst, &self.rate_read)?;
+        validate_burst("rate_write_burst", self.rate_write_burst, &self.rate_write)?;
+        Ok(())
+    }
```
86667: Fair. Renamed and re-doc'd:

```
    /// Check that the quota fields are internally consistent:
    /// - Rate values can be persisted (fit in the DB column).
    /// - Burst overrides have a corresponding rate `Value`.
    /// - Burst values are > 0.
    pub fn validate(&self) -> Result<(), String> {
        validate_rate_value("rate_read", &self.rate_read)?;
        validate_rate_value("rate_write", &self.rate_write)?;
        validate_burst("rate_read_burst", self.rate_read_burst, &self.rate_read)?;
        validate_burst("rate_write_burst", self.rate_write_burst, &self.rate_write)?;
        Ok(())
    }
```

pubky-homeserver/src/client_server/routes/tenants/write.rs
86667: Added back with exceeding quota check logic centralised into UserService

pubky-homeserver/src/persistence/sql/user_quota.rs
```diff
@@ -0,0 +1,1044 @@
...
+pub struct CachedUserQuota {
+    /// The resolved limit configuration, or `None` for a negative (user-not-found) entry.
+    pub config: Option<UserQuota>,
+    cached_at: Instant,
+    ttl: Duration,
+}
+
+impl CachedUserQuota {
+    /// Wrap a resolved config with a fresh timestamp.
+    pub fn new(config: UserQuota) -> Self {
```
86667: Moved.

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: The problem is that only `storage_quota_md` has a global default which can be merged, the rate limit's defaults are per-path so cannot be resolved here.

We could return the `QuotaOverride` value:
```
pub enum QuotaOverride<T> {
    /// No override — the system-wide default applies.
    Default,
    /// Explicitly bypass the limit for this user.
    Unlimited,
    /// A custom limit for this user.
    Value(T),
}
```
Which wouldnt give the actual `storage_quota_md` value if the user was using the global default. Merging for `storage_quota_md` but not for rate limits in the same response feels a bit wrong?

Or, how about a separate 
`GET /users/{}/quota/storage_quota_mb`
Which only returns the resolved value?

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: 1. No natural max representation for `BandwidthRate` types (eg `"10mb/s"`)
2. MAX values do not remove the need for custom parsing - we'd still want to distinguish between "admin set to unlimited, do not apply any rate limit" and "very large rate set, apply rate" for `BandwidthRate` because governor cannot handle very large values, so we'd still need a special string for the rate limits.
3. I prefer to keep the same pattern for all quota vars
4. Also, added bonus of not doing quota work when unlimited.

pubky-homeserver/src/persistence/sql/entities/signup_code.rs
86667: This is the user entity which matches the postgres types. At this level, -1 means unlimited for int types, the domain level types are unsigned: QuotaOverride<u64>

pubky-homeserver/src/admin_server/routes/user_quota.rs
SeverinAlexB: It is the question if we even need the speed rate limits per path anymore at all, no? We do need speed limits for  unauthenticated users but this doesn't need to be per path, right? Could be in the config too?

pubky-homeserver/src/admin_server/routes/user_quota.rs
SeverinAlexB: You are correct

pubky-homeserver/src/admin_server/app_state.rs
SeverinAlexB: 2min and letting is expire sounds like a good choice 👍

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
SeverinAlexB: I have never encountered weak arc references so hard for me to judge. Pretty sure it's all good though

pubky-homeserver/src/admin_server/app_state.rs
86667: We do have the `disable_user()` endpoint for emergencies, too. 
Updated to be 2 mins and remove cache invalidation from `patch_quota()`.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: Done

pubky-homeserver/src/data_directory/config_toml.rs
86667: Fixed

pubky-homeserver/src/persistence/files/user_quota_layer.rs
86667: Not sure i understand what you mean, the transaction and PGs row-lock dont help with concurrent atomic read-then-write:

We need to ensure that the value we _read_ is not updated by another process before we've done our _write_. 
Put another way, if another instance of this code is currently running then i should not read the current value yet because it may be updated by the other instance. Only once the other has committed should i read. Thats what the `FOR UPDATE` is for, do not `SELECT` until the lock is free.

Without the lock on `SELECT`, I would read and await the free lock before `UPDATE`, by which time the other process may have updated the value I read, then my update will overwrite theirs. 

Does that make sense or did i miss the point?

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: Okay here's what we could do, which would simplify somewhat:

## Endpoint protection (request-count, IP-keyed only)

Keep per-path config but only for request-count limits and key per-ip. This is useful for eg `/signup` to mitigate signup token brute force. 

It is path-specific, which does makes sense for per-request. For general reading/writing per-request limit is not super useful, bandwidth limit is.

I dont have another solid example of where this may be applied, but the implementation exists, is functional and may become more useful in the future. Im thinking about migration endpoints, user account config endpoints (?), who knows.

Edit: Just thought, we could still add a general per-ip request count limit on all writes, eg:
```
[[drive.rate_limits]]
path = "/pub/**"
method = "PUT"
quota = "100r/m"
key = "ip"
```

To mitigate an attacker using resources but writing a very small amount to avoid bandwidth limits.   

Perhaps one on read, too. Because why not. 

## User bandwidth (global, user-keyed)

Then for read/write endpoints we have a distinct rate limit enforcer with its own set of system defaults, which is applied per-user. 

This means that were being a little more explicit in changing rate limits from being per-ip to being per-resource (user path). Bandwidth limits are concerned only with the _resource_ being read/written, regardless of _who_ acting on it.


## Tradeoff

This distinction between per-request and per-resource is much easier to understand and implement. It also allows us to resolve a user's effective limit across all read/writes at any time - again easier to understand and implement.

We do lose the control over setting specific bandwidth limits for different paths, though this was only possible if the user didnt have any overrides set anyway. Plus, is that even useful? A read is a read, it doesnt matter where it was from.

pubky-homeserver/src/admin_server/routes/user_quota.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// GET /users/{pubkey}/quota — return the user's effective limits.
+pub async fn get_user_quota(
+    State(state): State<AppState>,
+    Path(pubkey): Path<Z32Pubkey>,
+) -> HttpResult<impl IntoResponse> {
+    let user = state
+        .user_service
+        .get_or_http_error(&pubkey.0, false)
+        .await?;
```
SeverinAlexB, line 38: In theory, a service should not know anything about the presentation layer. The user service should not handle any http related things. It having `http` in it's method name is a smell.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -6,31 +6,61 @@
...
 //! So we implement our own rate limiter here.
 //!
+//! # Two sources of limits
+//!
+//! 1. **Path limits** (`[[drive.rate_limits]]` in config) — matched by request
+//!    path and method. Only request-count quotas are supported here.
+//! 2. **Bandwidth config** (`[default_quotas]`):
+//!    - `rate_read` / `rate_write` — defaults for authenticated users,
+//!      with per-user DB overrides via `UserQuota`.
+//!    - `unauthenticated_ip_rate_read` — for unauthenticated IP reads.
```
SeverinAlexB: NIT: `RateLimiterLayer` would probably benefit from being split up. Your comment here is basically describing that it does 2 things:
- `RequestRateLimitLayer` doing request count quotas.
-  `BandwithRateLimitLayer` doing speed quotas.

This is quite nitpicky of me but the comment basically screams to me that there is a natural split already :) But I also don't know how much work this would be to split up so take it with a grain of salt.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -40,208 +70,269 @@ use super::extract_ip::extract_ip;
...
-    fn throttle_upload(
-        req: Request<Body>,
-        key: &LimitKey,
-        limiter: &Arc<RateLimiter<LimitKey, DashMapStateStore<LimitKey>, QuantaClock>>,
-    ) -> Request<Body> {
-        let (parts, body) = req.into_parts();
-        let new_body = Self::throttle_body(body, key, limiter);
-        Request::from_parts(parts, new_body)
+    /// Returns `true` when there is *something* to check for this request.
+    fn has_work(&self, req: &Request<Body>) -> bool {
```
SeverinAlexB:
```suggestion
    fn should_limit(&self, req: &Request<Body>) -> bool {
```

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
```diff
@@ -40,208 +70,269 @@ use super::extract_ip::extract_ip;
...
+                            self.defaults.rate_write.as_ref(),
+                            self.defaults.rate_write_burst,
+                            quota.rate_write_burst,
+                        )
+                    } else {
+                        (
+                            self.defaults.rate_read.as_ref(),
+                            self.defaults.rate_read_burst,
+                            quota.rate_read_burst,
+                        )
```
SeverinAlexB: As far as I can read, this method only every returns one throttler, is this correct? If yes, the throttlers vec is not needed?

---

There are a lot of if's in if's in if's in this method. I suggest to use early returns. This reduces the number of nesting and makes the method easier to read.

Example:

```rust
if !info.authenticated {
  self.push_ip_throttler(&info.client_ip, &mut throttlers);
  return Ok(throttlers);
}

let ref pubkey = match info.user_pubkey {
  Some(pubkey) => pubkey,
  None =>  {
    self.push_ip_throttler(&info.client_ip, &mut throttlers);
    return Ok(throttlers);
  }
}
```


---


Also long method smell https://refactoring.guru/smells/long-method

pubky-homeserver/src/data_directory/quota_config/path_limit.rs
```diff
@@ -57,6 +57,18 @@ impl PathLimit {
...
+        //   rate_read, rate_write, unauthenticated_ip_rate_read
+        if self.quota.rate_unit != RateUnit::Request {
+            return Err(anyhow::anyhow!(
+                "Path limit '{}' uses bandwidth quota '{}', but path limits only support \
+                 request-count quotas (e.g. \"10r/m\"). Use [default_quotas] for bandwidth \
+                 throttling (rate_read, rate_write, unauthenticated_ip_rate_read).",
+                self.path.0,
+                self.quota
+            ));
+        }
```
SeverinAlexB: This whole check might not be needed anymore if we split up the bandwidth/request count layer because then the two quota types can be split up as well.

pubky-homeserver/src/data_directory/config.default.toml
```diff
@@ -9,14 +9,45 @@ icann_listen_socket = "127.0.0.1:6286"
...
+# Per-user defaults (rate_read, rate_write) are the system-wide
+# fallback values used when a user's quota is "Default" (NULL in DB).
+# Per-user overrides are managed via the admin API:
+#   PATCH /users/{pubkey}/quota
+#
+# Default bandwidth limit for user reads / downloads.
+rate_read = "10mb/s"
+#
+# Default bandwidth limit for user writes / uploads.
+rate_write = "5mb/s"
```
SeverinAlexB: As far as I know, you can't set a value in toml to null. Therefore, setting default bandwidth limits can only be overriden by a high limit and never set to unlimited. I would not set default limits anyway.

Also: There is a lot of comments in here. Are you sure you want them here and not in `config.sample.toml`? You probably confused the sample with the default?

pubky-homeserver/src/persistence/sql/migrations/m20260327_add_quota_columns.rs
```diff
@@ -0,0 +1,210 @@
...
+/// default from config is resolved at enforcement time.
+pub struct M20260327AddQuotaColumnsMigration;
+
+/// Add the four limit columns to the given table.
+async fn add_quota_columns(
+    tx: &mut Transaction<'static, sqlx::Postgres>,
+    table: &str,
+) -> anyhow::Result<()> {
+    for (col, typ) in [
+        ("quota_storage_mb", "BIGINT"),
```
SeverinAlexB: `INTEGER` in postgres is +-2,147,483,648 = 2.1 billion megabytes = 2.1 petabytes. I don't think a BIGINT is needed.
```suggestion
        ("quota_storage_mb", "INTEGER"),
```

pubky-homeserver/src/services/user_service/quota_cache.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/// TTL cache for per-user quota overrides.
+///
+/// Stores resolved quotas with automatic expiry and background cleanup.
+/// Does not access the database — cache miss resolution is handled by
+/// [`super::UserService::resolve_quota`].
+#[derive(Debug, Clone)]
+pub(super) struct QuotaCache {
+    entries: Arc<DashMap<PublicKey, CachedEntry>>,
```
SeverinAlexB, line 71: NIT: Instead of building your own cache, why not use https://crates.io/crates/lru/? Probably simpler? Would also prevent a cache overflow.

Also curious: You use DashMap here. What advantage does Arc<DashMap> have over a regular HashMap? Doesn't Arc destroy the DashMap advantage? I think you are using Arc<DashMap> also in the RateLimiter.

SeverinAlexB (changes requested):
Some things but close to being mergeable!

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: I went for a single endpoint `/users/{pubkey}/quota` which returns eg:
```
{
  "effective": { "storage_quota_mb": 500, "rate_read": "10mb/s", "rate_write": "5mb/s" },
  "overrides": { "storage_quota_mb": 500 }
}
```

Firstly, to match the pattern of other endpoints in the codebase which opt for path-based params rather than query params. Secondly to be explicit to the admin immediately instead of relying on them understanding how rate limits work, and relying on them knowing that the `override-only` option exists.

pubky-homeserver/src/admin_server/routes/user_quota.rs
86667: Agreed, this service needs some love. 
For now, Ive copy-pasted the functions which should clearly live in here in order to reduce the amount of other code I need to touch. Will fix in a later refactor.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: Yes good point I'll do this. It'll make each layer more digestible for readers.

pubky-homeserver/src/client_server/layers/rate_limiter/layer.rs
86667: > As far as I can read, this method only every returns one throttler, is this correct? If yes, the throttlers vec is not needed?

True. `Vec` was left over from a previous iteration.

Done early return refactor.

pubky-homeserver/src/data_directory/quota_config/path_limit.rs
86667: Split up the types as a part of the Layer split refactor :+1:

pubky-homeserver/src/data_directory/config.default.toml
86667: > you can't set a value in toml to null.

Removing the key would mean the default is no quota, ie `unlimited`.

> setting default bandwidth limits can only be overridden by a high limit and never set to unlimited. 

A user override can be set to `unlimited`.

The rule is always:
- Read from user table.
  - If rate set, use it (can be Value or `unlimited`).
  - If rate not set, use default rate from config. Config not set means no rate -> `unlimited`

I'll remove the defaults from here though.

> There is a lot of comments in here

Removed comments

pubky-homeserver/src/services/user_service/quota_cache.rs
86667: Yes possibly, this was built off of what existed in the rate limit layer before. This works, ive created an issue to come back and consider changing.

`Arc` is used here to make `QuotaCache` `Clone`. Also, its used to hold the `Weak` references which we use to detect when the `Arc` has no more owners and thus the cleanup task can be terminated.

pubky-homeserver/src/data_directory/config.default.toml
SeverinAlexB: Cool thx. One thing:

Should we set these values in the sample conf or should we comment them?

```toml
# Default bandwidth limit for user reads (downloads).
rate_read = "10mb/s"
# Default bandwidth limit for user writes (uploads).
rate_write = "5mb/s"
# Bandwidth limit for unauthenticated IP reads (downloads).
# Not per-user — applies to all anonymous requests by IP.
unauthenticated_ip_rate_read = "1mb/s"
```

I am thinking from a personal homeserver POV that I install at home on my Umbrel for example. The sample config will be written to the `~/.pubky/config.toml` directly so any user that runs a new homeserver will have these values set if not changed otherwise. Most users will likely not change them. So should we really limit the speed here? My gut says no because it's going to be a source of confusion. We should comment these values in the sample.config.toml. What do you think?

SeverinAlexB (approved):

PR #366, opened 2026-04-06
Feat: Use http scheme not only for localhost but for ip address also

SpontaneousOverthrow:
Problem: Not it is not possible to run testnet on vms, because if you set the `icann_domain` to vm's ip the `https` scheme will be used.

Currently the the `http` scheme is used only for hardcoded `localhost`, this is not flexible. 

I suggest to also to check if domain is a valid IP address and if so apply the `http` scheme as a quick fix.

I think the proper solution would be - use flag for `homeserver` like `icann_protocol` this flag will be published to `pkarr` like it's done now with `public_icann_http_port`. And get this value on resolve, without any hardcoded values.
 I didn't implement this solution, because not sure if it's the right one.
So, @SeverinAlexB @86667 @SHAcollision  wanna hear your ideas on that topic, and implement the one you decided the best.

### DIFF
--- a/pubky-common/src/constants.rs
+++ b/pubky-common/src/constants.rs
@@ -15,7 +15,7 @@ pub mod reserved_param_keys {
 ///
 /// These hosts need an explicit [`reserved_param_keys::HTTP_PORT`] in their
 /// SVCB record so clients know which port to connect to over HTTP.
-/// 
+///
 /// TODO: This should live in pkarr crate with all other SVCB/HTTPS record logic.
 pub fn requires_http_port(domain: &str) -> bool {
     domain == "localhost" || domain.parse::<IpAddr>().is_ok()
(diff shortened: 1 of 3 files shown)


SHAcollision:
Hey @SpontaneousOverthrow  this could work. I pushed a small fix. I also think a better solution would be to have a new (optional?) config.toml field for `icann_protocol` . I see how this change could disable legit https for IPs and also have negative consequences. Thoughts @SeverinAlexB ?

SpontaneousOverthrow:
thanks, yeah, let's wait for @SeverinAlexB response and if he agrees on new config filed, will implement that solution.

SeverinAlexB:
Happy to go with whatever you think is worth doing. Just be aware this is likely a temporary fix because we should redo the way we do DNS resolution in the future https://github.com/pubky/pubky-core-braindump-jan2026/pull/7

PR #368, opened 2026-04-07
fix: Improve embedded-postgres shared instance handling to prevent hanging postgres bug

86667:
The embedded-postgres feature uses an in-process postgres instance. The previous implementation (`postgresql_embedded` crate) downloaded postgres binaries from github and ran them as a child process. However, Rust doesn't run Drop for statics at process exit, and child processes spawned by `postgresql_embedded` survived after the test runner exited, as identified by @ok300 [here in slack](https://synonymworkspace.slack.com/archives/C0ACPQKE5MY/p1774960145374609). 

## The light-weight approach (no docker dependency)

To fix the orphan problem, the first solution manually built cleanup machinery with unsafe code to register `atexit` handler, `SIGINT/SIGTERM handlers`, all sorts of low level code which we really dont want to be dealing with. See commit `0d7e3c85` for this code.

## testcontainers (docker dependency)
With testcontainers docker owns the postgres process lifecycle entirely and so handles the process kill when dropped. On normal exit we run a task to cleanup the containers.

## Tradeoff
This does however introduces the requirement of Docker running on the host. This is already common for development environments, and the feature remains opt-in behind the embedded-postgres flag.


pubky-testnet/src/embedded_postgres.rs
```diff
@@ -46,95 +40,78 @@ use std::time::Duration;
...
+        })
     }
 
     /// Get the connection string for this embedded PostgreSQL instance.
     pub fn connection_string(&self) -> anyhow::Result<ConnectionString> {
-        let settings = self.pg.settings();
         let url = format!(
-            "postgres://{}:{}@{}:{}/{}",
-            settings.username, settings.password, settings.host, settings.port, self.database_name
+            "postgres://postgres:postgres@{}:{}/postgres",
```
86667: Match Postgres docker image defautls

pubky-testnet/Cargo.toml
```diff
@@ -35,12 +35,12 @@ mainline = { workspace = true }
...
+testcontainers = { version = "0.27", features = ["watchdog"], optional = true }
+testcontainers-modules = { version = "0.15", features = ["postgres"], optional = true }
 rand = { version = "0.9", optional = true }
 
+[dev-dependencies]
+
 [features]
 default = []
-embedded-postgres = ["postgresql_embedded", "postgresql_commands", "postgresql_archive", "rand"]
+embedded-postgres = ["testcontainers", "testcontainers-modules", "rand"]
```
andrei-21: I guess the feature should be named differently now? Maybe `docker-postgres`.

86667: True. Renamed.

86667:
I have added a cleanup task for normal exit also. We had handled `SIGINT` and `SIGTERM` but not normal ending of the process.

andrei-21 (approved):

andrei-21 (approved):

PR #370, opened 2026-04-20
fix(auth): Check actual version byte

andrei-21:


docs/AUTH.md
```diff
@@ -79,7 +79,7 @@ AuthToken   = signature namespace version timestamp pubky capabilities
 
 signature      = 64OCTET ; ed25519 signature over the rest of the token
 namespace      = %x50.55.42.4b.59.3a.41.55.54.48 ; "PUBKY:AUTH" in UTF-8 (10 bytes)
-version        = 1*OCTET ; Version of the AuthToken for future proofing
+version        = OCTET ; Version of the AuthToken for future proofing
```
andrei-21, line 82: `1*OCTET` means **one or more**, I guess the intention was **exactly one**.

SeverinAlexB (approved):
Good catch.

pubky-common/src/auth.rs
```diff
@@ -381,4 +381,20 @@ mod tests {
...
         assert_eq!(guard.seen[0], recent_id);
     }
+
+    #[test]
+    fn unknown_version() {
+        let signer = Keypair::random();
+        let token = AuthToken {
+            signature: Signature::from_bytes(&[0; 64]),
+            namespace: *PUBKY_AUTH,
+            version: 1,
```
86667, line 391:
```suggestion
            version: CURRENT_VERSION+1,
```

86667: I got here late - no need to fix

PR #371, opened 2026-04-21
fix(sdk): lock wasm-pack install so that it uses the exact dependency versions from wasm-pack's lockfile

86667:


SeverinAlexB (approved):
Nice

andrei-21:
Right. Maybe it makes sense to add to other places?

andrei-21 (approved):

Issue #237, opened 2025-09-12
Homeserver webdav


dzdidi:
Implement webdav compatible routes on homeserver. 

This will allow users to use available webdav compatible clients like https://cyberduck.io/ to access and explore their files on their homeserver.

This is already implemetend for admin server of the homeserver with HTTP Basic authentication

Initial change should ideally be compatible with current authentication scheme. As possible temporal work around would be to put `session_id` into http authorization header for compatibility

SeverinAlexB:
Update: the client_server webdav endpoints are not standard compliant. Neither is the auth system. This is not a high priority task but to make it work, this would need to be tackled.
https://github.com/pubky/pubky-core/pull/373#discussion_r3180957005

PR #373, opened 2026-04-22
feat(hs/sdk/common)!: Grant-based auth

SeverinAlexB:
Introduces a new Grant (JWS) + Proof-of-Possession (PoP) authentication flow alongside the existing cookie auth, and reorganizes the auth code in both pubky-homeserver and pubky-sdk into clean, self-contained modules. Cookie auth is preserved for backwards compatibility but marked deprecated; Grant auth is the new default for long-lived, mirror-friendly, per-app sessions.

## Design

| Token | Signer | Lifetime | Purpose |
|-------|--------|----------|---------|
| **Grant JWS** (`pubky-grant`) | User keypair (Ed25519 / EdDSA) | Long-lived | Delegates scoped capabilities to a client app, bound to the client key via `cnf` |
| **PoP JWS** (`pubky-pop`) | Client keypair | ±3 min | Proves possession of the key in Grant `cnf`; `aud` is the target homeserver pubkey |
| **Session bearer** | 32 bytes from OsRng, SHA-256 hashed at rest | 1 hour | Opaque `Authorization: Bearer` token; carries no claims |

The bearer is an opaque pointer — capabilities and metadata live server-side, keyed by `SHA-256(bearer)`. A DB leak cannot yield usable bearers, and revocation is instant. PoP nonces are tracked per Grant for replay protection (unique constraint, GC after 360 s).

## Proposal Difference

This PR is based on the [JWT proposal](https://github.com/pubky/pubky-core-braindump-jan2026/blob/master/proposals/jwt-session/proposal-v4-pop.md). Two key differences emerged while developing:
- The proposal assumes two concurrently active access tokens per grant. This PR reduced it to one concurrently active access token per grant. From a security POV, this is better.
- The proposal assumes a JWT access token. This PR replaced the token with an opaque bearer (basically a secret). Having a JWT access token does not add any obvious advantage.

### `pubky-homeserver`
- New `client_server/auth/` module fully split into `cookie/` (deprecated) and `jwt/` sub-modules, each with its own `middleware`, `persistence`, `routes`, and crypto.
- Middleware split into `authentication` (Bearer/Cookie) and `authorization` (`WriteAccess`) layers; session extractor enforces auth at the type level (used to make `signout` idempotent).
- New routes: `POST /auth/jwt/session`, `GET /auth/jwt/sessions`, `DELETE /auth/jwt/session/{grant_id}`.
- New `grant_sessions` migration;
- `AuthVerifier` moved from `pubky-common` to the homeserver (it's server-side logic).

### `pubky-common`
- New `auth/grant.rs`, `auth/grant_session.rs`, `auth/jws.rs` (shared claims types between Ring, SDK, and homeserver).
- `auth.rs` split into `auth/auth_token.rs` (legacy) + new modules above.
- `Capabilities` builder extended.

### `pubky-sdk`
- `actors/auth/` reorganized into `cookie/` (deprecated) and `jwt/` sub-trees, each with `builder`, `credential`, `flow`, `approval`, `view`.
- New public API: `Pubky::start_jwt_auth_flow(..)`, `PubkyJwtAuthFlow`, `JwtCredential`, `JwtSessionView`, `ClientId`.
- `await_credential()` returns a raw `JwtCredential` / `CookieCredential` that can be persisted, inspected, or later lifted into a session via `PubkySession::from_{jwt,cookie}_credential`.
- Deep links split into `signin_jwt` / `signup_jwt` variants.
- Relay listener and channels moved under `actors/auth/relay/`.
- `SessionInfo` no longer wrapped in `Arc<RwLock<_>>`; credential trait refactored.

### Tests
- New homeserver unit tests for grant verification, PoP verification, session token hashing, and signup-token validation branches.
- `e2e/tests/auth.rs` extended (+440 lines) with end-to-end coverage of the new JWT flow, including `covers_scope()` regression fixes.

## Breaking changes

- Cookie auth remains functional but the SDK API surface has shifted: `PubkyAuthFlow` → `PubkyJwtAuthFlow` (or `PubkyCookieAuthFlow` for the legacy flow); `start_auth_flow` now takes an `AuthFlowKind` and a `ClientId` for JWT variants.

## TODO

- [ ] JS SDK still uses the cookie auth. I will add the JWT auth as soon as we agree on how the rust SDK should look like.
- [ ] Potentially better DX for persisting sessions to a file


pubky-homeserver/src/client_server/auth/state.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Auth-specific state. Auth route handlers extract this instead of the
+/// global `AppState`, keeping the auth module fully self-contained.
+#[derive(Clone, Debug)]
+pub struct AuthState {
+    pub(crate) auth_service: AuthService,
+    /// Database handle for deprecated cookie auth. JWT flows use `auth_service`.
+    pub(crate) sql_db: SqlDb,
+    pub(crate) verifier: AuthVerifier,
+    pub(crate) signup_mode: SignupMode,
+}
```
86667, line 17: I like the `AuthState` idea to break apart `AppState` into modules.

Why not have a `CookieAuthService` which wraps the `sql_db` and `verifier`. It isnt obvious that `AuthVerifier` is only for cookies.

`signup_mode` can live inside `AuthService`?

pubky-homeserver/src/client_server/routes/tenants/mod.rs
```diff
@@ -28,5 +30,5 @@ pub fn router(state: AppState) -> Router<AppState> {
         .layer(DefaultBodyLimit::max(100 * 1024 * 1024))
         // XXX: dzdidi - WebDAV compliant auth. Which is actually http basic auth so we need some magic here
```
86667: This comment doesnt apply here so much anymore. Shall we move it to the cookie auth layer or remove it completely in favour of an issue?

pubky-homeserver/src/client_server/auth/jwt/crypto/grant_verifier.rs
```diff
@@ -0,0 +1,185 @@
+//! Grant verification with full Ed25519 signature check.
+//!
+//! Verifies a Grant JWS compact string, extracting and validating all claims.
+//! Homeserver-only — the SDK only decodes grants without verification.
+//!
+//! Pubky Ring creates these grants to give the SDKs the necessary information to authenticate and authorize requests to the homeserver.
+//! The homesserver verifies the grant and returns a short-lived access token for API calls.
```
86667:
```suggestion
//! The homeserver verifies the grant and returns a short-lived access token for API calls.
```

pubky-homeserver/src/client_server/auth/jwt/crypto/pop_verifier.rs
```diff
@@ -0,0 +1,329 @@
...
+use super::jws_crypto::{self, JwsCompact};
+
+/// ±3 minutes — matches existing `AuthToken` `TIMESTAMP_WINDOW` in `pubky-common/src/auth.rs`.
+pub const POP_MAX_AGE_SECS: u64 = 180;
+
+/// Nonce GC threshold: 2x the PoP window to cover edge cases.
+pub const POP_NONCE_GC_THRESHOLD_SECS: u64 = 2 * POP_MAX_AGE_SECS;
+
+/// PoP proof JWS claims — the serializable payload.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
86667: Should this, or some simplified version of it, live in `pubky-common`? The schema here is not shared. SDK side:
```
    let claims = serde_json::json!({
        "aud": homeserver_pk.z32(),
        "gid": grant_id,
        "nonce": PopNonce::generate(),
        "iat": now_unix(),
    });
```

pubky-homeserver/src/client_server/auth/jwt/middleware.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    fn call(&mut self, mut req: Request<Body>) -> Self::Future {
+        let state = self.state.clone();
+        let mut inner = self.inner.clone();
+
+        Box::pin(async move {
+            let bearer = match extract_bearer_token(&req) {
+                Ok(Some(bearer)) => bearer,
+                Ok(None) => return inner.call(req).await.map_err(|e| match e {}),
```
86667:
```
.map_err(|e| match e {})
``` 
what is the purpose of this?

pubky-homeserver/src/client_server/routes/tenants/mod.rs
```diff
@@ -28,5 +30,5 @@ pub fn router(state: AppState) -> Router<AppState> {
...
         // to make session based auth look like http auth while also accepting http basic auth for webDAV comp
-        .layer(AuthorizationLayer::new(state.clone()))
+        .layer(AuthenticationLayer::new(state.auth_state))
```
86667: Having this at this level means that the session lookup is done for all routes, even if they dont need auth.

86667: With rate limits however we do want to know whether any request is authenticated, even if authentication is not required for that route. This pattern is the simplest solution. 

We could think about a future enhancement in which auth is detected, then rate limit layer applied, then only verified in those routes that use it.  Though this feels like over-engineering

pubky-homeserver/src/client_server/auth/jwt/middleware.rs
```diff
@@ -0,0 +1,308 @@
...
+use tower::{Layer, Service};
+
+// ── Bearer token extraction ────────────────────────────────────────────────
+
+/// Extract and validate the Bearer token from the Authorization header.
+///
+/// - `Ok(Some(bearer))` — a [`SessionBearer`] that passed `parse` (non-empty, within length bound).
+/// - `Ok(None)` — no Authorization header present.
+/// - `Err(HttpError)` — Authorization header present but malformed, empty, or oversized.
+fn extract_bearer_token(req: &Request<Body>) -> Result<Option<SessionBearer>, HttpError> {
```
86667: nit: This fn returns a http error which is never used - the code silently moves forward when Err is returned. Could be misleading

pubky-homeserver/src/client_server/app.rs
```diff
@@ -227,15 +223,18 @@ fn base() -> Router<AppState> {
 }
 
 pub fn create_app(state: AppState, context: &AppContext) -> Router {
+    let auth_state = state.auth_state.clone();
     let app = base()
         .merge(tenants::router(state.clone()))
+        .with_state(state)
+        .merge(auth::base_router(auth_state.clone()))
+        .merge(auth::tenant_router(auth_state))
         .layer(CookieManagerLayer::new())
```
86667: This worries me slightly, I had the same feeling in my own work on HS recently.

Layers are dependent on other layers running before them. Shall we document clearly here which relies on which, add a unit test which breaks when the ordering is changed or perhaps make use of `debug_assert` inside of the layer?

pubky-homeserver/src/client_server/auth/jwt/service.rs
```diff
@@ -0,0 +1,997 @@
...
+        bearer: &SessionBearer,
+    ) -> Result<GrantSession, AuthServiceError> {
+        let hash = bearer.hash();
+        let session = map_not_found(
+            GrantSessionRepository::get_by_token_hash(&hash, &mut self.sql_db.pool().into()).await,
+            AuthServiceError::SessionNotFound,
+        )?;
+
+        let grant = self.get_grant(&session.grant_id).await?;
+        grant.require_active(Utc::now().timestamp())?;
```
86667: Does the session's expiry time need to be checked too? If this check is done elsewhere, can you doc

pubky-homeserver/src/client_server/auth/grant/service_error.rs
```diff
@@ -0,0 +1,84 @@
...
+    GrantOwnershipMismatch,
+
+    /// Session lacks root capability.
+    #[error("Root capability required")]
+    RootCapabilityRequired,
+
+    /// Database or infrastructure error.
+    #[error("Internal error: {0}")]
+    Internal(#[from] sqlx::Error),
+}
```
86667, line 80: Missing SessionExpired

pubky-homeserver/src/client_server/auth/grant/service.rs
```diff
@@ -0,0 +1,997 @@
...
+        if let Err(e) = PopNonceRepository::garbage_collect(
+            POP_NONCE_GC_THRESHOLD_SECS,
+            &mut self.sql_db.pool().into(),
+        )
+        .await
+        {
+            tracing::warn!("PoP nonce garbage collection failed: {e}");
+        }
+
+        PopNonceRepository::check_and_track(&pop.nonce, &mut self.sql_db.pool().into())
```
86667, line 336: nit: General db error returned at NonceReplay

pubky-homeserver/src/persistence/sql/migrations/m20260325_create_grant_sessions.rs
```diff
@@ -0,0 +1,219 @@
...
+            .from(GRANT_SESSIONS_TABLE, GrantSessionIden::GrantId)
+            .to(GRANTS_TABLE, GrantIden::Id)
+            .on_delete(ForeignKeyAction::Cascade)
+            .to_owned();
+        let query = fk.build(PostgresQueryBuilder);
+        sqlx::query(query.as_str()).execute(&mut **tx).await?;
+
+        // Index on token_hash (unique)
+        let idx = sea_query::Index::create()
+            .name("idx_grant_sessions_token_hash")
```
86667: `TokenHash` is `unique_key()` so already has an index created.

pubky-homeserver/src/persistence/sql/migrations/m20260325_create_grant_sessions.rs
```diff
@@ -0,0 +1,219 @@
...
+        let statement = Table::create()
+            .table(POP_NONCES_TABLE)
+            .if_not_exists()
+            .col(
+                ColumnDef::new(PopNonceIden::Nonce)
+                    .string_len(36)
+                    .primary_key(),
+            )
+            .col(
+                ColumnDef::new(PopNonceIden::CreatedAt)
```
86667, line 170: Index on created_at maybe?

pubky-homeserver/src/client_server/auth/jwt/service.rs
```diff
@@ -0,0 +1,997 @@
...
+const DEFAULT_SESSION_TOKEN_LIFETIME_SECS: u64 = 3600;
+
+/// Facade for all grant-based auth operations.
+///
+/// Constructed once and stored in `AppState`. Encapsulates the verify → persist
+/// → mint pipeline so route handlers stay thin.
+#[derive(Clone, Debug)]
+pub struct AuthService {
+    sql_db: SqlDb,
+    homeserver_keypair: Keypair,
```
86667: Can have the public key only

86667 (commented):
Ive looked through the Homeserver side mainly to understand the overall picture. I'll go through the crypto parts again before merge.

Overall the code is very well structured. Its easy to read and intuitive. Nicely done.

Comments are largely nits and questions.

pubky-homeserver/src/client_server/auth/router.rs
```diff
@@ -0,0 +1,60 @@
...
+/// - `POST /session` — cookie-based signin (deprecated)
+/// - `POST /auth/jwt/session` — grant-based JWT session creation
+pub fn base_router(auth_state: AuthState) -> Router<()> {
+    Router::new()
+        // Cookie (deprecated)
+        .route("/signup", post(cookie::routes::signup))
+        .route("/session", post(cookie::routes::signin))
+        // JWT
+        .route("/auth/jwt/signup", post(jwt::routes::signup))
+        .route("/auth/jwt/session", post(jwt::routes::create_grant_session))
```
86667: Im wondering if the `/jwt` naming is good here - there is actually no JWT involved. Its confusing.

How about `/auth/grant/*`. This would match the naming youve used throughout the code too.

pubky-homeserver/src/client_server/auth/state.rs
SeverinAlexB: Added CookieAuthService + UserAuthService.

pubky-homeserver/src/client_server/routes/tenants/mod.rs
SeverinAlexB: I think this is just a comment in case we want to make this endpoint fully WebDav compliant. Correct @dzdidi ?

dzdidi: You both a right. The webdav was never fully implemented, well functionality is there but it is not properly  surfaced with he biggest roadblock being auth also because knowing about upcoming auth re-write we decided to not touch it and only tested with http basic (not sure if it even works now), for it users would need to extract their actual cookie secret value and pass it as http auth password. So maybe the best would be to remove comment in favour of issue "surfact webdav functionality behind new auth in web dav compatible way" the last part may be tricky because I remember different webdav clients supporting different sets of authentication with small overlap (while standard has a long list)

pubky-homeserver/src/client_server/routes/tenants/mod.rs
SeverinAlexB: I think we could optimize a lot here but I also think it's over-engineering at this point like you said.

Two points:
- The authentication is only done when an actual auth token (header or cookie) is sent with. So for unauthenticated requests, there is no overhead.
-`AuthenticationLayer` does not reject unauthenticated requests. It just adds the option to retrieve the session if the route requests it.

pubky-homeserver/src/client_server/app.rs
SeverinAlexB: Added comments + tests.

pubky-homeserver/src/persistence/sql/migrations/m20260325_create_grant_sessions.rs
SeverinAlexB: Removed 👍

pubky-homeserver/src/client_server/auth/jwt/crypto/pop_verifier.rs
SeverinAlexB: Extracted to pubky-common

pubky-homeserver/src/client_server/auth/jwt/middleware.rs
SeverinAlexB: AI artifact. Removed

pubky-homeserver/src/client_server/auth/jwt/service.rs
SeverinAlexB: Good point. The code assumed that the session expiry is always linked to the grant expire so checking the grant is enough. Custom made grants may get around this though. This is not really a big security concern but a check worth doing. Fixed.

pubky-homeserver/src/client_server/auth/grant/service_error.rs
SeverinAlexB: Fixed with https://github.com/pubky/pubky-core/pull/373#discussion_r3195909123

pubky-homeserver/src/client_server/auth/grant/service.rs
SeverinAlexB: Fixed

pubky-homeserver/src/persistence/sql/migrations/m20260325_create_grant_sessions.rs
SeverinAlexB: Added. Probably will not matter that much as the PopNonces older than 6min will be cleaned up anyway but it also doesn't hurt.

pubky-homeserver/src/client_server/auth/router.rs
SeverinAlexB: Renamed.

pubky-homeserver/src/client_server/routes/tenants/mod.rs
SeverinAlexB: I removed the comment. The webdav endpoints in the client server, as far as I know, are not webdav standard compliant anyway so to make it work with webdav clients, this would need to be fixed first anyway. Even though I love the idea of standard webdav endpoints and auth, I think we should move this a low priority task into the future.  https://github.com/pubky/pubky-core/issues/237#issuecomment-4395176557

SeverinAlexB:
Incorporated all the feedback and merged in all the changes from the `main` branch.
The rate limiting features had to be adapted to the new auth system.
Ready for another review round.

pubky-common/src/auth/grant_session.rs
```diff
@@ -0,0 +1,103 @@
+//! Grant-based session response types.
+//!
+//! These types represent the response from `POST /session` and grant management
```
86667: Should be `/auth/grant/session`?

pubky-common/src/auth/grant.rs
```diff
@@ -0,0 +1,114 @@
...
+///   "client_id": "franky.pubky.app",
+///   "caps": ["/pub/franky/:rw"],
+///   "cnf": "{client_pubky_z32}",
+///   "jti": "{grant_id}",
+///   "iat": 1700000000,
+///   "exp": 1731536000
+/// }
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GrantClaims {
```
86667, line 34: nit: file names are not great here. `grant.rs` has `GrantClaims` and `grant_session.rs` has `GrantInfo` and `GrantSessionInfo`. Maybe one `grant_types.rs`

pubky-common/src/auth/jws.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /// Parse and validate an existing ID string.
+    ///
+    /// Must be non-empty and at most 36 characters.
+    pub fn parse(s: &str) -> Result<Self, Error> {
+        if s.is_empty() {
+            return Err(Error::InvalidFormat("RandomId must not be empty"));
+        }
+        if s.len() > RANDOM_ID_MAX_LENGTH {
```
86667, line 105: curious why we set a max at all, and why 36 chars if we generate 22 chars?

pubky-common/src/auth/jws.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// ── PublicKey serde as z32 string ────────────────────────────────────────────
+
+/// Serde helper for serializing/deserializing [`PublicKey`] as a z32 string in JSON.
+///
+/// Use with `#[serde(with = "pubky_common::auth::jws::pubkey_z32")]` on struct fields.
+/// This is needed because `PublicKey`'s default serde (via ed25519-dalek) serializes
+/// as raw bytes, but JWS payloads need human-readable z32 strings.
+pub mod pubkey_z32 {
```
86667: this looks like its more generally useful than just jws. Pull out into own file?

pubky-sdk/bindings/js/src/actors/auth_flow.rs
```diff
@@ -1,4 +1,9 @@
-use pubky::PubkyAuthFlow;
+#![allow(
+    deprecated,
+    reason = "JS binding wraps the deprecated cookie auth flow; grant wrapper will ship in a follow-up"
```
86667: In a follow-up?

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
```diff
@@ -57,21 +66,53 @@ impl FromStr for DeepLink {
...
+            "cid" => has_cid = true,
+            "cpk" => has_cpk = true,
+            _ => {}
+        }
+    }
+    match (has_cid, has_cpk) {
+        (true, true) => Ok(true),
+        (false, false) => Ok(false),
+        (true, false) => Err(DeepLinkParseError::MissingQueryParameter("cpk")),
+        (false, true) => Err(DeepLinkParseError::MissingQueryParameter("cid")),
```
86667, line 112: This looks like logic which should live with the `SigninGrantDeepLink` type itself

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
```diff
@@ -57,21 +66,53 @@ impl FromStr for DeepLink {
         }
         let intent = url.host_str().unwrap_or("").to_string();
         match intent.as_str() {
-            "signin" => Ok(DeepLink::Signin(s.parse()?)),
-            "signup" => Ok(DeepLink::Signup(s.parse()?)),
+            "signin" => {
+                if has_grant_params(&url)? {
```
86667, line 70: Match on a parse into each of the types of deep links rather than manually checking for specific fields?

pubky-sdk/src/actors/auth/grant/view.rs
```diff
@@ -0,0 +1,149 @@
+//! grant-only capability view — type-safe access to grant-specific operations.
+//!
+//! [`GrantSessionView`] is obtained via
+//! [`PubkySession::as_grant`](crate::actors::session::core::PubkySession::as_grant).
+//! The view borrows the session, so it cannot outlive it; this is what makes
+//! the grant-only API impossible to misuse against a cookie session.
```
86667, line 6: Interesting idea

pubky-sdk/src/actors/session/credential.rs
```diff
@@ -0,0 +1,100 @@
...
+/// loop is single-threaded and the underlying types contain `Rc`/`RefCell`.
+///
+// `async-trait` defaults to boxing futures as `Pin<Box<dyn Future + Send>>`,
+// which doesn't compile on `wasm32-unknown-unknown` because the JS/WASM
+// futures from `wasm-bindgen-futures` hold `Rc<RefCell<…>>` and aren't
+// `Send`. The `?Send` variant drops that bound for WASM only; native keeps
+// the `Send` bound so tokio's multi-threaded runtime stays happy.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub(crate) trait SessionCredential: Debug + Send + Sync {
```
86667, line 64: Good work splitting this out and keeping the core functionality non-breaking

86667 (commented):
I like the route youve gone for the SKD a lot 👍 

The main flow is the same for legacy cookie, new cookie and new grant flows, with all complexity hidden behind flow and session structs. good stuff

I dont see recovery file read/write option for the new `GrantCredential`. Is that intentional? 

Here are some more nits and small change requests. I havent yet gone through the sdk side fully. When youre happy with the route taken let me knwo and I can do that.

pubky-common/src/auth/grant.rs
SeverinAlexB: They are kind of different. `grant.rs` contains the actual JWS grant type. `grant_session.rs` contains the api responses. I renamed `grant_session.rs` to `grant_session_responses.rs` to make the distinction clearer. I hope this is enough.

pubky-common/src/auth/jws.rs
SeverinAlexB: Changed to 22. This was overengineering on my side to also allow uuid random ids but they are not really needed.

pubky-common/src/auth/jws.rs
SeverinAlexB: Moved to the PublicKey definition in `keys.rs`. This also remove the `#[serde(with = "crate::auth::jws::pubkey_z32")]` decorations

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
SeverinAlexB: The whole deep link module needs a refactor. These are existing issues that need to be dealt with. Should do in a separate PR to avoid blowing this one up.
https://github.com/pubky/pubky-core/issues/378

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
SeverinAlexB: https://github.com/pubky/pubky-core/issues/378

pubky-sdk/src/actors/auth/grant/view.rs
SeverinAlexB: Not sure if the best but it looks like it works

SeverinAlexB:
Incorporated the feedback. I refactored the auth e2e tests, they are now split into `grant.rs` and `cookie.rs`. + added more coverage to really know that it works.

I did introduce a few breaking changes:
- `signer.signup() -> Result<PubkySession>` is not returning `Result<()>`. The signup should strictly sign up, not sign in. This makes sure that we don't create unnecessary sessions that are the listed as session in Pubky Ring. `signer.signup_cookie()` stays available if needed.
- I changed the default behaviour of `signer.signin()` to use the grant, not cookies. Cookie signin is available as `signer.signin_cookie()`.

I am a bit unsure if I should do these breaking changes. Another path would be to keep the `signup()`/`signin()` methods unchanged but add `signup_grant()`/`signin_grant()`. I am afraid people will not bother to upgrade then.

Anyway, I will need to write a migration guide before we release the new SDK.

What I am not doing in this PR because otherwise it will grown in size even more:
- Optimize the JS SDK
- Optimize saving and restore a grant session
- Optimizing examples
- Migration guide

Let me know what you think!

examples/rust/0-logging/main.rs
```diff
@@ -50,7 +53,8 @@ async fn main() -> Result<()> {
 
     let signer = pubky.signer(keypair);
     info!(homeserver = %homeserver.public_key(), "Signing up");
-    let session = signer.signup(&homeserver.public_key(), None).await?;
+    signer.signup(&homeserver.public_key(), None).await?;
+    let session = signer.signin(ClientId::new("logging.example")?).await?;
```
86667, line 57: I dont like how ClientId is required here, its cluttering what should be a simple interface. 

Can we generate these internally by default, with a `signup_with_client_id()` or a builder method for specifying ClientId?

pubky-sdk/src/actors/auth/grant/flow.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// [`PubkySession::from_grant_credential`]. Use
+    /// [`await_credential`](Self::await_credential) directly if you need to
+    /// inspect or persist the credential before building a session.
+    ///
+    /// # Errors
+    /// - Returns [`crate::errors::Error::Authentication`] if the relay channel
+    ///   expires before approval.
+    /// - Propagates HTTP/transport failures while polling the relay or
+    ///   exchanging the grant for a bearer.
+    /// - Returns [`crate::errors::Error::Authentication`] if the issuer's
```
86667, line 238: I noticed that this says relay channel expiry and pkarr resolution failure both return `crate::errors::Error::Authentication`. looking into it i see that they have distinct `AuthError` variants, we should document these here too

pubky-sdk/src/actors/signer/session.rs
```diff
@@ -133,6 +228,48 @@ impl PubkySigner {
         AuthToken::sign(&self.keypair, capabilities)
     }
 
+    fn signup_grant(&self, client_keypair: &Keypair) -> Result<(String, GrantClaims)> {
+        let client_id = ClientId::new(SIGNUP_CLIENT_ID)
+            .map_err(|e| crate::errors::AuthError::Validation(e.to_string()))?;
+        let claims = self.grant_claims(client_id, client_keypair, SIGNUP_GRANT_LIFETIME_SECS);
+        let jws = pubky_common::auth::jws::sign_jws(&self.keypair, GRANT_JWS_TYP, &claims);
```
86667: Could this live inside the `GrantClaims` type with a `GrantClaims.sign(keypair)`? Maybe renaming to `Grant`. Could replace the existing `GrantApproval`. Or at least return `GrantApproval` from this fn.

I dont know, I havent confirmed but Im thinking that these different Grant  types could be combined which would make the abstractions easier to understand and fn signatures cleaner.

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// Refresh the bearer proactively when it has less than this many seconds left.
+pub(crate) const REFRESH_SLACK_SECS: u64 = 300;
+
+/// Current Unix timestamp in seconds, cross-target.
+pub(crate) fn now_unix() -> u64 {
+    web_time::SystemTime::now()
+        .duration_since(web_time::UNIX_EPOCH)
+        .map(|d| d.as_secs())
+        .unwrap_or(0)
```
86667: I feel that this default will only ever cause confusing errors. Better to return an error?

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/// Mutable grant credential state. Always wrapped in `Arc<Mutex<...>>`.
+///
+/// Refresh paths take the mutex and hold it across the HTTP call so
+/// concurrent refreshes serialize.
```
86667:
```suggestion
/// concurrent refreshes dont race.
```

This confused me. I thought it meant serialise as in serde

pubky-sdk/src/actors/signer/session.rs
```diff
@@ -55,8 +74,9 @@ impl PubkySigner {
     /// # Errors
     /// - Propagates transport failures during the session exchange.
     /// - Propagates validation errors from the session exchange or PKDNS publishing.
-    pub async fn signin(&self) -> Result<PubkySession> {
-        self.signin_with_publish(PublishMode::Background).await
+    pub async fn signin(&self, client_id: ClientId) -> Result<PubkySession> {
```
86667, line 77: I wonder if it should be possible here to create a grant with capabilities (ie not root) and sign it immediately. Having to use a relay in this case seems silly.

Also, it may be worth being more explicit that this `signin()` fn is all-powerful root access. `signin_root()`? what do you think

pubky-sdk/src/actors/auth/cookie/legacy_api.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// # Errors
+    /// - Returns [`std::io::Error`] if the file cannot be written or
+    ///   permissions cannot be set.
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn write_secret_file<P: AsRef<std::path::Path>>(
+        &self,
+        secret_file_path: P,
+    ) -> std::io::Result<()> {
+        self.as_cookie()
+            .expect("write_secret_file() is only valid for cookie sessions")
```
86667, line 84: Is there any warning of this at compile time? Panic feels harsh. 

How about marking these cookie-only fns as deprecated as an early warning?

pubky-sdk/src/actors/signer/auth.rs
```diff
@@ -3,104 +3,159 @@ use reqwest::Method;
...
+    actors::auth::deep_links::{DeepLink, DeepLinkParseError},
+    cross_log,
     errors::{AuthError, Result},
     util::check_http_status,
 };
 
 use super::PubkySigner;
 
+/// Default lifetime for issued grants: 2 years (per proposal v4-pop §"Token Lifetime").
+const DEFAULT_GRANT_LIFETIME_SECS: u64 = 2 * 365 * 24 * 3600;
```
86667: This exists in both `/signer` files

pubky-sdk/src/pubky.rs
```diff
@@ -323,6 +356,11 @@ fn parse_auth_deep_link(url: &str) -> Result<(Capabilities, url::Url, [u8; 32],
             *s.secret(),
             AuthFlowKind::signup(s.homeserver().clone(), s.signup_token()),
         )),
+        DeepLink::SigninGrant(_) | DeepLink::SignupGrant(_) => Err(AuthError::Validation(
+            "grant auth flows cannot be resumed via resume_auth_flow; this API is legacy cookie only."
```
86667: why cant grant flows be resumed? do we lose any usefulness by not supporting for grants?

examples/rust/0-logging/main.rs
86667: I see that this is how OAuth2 does things, which makes me feel slightly better about it.

We could probably do an okay-ish job of generating these ourselves from local env and falling back to a random string. Or even just "sdk-{env}" if clashes dont matter? In browser we could use the domain I expect.

This is adding more complexity, though this is the main entry point so we need to get it right first time and i'm favouring simplicity with builder methods for customising.

SeverinAlexB: I am torn with this. The Client ID is needed. It is going to be displayed in Pubky Ring so users know which session belongs to which application. I feel like if we make it optional, devs will not bother and our session list in Ring will look like a mess.

Let me know what you think

pubky-sdk/src/actors/auth/grant/flow.rs
SeverinAlexB: As far as I know this is [an existing error](https://github.com/pubky/pubky-core/blob/main/pubky-sdk/src/actors/auth/auth_flow.rs#L140). Opened an issue https://github.com/pubky/pubky-core/issues/391

pubky-sdk/src/actors/signer/session.rs
SeverinAlexB: Added a `GrantClaims::sign` method. I will specifically not rename it because Claims are something different then the grant. Claims = data body of the JWS. Grant = the JWS with the grant body.

GrantApproval is just a helper type to deal with the data better.

pubky-sdk/src/actors/auth/grant/credential.rs
SeverinAlexB: Added an expect. This error should never happen anyway.

pubky-sdk/src/actors/auth/grant/credential.rs
SeverinAlexB: Serialize = serial execution. I replaced it with the race wording though for easier understanding.

pubky-sdk/src/actors/signer/session.rs
SeverinAlexB: > I wonder if it should be possible here to create a grant with capabilities (ie not root) and sign it immediately. Having to use a relay in this case seems silly.

Feel free to open an issue on this. This is an existing shortcoming and not associated with this PR.

> Also, it may be worth being more explicit that this signin() fn is all-powerful root access. signin_root()? what do you think

I am open to this. I had the same discussion with @SHAcollision. This should not be part of this PR though.

pubky-sdk/src/actors/auth/cookie/legacy_api.rs
SeverinAlexB: Added deprecation warnings.

pubky-sdk/src/pubky.rs
SeverinAlexB: I does not support it because the Proof of Possession private key is not in the url. I clarified the error text. I need to think about how to make resuming possible

SeverinAlexB: I added a way to save/restore a GrantAuthFlow. I didn't add this to the `pubky` facade to prevent the facade to get too big. Happy to be convinced otherwise though.

I am really having the urge to refactor a lot in the SDK but this PR can't get any bigger.

examples/rust/0-logging/main.rs
86667: The trade-off is between the mental load a dev needs to take on to first hello-world and the user experience of a user signed in to multiple different apps. 

Since the only currently existing apps are either controlled by us (pubky.app, eventky, etc.) or are vibed and basically unused, I think we can afford to focus on the dev experience for now. If it does become a problem later we can always change the interface to force/encourage `ClientID` setting.  

I'd say lets do the following:
- Set a default now, eg `sdk-session-${rand_8_bytes}`
- Encourage devs to set their own value in the docs and examples
- Create a ticket to be implemented later in which we attempt to fetch info from the environment for the default. eg the UserAgent in browsers - `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:150.0)`

SeverinAlexB: Even AI is pushing back when I ask it what it thinks about making it optional.

> I’d be cautious. A random default is convenient, but it weakens the main value of client_id: a stable, user-visible application identifier in grant claims and grant management.
My take:
> - Good idea if client_id is only a debug/audit label and not meant to identify the app meaningfully.
> - Risky if users will see grants later and need to revoke/understand them, because unknown-not-set-by-dev-{random} makes every grant look like a different anonymous app.
> - It also removes compile-time pressure on developers to set a real app/domain identifier, which is probably desirable for production auth flows.
> - The proposed prefix is good as a warning signal, but because it is random per flow, it still creates noisy grant lists and poor UX.

The more I look into it, the less I think it is a good idea to drop it. The friction for new devs is not that high. They likely copy the code example anyway where there is an example Client ID configured. Removing the client id would also remove it from the default facade `pubky::start_grant_auth_flow(capabilities, auth_kind, client_id)`.  A serious developer must then use `PubkyGrantAuthFlow` which is way more difficult.

So I am biased towards keeping the client id. It's not such a big deal IMO.

SeverinAlexB:
TODO after the PR:

- [x] Refactor deep links https://github.com/pubky/pubky-core/pull/392
- [ ] Optimize saving and restore a grant session
- [ ] Optimize the JS SDK
- [ ] Optimizing examples
- [ ] Migration guide
- [ ] Potentially refactor `PubkySession` to `PubkySession<AuthBackend>`

pubky-sdk/src/actors/signer/session.rs
86667: Not your code, worth logging result or error here

pubky-sdk/src/pubky.rs
86667: Examples using `Capabilities::default()` tripped my AI up - it tried to use Storage with using a Session with no Capabilities.  

May be worth a comment `// no capabilites` or using `Capabilities::from(Capability::read("/pub/app"))` for relevant doc examples.

86667 (commented):
Good stuff. Im happy to merge this as is with the follow-ups to come. Here are a few notes after using the SDK. Deal with them in follow-ups, create tickets or ignore. Just notes:

- Timestamps in grants and cookies are of different formats - seconds and ms resp. 
- It would be good to see the migration guide for the cookie flow. If we have too many asks on devs here to upgrade then we should consider keeping the old code's interface, though im hoping that `PubkyAuthFlow` -> `PubkyCookieAuthFlow` shouldnt be visible to a dev migrating but continuing to use cookies. A quick AI check seems to think it wont change
- AI was also convinced that `await_approval()` and `approve_auth()` had to run concurrently. Im sure your basic flow docs/examples will help there.
- .list() returns a builder whilst the other methods return a future. Consistent pattern better for devs and AI.

86667 (approved):

86667 (changes requested):
Seems like there are 2 breaking changes to the API here:

- Writes outside /pub/: Status code changed from 400 to 403
- Capability scope matching tightened: File-scope capabilities (no trailing /) no longer match descendant paths

I want to confirm that these are known about. The first im not too concerned with, but the second could cause issues. Thoughts?

SeverinAlexB:
> Capability scope matching tightened: File-scope capabilities (no trailing /) no longer match descendant paths

I see this more as a bug fix, not really a breaking change. If you request permission to `/pub/pubky.app/settings.json` you should not have access to `/pub/pubky.app/settings.json/...`.

I didn't think this is going to be a problem downstream but good bringing it to my attention. Maybe we need to notify downstream devs about this. I hope they requested folder permissions, not file.

Edit: On our prod homeserver, no prod related app ever requested a folder permission without a trailing slash.

> Writes outside /pub/: Status code changed from 400 to 403

Same as you, I am not concerned.

Thx for the thorough testing.

86667 (approved):

Issue #364, opened 2026-04-01
[homeserver] Admin DAV endpoint returns 500 instead of 507 when storage quota is exceeded


86667:
When writing files through the admin WebDAV endpoint (/dav/...), exceeding the user's storage quota causes a 500 Internal Server Error with an opendal warning instead of a proper 507 Insufficient Storage response.

## Steps to reproduce
- Create a user with a storage quota
- PUT a file via the admin DAV endpoint that fills the quota
- PUT another file to the same endpoint

#### Expected behavior
The second PUT returns 507 Insufficient Storage.

#### Actual behavior
The second PUT returns 500 Internal Server Error with this log:

```
WARN opendal::layers::complete: writer has not been closed or aborted, must be a bug
DEBUG dav_server::davhandler: == END REQUEST result FsError(GeneralFailure)
ERROR tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error
```

andrei-21:
The warning you see is because of the bug in `dav_server_opendalfs` https://github.com/apache/opendal/issues/7425.

andrei-21:
The 500 comes from the admin WebDAV path using `dav_server_opendalfs`.

That adapter converts OpenDAL errors into `dav_server::fs::FsError`, and in this case an OpenDAL `ErrorKind::RateLimited` becomes `FsError::GeneralFailure`, which the DAV layer surfaces as 500. From the adapter’s point of view, that mapping is not obviously wrong: in Pubky, the distinction that this particular rate-limit is actually a storage quota condition is carried in the error message, not in a dedicated error kind that `dav_server_opendalfs` can reliably translate to 507 `Insufficient Storage`.

This does not affect the user-facing WebDAV interface. That path does not rely on `dav_server_opendalfs` in the same way; it handles WebDAV behavior more directly and already treats the quota-exceeded case explicitly, so users get the expected 507.

I am closing this because there is no straightforward fix in the current integration boundary, and the issue does not appear to cause a real functional problem.

PR #374, opened 2026-04-22
test: add unit test to verify actual behavior on quota excess

andrei-21:
Currently it returns 500. See #364.


86667 (approved):
Approving but why not fix the issue at the same time?

andrei-21:
> Approving but why not fix the issue at the same time?

The fix is not trivial (and not yet ready).

PR #375, opened 2026-04-24
refactor(common)!: normalize capabilities construction

### LINKED ISSUE #232: fix(common): sanitize dedup redundant capabilities
An app could request capabilities like
```
/pub/example.com/:rw
/pub/example.com/contacts:w
```
In this case, `/pub/example.com/contacts:w` is fully covered by `/pub/example.com/:rw` . Ideally our `Capabilities` builder and validator should clean up and sanitise this case.

After the fix, these tests should be satisfied.
```
    #[test]
    fn redundant_capabilities_builder_dedup() {
        let caps = Capabilities::builder()
            .rw("/pub/example.com/")
            .rw("/pub/example.com/") // Same capability twice
            .write("/pub/example.com/subfolder") // Unnecessary capability, already covered by `/pub/example.com/`
            .finish();
        assert_eq!(caps.to_string(), "/pub/example.com:rw");
    }

    #[test]
    fn redundant_caps_string_dedup() {
        // Repeated capability and unnecessary capability subfolder on the str to parse
        let parsed = Capabilities::try_from("/pub/example.com/:rw,/pub/example.com/:rw,/pub/example.com/subfolder:w").unwrap();
        let caps = Capabilities::builder()
            .rw("/pub/example.com/")
            .finish();
        assert_eq!(caps.to_string(), "/pub/example.com:rw");
        assert_eq!(parsed.to_string(), caps.to_string());
    }
```

andrei-21:
Keep capability lists canonical by deduplicating repeated scopes and removing entries covered by broader capabilities.

BREAKING CHANGE: `Capabilities(vec![...])` construction, `to_vec()`, and `from_url()` were removed. Use `Capabilities::from(vec![...])` or `Capabilities::builder()` to construct values, `Into<Vec<Capability>>` for owned access, and `(&url).into()` for URL parsing instead.

Closes #232.

### DIFF
--- a/pubky-common/src/capabilities.rs
+++ b/pubky-common/src/capabilities.rs
@@ -272,7 +272,7 @@ impl TryFrom<&str> for Capability {
 
         let scope = value[0..value.len() - actions_str.len() - 1].to_string();
 
-        Ok(Capability { scope, actions })
+        Ok(Self { scope, actions })
     }
 }
 
@@ -327,9 +327,30 @@ pub enum Error {
 /// bespoke encoder/decoder instead of serde.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
 #[must_use]
-pub struct Capabilities(pub Vec<Capability>);
+pub struct Capabilities(Vec<Capability>);
 
 impl Capabilities {
+    /// Return a normalized capability list.
+    ///
+    /// Normalization merges duplicate scopes, de-duplicates and sorts actions,
+    /// and removes capabilities already covered by broader capabilities.
+    ///
+    /// # Examples
+    /// ```
+    /// use pubky_common::capabilities::{Capability, Capabilities};
+    ///
+    /// let caps = Capabilities::from(vec![
+    ///     Capability::read("/pub/"),
+    ///     Capability::write("/pub/"),
+    ///     Capability::read("/pub/file.txt"),
+    /// ]);
+    ///
+    /// assert_eq!(caps.normalize().to_string(), "/pub/:rw");
+    /// ```
+    pub fn normalize(self) -> Self {
+        Self(normalize(self.0))
+    }
+
     /// Returns true if the list contains `capability`.
     pub fn contains(&self, capability: &Capability) -> bool {
         self.0.contains(capability)
@@ -350,37 +371,6 @@ impl Capabilities {
         self.0.iter()
     }
 
-    /// Parse capabilities from the `caps` query parameter.
-    ///
-    /// Expects a comma-separated list of capability strings, e.g.:
-    /// `?caps=/pub/my-cool-app/:rw,/foo:r`
-    ///
-    /// Invalid entries are ignored.
-    ///
-    /// # Examples
-    /// ```
-    /// # use url::Url;
-    /// # use pubky_common::capabilities::Capabilities;
-    /// let url = Url::parse("https://example/app?caps=/pub/my-cool-app/:rw,/foo:r").unwrap();
-    /// let caps = Capabilities::from_url(&url);
-    /// assert!(!caps.is_empty());
-    /// ```
-    pub fn from_url(url: &Url) -> Self {
-        // Get the first `caps` entry if present.
-        let value = url
-            .query_pairs()
-            .find_map(|(k, v)| (k == "caps").then(|| v.to_string()))
-            .unwrap_or_default();
-
-        // Parse comma-separated capabilities, skipping invalid pieces.
-        let caps = value
-            .split(',')
-            .filter_map(|s| Capability::try_from(s).ok())
-            .collect();
-
-        Capabilities(sanitize_caps(caps))
-    }
-
     /// Start a fluent builder for multiple capabilities.
     ///
     /// ```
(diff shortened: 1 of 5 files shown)


pubky-common/src/capabilities.rs
```diff
@@ -523,21 +475,47 @@ impl TryFrom<&str> for Capabilities {
...
+            .find_map(|(k, v)| (k == "caps").then(|| v.to_string()))
+            .unwrap_or_default();
+
+        // Parse comma-separated capabilities, skipping invalid pieces.
+        let caps: Vec<_> = value
+            .split(',')
+            .filter_map(|s| Capability::try_from(s).ok())
+            .collect();
+
+        Self::from(caps)
```
SeverinAlexB: Hmmm, not sure this is a smart move. Using `Capabilities::from(&url)` doesn't show any comment anymore. The `from_url()` method is very specific in converting the `caps` query parameter to capabilities. Using the `From` traits feels like it will make the functionality even more opaque.

What do you think of keeping it as a method but renaming the method to `from_caps_url()` or so?

pubky-common/src/capabilities.rs
```diff
@@ -327,7 +327,7 @@ pub enum Error {
 /// bespoke encoder/decoder instead of serde.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
 #[must_use]
-pub struct Capabilities(pub Vec<Capability>);
+pub struct Capabilities(Vec<Capability>);
```
SeverinAlexB, line 330: Nice

pubky-common/src/capabilities.rs
```diff
@@ -847,4 +836,18 @@ mod tests {
         let back: Capabilities = serde_json::from_str(&json).unwrap();
         assert_eq!(back, caps);
     }
+
+    #[test]
+    fn redundant_caps_string_dedup() {
```
SeverinAlexB: Duplicated tests?

- `redundant_caps_string_dedup`
- `redundant_capabilities_string_dedup`

Was this functionality already implement?

pubky-common/src/capabilities.rs
```diff
@@ -495,13 +447,13 @@ impl CapsBuilder {
...
-        Capabilities(sanitize_caps(self.caps))
+        Capabilities::from(self.caps)
     }
 }
 
 impl From<Vec<Capability>> for Capabilities {
     fn from(value: Vec<Capability>) -> Self {
-        Self(value)
+        Self(sanitize_caps(value))
     }
```
SeverinAlexB, line 509: Not sure this is the best DX strategy for this struct at the moment. `from` is sanitizing now so `Capabilities::from(vec).into::<Vec<_>>() ≠ original vec`. No idea how it worked before but I think this is now a regression.

Maybe it's worth making `sanitize()` an explicit method?

SeverinAlexB (changes requested):

pubky-common/src/capabilities.rs
andrei-21: I moved the documentation to the `from` method (before it was for the trait), now it should work.

pubky-common/src/capabilities.rs
andrei-21: Oh, nice catch. Yes the functionality was there.

pubky-common/src/capabilities.rs
andrei-21: Hm, where does this expectation come from? If I convert data from one type to another, it might change; it's quite natural.
If `Capabilities` keeps the invariant (that capabilities are sanitized), then `sanitize()` should be private. And that is the whole point of having this type.

SeverinAlexB: The comment above https://github.com/pubky/pubky-core/pull/375#discussion_r3145541218 and this comment feeds into the same problem:

It's a classic dishonest function according to [Uncle Bobs Clean Code book](https://claude.ai/share/6fb4ad24-3200-441b-bc77-55f684644a06). `From` should convert. Sanitization is a side effect that is not described in the `From` function name.

Best approach IMO would be `Capabilities::normalize(...)` or so. That's also why I am proposing the `from_caps_url()` above. It makes the function contract more explicit and honest. The more explicit the code is the better.

andrei-21: You are right, `from()` must be [value-preserving](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from:~:text=The%20conversion%20is%20value%2Dpreserving).
I will change the behavior for other `From` implementations.

andrei-21: Done. Also I added explicit `normalize()` method.

pubky-common/src/capabilities.rs
SeverinAlexB: My point here is the same as in https://github.com/pubky/pubky-core/pull/375#discussion_r3146363434. Even with the comment moved, the function name covers what the method does only poorly. That's why I proposed `from_caps_url()`.

andrei-21: Are you suggesting to remove `from(&Url)` then?

SeverinAlexB: Yes. Original, we had the `from_url()` method. Just looking at this deceiving method name, it does make sense to turn it into a `From` trait. But apart from the side effect, the method does not turn every url into Capabilities so the naming is off.

```
    /// Parse capabilities from the `caps` query parameter.
    ///
    /// Expects a comma-separated list of capability strings, e.g.:
    /// `?caps=/pub/my-cool-app/:rw,/foo:r`
    ///
    /// Invalid entries are ignored.
    ///
    /// # Examples
    /// ```
    /// # use url::Url;
    /// # use pubky_common::capabilities::Capabilities;
    /// let url = Url::parse("https://example/app?caps=/pub/my-cool-app/:rw,/foo:r").unwrap();
    /// let caps = Capabilities::from_url(&url);
    /// assert!(!caps.is_empty());
    /// ```
    pub fn from_url(url: &Url) -> Self {
        // Get the first `caps` entry if present.
        let value = url
            .query_pairs()
            .find_map(|(k, v)| (k == "caps").then(|| v.to_string()))
            .unwrap_or_default();

        // Parse comma-separated capabilities, skipping invalid pieces.
        let caps = value
            .split(',')
            .filter_map(|s| Capability::try_from(s).ok())
            .collect();

        Capabilities(sanitize_caps(caps))
    }
```

My proposal would be to have a `Capabilities::from_caps_url(url)` method. But happy to listen to you if you have a different idea.

andrei-21: I moved `from()` method logic to `from_caps_url()`.

SeverinAlexB:
Great. The PR looks good. One thing before I approve: `normalize()` is now an explicit action where it was an implicit before. Do we need to call normalize() anywhere in the stack now that we didn't before?

andrei-21:
> Great. The PR looks good. One thing before I approve: `normalize()` is now an explicit action where it was an implicit before. Do we need to call normalize() anywhere in the stack now that we didn't before?

I put it in two places, do you see anything else?

SeverinAlexB:
> I put it in two places, do you see anything else?

You probably want to look at any method that called sanitize() before but doesn't anymore to catch regressions. The `finish()` method looks like it did for example.

andrei-21:
> > I put it in two places, do you see anything else?
> 
> You probably want to look at any method that called sanitize() before but doesn't anymore to catch regressions. The `finish()` method looks like it did for example.

Right. I added it to `finish()` method.

SeverinAlexB (approved):

PR #380, opened 2026-05-11
pkarr v6

dzdidi:
Ar's relay is still not the most reliable and it was kicked out of pkarr in 5.0.4 (current is 5.0.5) but pubky-core (SDK) is still using 5.0.3 (because of Cargo.lock)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1165,9 +1165,9 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "5.0.0-pre.5"
+version = "5.0.0-pre.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a434aec7908df6ca86cda069864d7686aea8afad979aadc9e30e50ac3e40b45a"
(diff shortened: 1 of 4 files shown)


dzdidi:
Updating patch version of one

SeverinAlexB:
Hmmm, pkarr 5.0.5 somehow brings in a breaking change. This is no bueno. Need to check this

Edit: https://github.com/pubky/pkarr/issues/241

ok300:
Nice, I see `pkarr 6.0.0` was recently released.

Then changing

https://github.com/pubky/pubky-core/blob/1d07863943ef10f2a7ecafe6949d1b513b83f06a/Cargo.toml#L41

to `version = "6"` and then

```
cargo clippy
```

to update `Cargo.lock` should do the trick.

ok300:
In the interest of time, I took the liberty to push these changes in 366cceee9fee8aa8e8da481ba18e34d1ea2333de :
- bumped `pkarr` to `version = "6"`
- bumped `pkarr-relay` to `version = "0.12"` (since the old version was still using `pkarr 5.0.5`)
- removed the `simple-dns` pin to `=0.9.3` as it's not needed anymore

SeverinAlexB (approved):
Let's go

PR #381, opened 2026-05-12
feat(homeserver): UserQuota option allowed_write_paths

86667:
Adds to the `UserQuota` an option to specify which write paths the user can write to.  

Returns `403 Forbidden - Write to this path is not allowed` when path is not allowed.

- `allowed_write_paths: null` means all writes allowed.
- `allowed_write_paths: []` means no writes allowed.


pubky-homeserver/src/client_server/routes/tenants/write.rs
```diff
@@ -117,10 +117,7 @@ async fn fail_if_size_hint_exceeds_quota<'a>(
 
     let max_bytes = resolve_storage_max_bytes(user, default_storage_mb);
     if would_exceed_limit(user.used_bytes, bytes_delta, max_bytes) {
-        return Err(HttpError::new_with_message(
-            StatusCode::INSUFFICIENT_STORAGE,
-            "Disk space quota exceeded. Write operation failed.",
-        ));
+        return Err(HttpError::insufficient_storage());
```
86667, line 120: Make disk quota exceed error messages consistent (between here and in opendal layer)

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
```diff
@@ -22,6 +23,26 @@ use opendal::Buffer;
...
 
 use super::super::{FileIoError, FileMetadata, FileMetadataBuilder, FileStream, WriteStreamError};
+use crate::persistence::files::layer_domain_error::LayerDomainError;
+
+/// Map OpenDAL errors to domain-specific [`FileIoError`] variants.
+///
+/// Our custom layers embed a [`LayerDomainError`] as the error source via
+/// [`opendal::Error::set_source`]. We downcast from the source to recover the
+/// typed variant, avoiding fragile `ErrorKind` / message-string matching.
+fn map_opendal_error(e: opendal::Error) -> FileIoError {
```
andrei-21: You can implement `From<opendal::Error> for FileIoError` for that.

pubky-homeserver/src/shared/user_quota.rs
```diff
@@ -271,20 +297,37 @@ pub struct UserQuota {
...
     pub fn from_nullable_columns(
         storage_quota_mb: Option<i32>,
         rate_read: Option<String>,
         rate_write: Option<String>,
         rate_read_burst: Option<i32>,
         rate_write_burst: Option<i32>,
+        allowed_write_paths: Option<String>,
     ) -> Self {
+        let allowed_write_paths = allowed_write_paths.map(|s| {
+            serde_json::from_str::<Vec<WebDavPath>>(&s).unwrap_or_else(|e| {
```
andrei-21, line 324: Why do we swallow invalid string here? Wouldn't it be better to return an error?

pubky-homeserver/src/shared/user_quota.rs
```diff
@@ -322,6 +366,29 @@ impl UserQuota {
         burst_to_i32("rate_write_burst", self.rate_write_burst)
     }
 
+    /// Allowed write paths as DB-column type (`TEXT`): JSON array string or NULL.
+    pub fn allowed_write_paths_db(&self) -> Option<String> {
+        self.allowed_write_paths
+            .as_ref()
+            .map(|paths| serde_json::to_string(paths).expect("Vec<WebDavPath> always serializes"))
```
andrei-21: In that instance, maybe you are right that `Vec<WebDavPath>` always serializes, but this code looks brittle to me. In the future, we can change how `WebDavPath` serializes, and all of a sudden the statement will not hold anymore.
I think it is almost always better to return an error in such cases.

pubky-homeserver/src/persistence/files/write_path_layer.rs
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    async fn list(&self, path: &str, args: OpList) -> Result<(RpList, Self::Lister)> {
+        self.inner.list(path, args).await
+    }
+
+    async fn presign(&self, path: &str, args: OpPresign) -> Result<RpPresign> {
+        if matches!(
+            args.operation(),
+            PresignOperation::Write(_) | PresignOperation::Delete(_)
```
andrei-21: In future versions of opendal there might be a new operation we will consider "write"; maybe it is better to reverse the check and explicitly list operations that do not require `check_write_path_allowed`?

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
```diff
@@ -53,6 +79,7 @@ pub fn build_storage_operator(
                 .layer(user_quota_layer)
                 .layer(entry_layer)
                 .layer(events_layer)
+                .layer(write_path_layer)
```
andrei-21: Will this policy apply for the admin interface? Is it what we want?

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
```diff
@@ -203,23 +199,7 @@ impl OpendalService {
     /// Get the content of a file as a stream of bytes.
     /// The stream is chunked by the CHUNK_SIZE.
     pub async fn get_stream(&self, path: &EntryPath) -> Result<FileStream, FileIoError> {
-        match self.get_stream_inner(path).await {
-            Ok(stream) => Ok(stream),
-            Err(e) => match e.kind() {
-                opendal::ErrorKind::NotFound => Err(FileIoError::NotFound),
-                opendal::ErrorKind::PermissionDenied => {
```
86667, line 210: By implementing `From<opendal::Error> for FileIoError` we do lose this 

```
opendal::ErrorKind::PermissionDenied -> FileIoError::NotFound
```
behaviour. 

We now return the actual error `FileIoError::OpenDAL(e)`. Technically a break but worth it for correctness IMO. If there is a permissions issue from the storage backend then it should be surfaced and not masked as a `NotFound`.

pubky-homeserver/src/shared/user_quota.rs
86667: This is the DB read so the data has already been validated. In the unlikely chance of bad data getting to the db somehow we fail-closed by returning the user object with no write permissions, rather than blocking all actions on the User by erroring on read.

pubky-homeserver/src/shared/user_quota.rs
86667: fixed

pubky-homeserver/src/shared/user_quota.rs
andrei-21: I see, I am not fully convinced, but I will accept this instance. Later we can discuss how we want to tackle such cases in the future.

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
86667: True, good point. Ive splut `OpendalService` into 2 `Operator`s, one with the `WritePathLayer` and another without. The one without is used by admin server.

pubky-homeserver/src/persistence/files/write_path_layer.rs
86667: fixed

pubky-homeserver/src/shared/webdav/entry_path.rs
```diff
@@ -51,6 +51,15 @@ impl EntryPath {
...
     }
+
+    /// Parse a path string into an `EntryPath`, returning an [`opendal::Error`]
+    /// on failure. Convenience wrapper for use inside OpenDAL layers where the
+    /// trait methods receive `&str` and must return `opendal::Result`.
+    pub fn parse_opendal(path: &str) -> Result<Self, opendal::Error> {
+        path.parse().map_err(|e: EntryPathError| {
+            opendal::Error::new(opendal::ErrorKind::Unexpected, e.to_string())
+        })
+    }
```
86667, line 62: I used this opportunity to remove `ensure_valid_path()` fn in favour of this parse() on the `EntryPath` type. In that code, we now return `Unexpected` instead of `PermissionDenied` because:
- The path value will have been validated already before getting to this point. Therefore this is not a breaking change
- For a parse failure, neither are ideal but `Unexpected` is better than `PermissionDenied`
- `PermissionDenied` is now used only for path not in allow list

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
```diff
@@ -117,31 +140,40 @@ impl OpendalService {
...
     ) -> Result<Self, FileIoError> {
         let operator = build_storage_operator(
+            storage_config,
+            data_directory,
+            db,
+            events_service.clone(),
+            user_service.clone(),
+            true,
+        )?;
+        let admin_operator = build_storage_operator(
```
andrei-21: Now when we have two storage operators they may note share the actual storage, for example in InMemory case.
This addition to the test makes it fail:
```diff
diff --git a/pubky-homeserver/src/admin_server/routes/delete_entry.rs b/pubky-homeserver/src/admin_server/routes/delete_entry.rs
index e7dab68..95594a3 100644
--- a/pubky-homeserver/src/admin_server/routes/delete_entry.rs
+++ b/pubky-homeserver/src/admin_server/routes/delete_entry.rs
@@ -76,6 +76,10 @@ mod tests {
         EntryRepository::get_by_path(&entry_path, &mut db.pool().into())
             .await
             .expect_err("Should be deleted");
+        file_service
+            .get(&entry_path)
+            .await
+            .expect_err("Blob should be deleted from storage");
         let events = EventRepository::get_by_cursor(None, Some(10), &mut db.pool().into())
             .await
             .unwrap();
```

Not sure if it is a big deal, since InMemory backend is supposed to be used in tests.

86667: Good catch! Fix was straight forward.

andrei-21 (approved):

PR #382, opened 2026-05-12
feat(examples): Improve write abilities

86667:
- `keygen` script generates a key and writes to a recovery file ready for usage with other examples
- Expand `4-storage` to include write example

### DIFF
--- a/examples/rust/keygen.rs
+++ b/examples/rust/keygen.rs
@@ -15,8 +15,7 @@ struct Cli {
     output: PathBuf,
 }
 
-#[tokio::main]
-async fn main() -> Result<()> {
+fn main() -> Result<()> {
     let cli = Cli::parse();
 
     // 1) Generate a fresh keypair
@@ -27,6 +26,10 @@ async fn main() -> Result<()> {
     // 2) Encrypt and save the recovery file
     println!("Enter a passphrase to encrypt the recovery file:");
     let passphrase = rpassword::read_password()?;
+    if passphrase.is_empty() {
+        anyhow::bail!("Passphrase cannot be empty");
+    }
+
     println!("Confirm passphrase:");
     let confirm = rpassword::read_password()?;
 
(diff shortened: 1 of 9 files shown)


examples/rust/keygen.rs
```diff
@@ -0,0 +1,42 @@
...
+    version,
+    about = "Generate a keypair and save a passphrase-encrypted recovery file"
+)]
+struct Cli {
+    /// Path to write the recovery file
+    #[arg(short, long, default_value = "recovery.key")]
+    output: PathBuf,
+}
+
+#[tokio::main]
```
SHAcollision: Nit: looks like this binary is fully synchronous, so we can simplify by using a plain `fn main() -> Result<()>` and dropping `#[tokio::main]`. Keeps the example minimal.

examples/rust/keygen.rs
```diff
@@ -0,0 +1,42 @@
...
+    let cli = Cli::parse();
+
+    // 1) Generate a fresh keypair
+    let keypair = Keypair::random();
+    println!("Generated new keypair");
+    println!("Public key: {}", keypair.public_key());
+
+    // 2) Encrypt and save the recovery file
+    println!("Enter a passphrase to encrypt the recovery file:");
+    let passphrase = rpassword::read_password()?;
```
SHAcollision, line 28: Nit: up to us if we want to take this chance to teach best practices for passwords. Maybe we should reject empty passphrases here before writing the recovery file. Right now I think an accidental `Enter`+`Enter` creates an effectively unprotected key file. Suggested check right after reading passphrase:
```rust
if passphrase.is_empty() {
    anyhow::bail!("Passphrase cannot be empty");
}
```

examples/rust/keygen.rs
```diff
@@ -0,0 +1,42 @@
...
+    println!("Confirm passphrase:");
+    let confirm = rpassword::read_password()?;
+
+    if passphrase != confirm {
+        anyhow::bail!("Passphrases do not match");
+    }
+
+    let recovery_bytes = create_recovery_file(&keypair, &passphrase);
+    std::fs::write(&cli.output, &recovery_bytes)?;
+    println!("Recovery file written to {:?}", cli.output);
```
SHAcollision: Codex suggested:
Small UX nit: prefer `cli.output.display()` over `{:?}` for user-facing path output.

SHAcollision (approved):
LGTM, merge ready. If anything I think the empty-passphrase guard is more important, but not blocking.

SHAcollision (approved):
Let's go!

PR #383, opened 2026-05-15
fix(homeserver): apply burst to request rate limits

andrei-21:
Add coverage for PathLimit conversion to governor quotas with explicit
burst overrides.

Store prepared request rate limit tuples in the layer and propagate
invalid quota configuration errors during app creation.

### DIFF
--- a/pubky-homeserver/src/client_server/app.rs
+++ b/pubky-homeserver/src/client_server/app.rs
@@ -44,13 +44,10 @@ pub enum ClientServerBuildError {
     PubkyTlsServer(anyhow::Error),
     /// Failed to convert the data directory to an AppContext.
     #[error("AppContext conversion error: {0}")]
-    AppContext(AppContextConversionError),
-}
-
-impl From<AppContextConversionError> for ClientServerBuildError {
-    fn from(error: AppContextConversionError) -> Self {
-        ClientServerBuildError::AppContext(error)
-    }
+    AppContext(#[from] AppContextConversionError),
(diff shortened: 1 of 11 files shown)


pubky-homeserver/src/client_server/app.rs
```diff
@@ -124,7 +123,7 @@ impl ClientServer {
             user_service: context.user_service.clone(),
             default_storage_mb: context.config_toml.storage.default_quota_mb,
         };
-        super::create_app(state.clone(), context)
+        super::create_app(state.clone(), context).map_err(ClientServerBuildError::RequestRateLimits)
```
86667: This feels like the wrong place for this mapping. its fine for now for `RequestRateLimitLayer` to not define its own error type, but maybe lets move the mapping to the call site of the layer, ie create_app()?

86667 (changes requested):
Nice. Would be good to add the same build time failure to `BandwidthQuota` too. Happy to leave for now though to get these changes in first since they fix a separate bug in which `burst` is not applied to per-request limits.

Just the one change requested.

pubky-homeserver/src/client_server/app.rs
andrei-21: I agree, done.

andrei-21:
> Would be good to add the same build time failure to `BandwidthQuota` too. 

I will address this in a following PR.

86667 (approved):

86667 (approved):

PR #400, opened 2026-05-26
feat(sdk): restore grant session

SeverinAlexB:
Adds the ability to save a grant session by exporting a string token and restore a grant session by importing it again.

```rust
let secret_token = session.as_grant().unwrap().export_secret().await;
let restored = pubky.restore_session(&secret_token).await.unwrap();
```

`pubky.restore_session` is a general facade method that takes either a grant token or a cookie token to restore a session.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,6 +1594,7 @@ checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 name = "e2e"
 version = "0.9.0"
 dependencies = [
+ "base64 0.22.1",
  "bytes",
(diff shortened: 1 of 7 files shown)


86667 (changes requested):
Looks reasonable. Missing JS side?

SeverinAlexB:
> Missing JS side?

JS side is it's own task which comes next. I want to keep PRs simple and sweet.

I opened this issue https://github.com/pubky/pubky-core/issues/401 for better tracking.

86667 (commented):
AI noticed that there are no e2e test for attempting to restore an expired or revoked grant. Worth adding imo

86667 (approved):

PR #406, opened 2026-05-27
feat: OpenAPI yaml

### LINKED ISSUE #369: [homeserver]: Add openapi.yaml

86667:
Generated with AI, checked with AI over many iterations.

### DIFF
--- a/pubky-homeserver/openapi.yml
+++ b/pubky-homeserver/openapi.yml
@@ -1,0 +1,1323 @@
+openapi: 3.1.0
+info:
+  title: Pubky Homeserver API
+  description: |
+    The Pubky Homeserver exposes three servers:
+    - **Client Server** — Primary API for user authentication, file storage, and event streaming.
+    - **Admin Server** — Password-protected administration endpoints.
+    - **Metrics Server** — Prometheus metrics (`GET /metrics`).
+
+    This specification covers the Client Server and Admin Server.
+
+    ## Tenant Identification
+
+    Per-tenant (user) routes resolve the target user from these sources (highest priority first):
+    1. `pubky-host` header (explicit override; takes precedence over `Host`)
+    2. `Host` header (e.g. from TLS SNI on the Pubky socket)
+    3. `?pubky-host=` query parameter (fallback when no key found in headers)
+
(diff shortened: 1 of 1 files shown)


MCarlomagno (commented):
Overall LGTM, just 2 comments
- `/admin/...` paths don't seem to resolve to actual endpoints (maybe I'm missing something, like a reverse proxy that adds the `/admin` prefix?)
- I think we are missing the recently added GET `/signup_tokens` endpoint

andrei-21:
> Overall LGTM, just 2 comments
> 
>     * `/admin/...` paths don't seem to resolve to actual endpoints (maybe I'm missing something, like a reverse proxy that adds the `/admin` prefix?)

>     description: |
>      Admin server endpoints (password-protected). These run on a **separate server**
>      (`localhost:6288` by default). Paths use an `/admin` prefix to avoid collisions
>      with client server paths in this spec (e.g. both servers have `GET /`).
>      The actual admin server paths omit this prefix
>      (e.g. `/admin/info` → `GET /info` on the admin port).

pubky-homeserver/openapi.yml
```diff
@@ -0,0 +1,1323 @@
...
+    description: Event streaming and historical feeds
+  - name: Signup Tokens
+    description: Signup token management
+  - name: Admin
+    description: |
+      Admin server endpoints (password-protected). These run on a **separate server**
+      (`localhost:6288` by default). Paths use an `/admin` prefix to avoid collisions
+      with client server paths in this spec (e.g. both servers have `GET /`).
+      The actual admin server paths omit this prefix
+      (e.g. `/admin/info` → `GET /info` on the admin port).
```
andrei-21: It is confusing. There are no collisions except `/`, can we just ignore admin `/` endpoint and then we can drop the confusing `/admin` prefix?

MCarlomagno (approved):

pubky-homeserver/openapi.yml
andrei-21: Or we just define two files, one for the admin interface, one for the client?

86667: Good point, thanks. Removed.

andrei-21 (approved):

PR #411, opened 2026-05-29
feat(sdk): Webcrypto delegated PoP key + Browser session store

SeverinAlexB:
## Summary

Adds browser-aware grant auth persistence to the JS SDK.

The SDK can now use non-extractable WebCrypto Ed25519 keys for delegated grant Proof of Possession (PoP signing when the browser supports the required secure-context APIs, and automatically falls back to local PoP keys when delegation is unavailable. Browser sessions can be persisted in IndexedDB through `pubky.browserSessionStore`, including multi-account listing and restoring.

## What Changed

- Added delegated grant PoP signing support backed by browser-held WebCrypto keys.
- Added `BrowserSessionStore` for IndexedDB session persistence.
  - Stores delegated sessions without exporting private PoP key material.
  - Stores local fallback sessions when delegation is unavailable.
  - Supports multiple accounts via `save`, `list`, `restore`, `remove`, and `clear`.
- Consolidates browser auth storage into a single IndexedDB database: `pubky-auth`.
- Updated `Pubky.startGrantAuthFlow(...)` to choose delegated vs local grant auth automatically.
- Added low-level delegated signer plumbing in the Rust SDK so delegated signing is not hard-coupled to WebCrypto internally.
- Split JWS signing into signing-input construction and signature finalization so external signers can produce compatible JWS values.
- Updated JS docs and tests for the async `startGrantAuthFlow(...)` API.
- Updated browser test runtime to a newer Electron version with Ed25519 WebCrypto support.


pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -211,14 +238,29 @@ impl GrantCredential {
     /// revoked.
     pub async fn export_secret(&self) -> String {
         let state = self.state.lock().await;
+        let Some(client_key_secret) = state.client_signer.local_secret() else {
+            return String::new();
+        };
```
andrei-21:
```suggestion
        let client_key_secret = state.client_signer.local_secret().unwrap_or_default();
```

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -211,14 +238,29 @@ impl GrantCredential {
     /// revoked.
     pub async fn export_secret(&self) -> String {
```
andrei-21: Shouldn't it return `Option<String>` now, since it is only applicable for local signer, not delegated?

pubky-sdk/src/actors/auth/grant/flow.rs
```diff
@@ -172,20 +185,39 @@ impl PubkyGrantAuthFlow {
...
     #[must_use]
-    pub fn save(&self) -> GrantAuthFlowState {
+    pub fn save_local(&self) -> GrantAuthFlowState {
         GrantAuthFlowState {
             authorization_url: self.authorization_url().to_string(),
-            client_key_secret: self.client_keypair.secret(),
+            client_key_secret: self
+                .client_signer
+                .local_secret()
+                .expect("local grant flow state requires an exportable signer"),
```
andrei-21: Shouldn't it return `Option<GrantAuthFlowState>` instead of panicking, similar to `save_delegated()`?

pubky-sdk/bindings/js/src/actors/session_store.rs
```diff
@@ -0,0 +1,429 @@
...
+            BrowserGrantKeyStore::delete_key(state.key_id).await?;
+        }
+
+        Ok(())
+    }
+
+    /// Clear all local stored session records for this origin.
+    ///
+    /// This removes session records. Delegated keys may remain if IndexedDB cleanup
+    /// fails midway; use `remove(id)` for precise delegated key deletion.
```
andrei-21: It is not clear, will the method remove delegated keys from IndexedDB or not?

pubky-sdk/bindings/js/pkg/README.md
```diff
@@ -187,16 +187,34 @@ const storage = session.storage; // -> This User's storage API (absolute paths)
...
+const session = await flow.awaitApproval();
+await pubky.browserSessionStore.save(session);
+```
+
+For advanced/manual storage, grant sessions still expose `exportSecret()` and
+`pubky.restoreSession(secret)`. Treat that string like a bearer token.
+
 > `session.export()` is still available for legacy cookie sessions, but new
-> applications should use grant auth plus `exportSecret()`.
+> applications should use grant auth plus `browserSessionStore` or `exportSecret()`.
```
andrei-21: `exportSecret()` should not be mentioned here?

pubky-sdk/bindings/js/src/actors/session_store.rs
SeverinAlexB: Updated. It only cleans the stored session keys. Added a new method `clearAll()` that deletes all keys.
Key handling is a bit tricky because each session creates a new key but its never clear when the session ends because it can always be restored.

pubky-sdk/bindings/js/src/actors/grant_auth_flow.rs
```diff
@@ -125,27 +185,81 @@ impl GrantAuthFlow {
         Ok(PubkyGrantAuthFlow::restore(state, client)?.into())
     }
 
+    /// Resume a previously saved pending delegated grant auth flow.
+    ///
+    /// Runtime: delegated grant keys require a secure browser context with
+    /// WebCrypto `crypto.subtle` and IndexedDB. The saved `keyId` must still
+    /// exist in IndexedDB for the same origin. Unsupported runtimes reject with
```
andrei-21, line 190: >  The saved `keyId` must still exist in IndexedDB for the same origin.

Shouldn't we validate it in `resume_delegated_with_client()`?

pubky-sdk/bindings/js/src/actors/browser_grant_key_store.rs
```diff
@@ -0,0 +1,409 @@
...
+/**
+ * Persist a delegated-key record.
+ *
+ * The private key is a non-extractable CryptoKey, so IndexedDB stores a browser
+ * handle rather than raw private key bytes.
+ */
+async function putRecord(record) {
+  await withGrantKeyStore("readwrite", (store, resolve, reject) => {
+    const request = store.put(record);
+    request.onsuccess = () => resolve();
```
andrei-21: It is a bit beyond my mental ability to reason about async JS code, but AI claims that we should not `resolve()` on request success here because `withGrantKeyStore()` resolves on transaction completion. Resolving from `request.onsuccess` can report success before the IndexedDB transaction commits, so a later transaction abort would be missed.

Same thing applies to `withSessionStore()`.

andrei-21 (changes requested):

pubky-sdk/bindings/js/src/actors/browser_grant_key_store.rs
SeverinAlexB: I didn't check this code yet so good call. I refactored it. These are hard ones to test/write.

pubky-sdk/bindings/js/src/actors/grant_auth_flow.rs
```diff
@@ -125,27 +185,88 @@ impl GrantAuthFlow {
...
+        })?;
+        serde_json::to_string(&state).map_err(|e| {
             PubkyError::new(
                 PubkyErrorName::InternalError,
                 format!("Failed to serialize grant auth flow state: {e}"),
             )
         })
     }
 
+    /// Save non-secret state required to resume this delegated pending grant flow.
```
andrei-21: Hm. Seems that there is sensitive information.
The returned JSON includes the authorization URL, which carries the relay secret for the pending approval. Maybe it is not a big deal, but worth explicitly documenting.

SeverinAlexB: Fixed. The secret is not the biggest deal but you are technically correct.

SeverinAlexB:
By the way, the whole grant auth API surface will undergo a final review by me and Tomos before we publish. So any docstring or missing capability will hopefully be detected then latest.

andrei-21 (approved):

PR #414, opened 2026-06-02
chore(deps): separate library dependency ranges

andrei-21:
Use broad dependency requirements in published library crates while
keeping workspace dependency versions for app and test crates.

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -6,13 +6,13 @@ version = "0.9.0"
 publish = false
 
 [dependencies]
-pubky-testnet = { path = "../pubky-testnet" , version = "0.9.0" }
-pubky-common = { path = "../pubky-common" , version = "0.9.0" }
+pubky-testnet = { path = "../pubky-testnet", version = "0.9" }
+pubky-common = { path = "../pubky-common", version = "0.9" }
 base64 = "0.22"
 tokio = { workspace = true, features = ["full", "test-util"] }
 tracing-subscriber.workspace = true
 pkarr = { workspace = true }
-bytes = "1.10.1"
+bytes = "1"
 rand = { workspace = true }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/examples/rust/Cargo.toml
+++ b/examples/rust/Cargo.toml
@@ -48,15 +48,15 @@ docker-postgres = ["pubky-testnet/docker-postgres"]
 embedded-postgres = ["docker-postgres"]
 
 [dependencies]
-futures-util = "0.3.31"
+futures-util = "0.3"
 anyhow.workspace = true
-base64 = "0.22.1"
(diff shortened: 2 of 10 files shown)


SeverinAlexB (approved):
Can't see any obvious flaws so let's go

SeverinAlexB (approved):
LGTM. @ok300 should also quickly check though

e2e/Cargo.toml
```diff
@@ -6,13 +6,13 @@ version = "0.9.0"
 publish = false
 
 [dependencies]
-pubky-testnet = { path = "../pubky-testnet" , version = "0.9.0" }
-pubky-common = { path = "../pubky-common" , version = "0.9.0" }
+pubky-testnet = { path = "../pubky-testnet", version = "0.9" }
```
ok300, line 9: NIT: I don't think you need to specify the version for dependencies with relative path inside the workspace. It will use the code from that path from this commit, regardless of specified version.

examples/rust/Cargo.toml
```diff
@@ -48,15 +48,15 @@ docker-postgres = ["pubky-testnet/docker-postgres"]
 embedded-postgres = ["docker-postgres"]
 
 [dependencies]
-futures-util = "0.3.31"
+futures-util = "0.3"
```
ok300, line 51: NIT: this is referenced several times, could probably be extracted in `workspace` then
```
futures-util.workspace = true
```

There are a few other such cases, like `base64`, `futures-util`, etc.

pubky-sdk/bindings/js/Cargo.toml
```diff
@@ -29,21 +25,22 @@ name = "bundle_npm"
 path = "scripts/bundle_npm.rs"
 
 [dependencies]
-wasm-bindgen = "0.2.108"
-console_log = { version = "1.0.0", features = ["color"] }
+wasm-bindgen = "0.2"
```
ok300, line 28: NIT / FYI: these are very tricky ones.

Patch version updates often caused havoc in the JS bindings for `pubky-app-specs`. See https://github.com/pubky/pubky-app-specs/pull/82 . For this reason, we don't bump even the patch versions of these js dependencies in `pubky-app-specs` unless absolutely necessary.

ok300 (approved):
A few NITs, but overall LGTM and can be merged as-is.

e2e/Cargo.toml
andrei-21: If `pubky-homeserver` is published, the version lets Cargo know which registry version of `pubky-common` should be used by downstream users. The local path is for development; the version is the publishable dependency requirement.

examples/rust/Cargo.toml
andrei-21: It is going to be addressed in https://github.com/pubky/pubky-core/pull/417.

PR #416, opened 2026-06-04
feat: Add docker build pipeline

SpontaneousOverthrow:


SHAcollision (approved):
LGTM 🚀 

1 Nit: our Release workflow rejects tags on branches that are not `main` while this workflow will accept a tag pushed from a non-main commit and publish Docker images even if the regular release workflow later rejects that tag. The risk is that we could create divergent releases: GitHub releases/artifacts are protected by the main-branch check, Docker Hub images are not.

PR #417, opened 2026-06-05
chore(cargo): centralize workspace manifest metadata

andrei-21:
Move repeated package version, author, license, repository, and
dependency declarations into the workspace manifest. Update member
crates to inherit shared Cargo metadata and dependency versions.

### DIFF
--- a/.scripts/set-version.sh
+++ b/.scripts/set-version.sh
@@ -9,11 +9,20 @@ set -e # fail the script if any command fails
 set -u # fail the script if any variable is not set
 set -o pipefail # fail the script if any pipe command fails
 
-# Check if cargo-set-version is installed
-if ! cargo --list | grep -q "set-version"; then
-  echo "Error: cargo-set-version is not installed but required."
-  echo "Please install it first by running:"
-  echo "  cargo install cargo-set-version"
+# Check if cargo set-version is installed from cargo-edit.
+if ! SET_VERSION_OUTPUT=$(cargo set-version --version 2>/dev/null); then
+  echo "Error: cargo set-version is not installed but required."
+  echo "Please install it from cargo-edit first by running:"
+  echo "  cargo install cargo-edit --no-default-features --features set-version"
+  exit 1
+fi
+
+if [[ ! "$SET_VERSION_OUTPUT" =~ ^cargo-edit-set-version[[:space:]][0-9] ]]; then
+  echo "Error: cargo set-version must be provided by cargo-edit."
+  echo "Expected 'cargo set-version --version' to return 'cargo-edit-set-version <version>'."
+  echo "Found: $SET_VERSION_OUTPUT"
+  echo "Please install it from cargo-edit by running:"
+  echo "  cargo install cargo-edit --no-default-features --features set-version"
   exit 1
 fi
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ default-members = [
 resolver = "2"
 
 [workspace.package]
+version = "0.9.0"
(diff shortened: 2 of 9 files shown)


SeverinAlexB:
Without having reviewed the change completely, I think we had a reason to define the versions individually. Maybe it was the version bump script that required it? Can you check?

andrei-21:
> Without having reviewed the change completely, I think we had a reason to define the versions individually. Maybe it was the version bump script that required it? Can you check?

Right, the standalone `cargo-set-version` command does not respect workspace version references, so I changed it to the `cargo-edit` implementation instead.

.scripts/set-version.sh
```diff
@@ -9,11 +9,20 @@ set -e # fail the script if any command fails
...
-  echo "Error: cargo-set-version is not installed but required."
-  echo "Please install it first by running:"
-  echo "  cargo install cargo-set-version"
+# Check if cargo set-version is installed from cargo-edit.
+if ! SET_VERSION_OUTPUT=$(cargo set-version --version 2>/dev/null); then
+  echo "Error: cargo set-version is not installed but required."
+  echo "Please install it from cargo-edit first by running:"
+  echo "  cargo install cargo-edit --no-default-features --features set-version"
+  exit 1
+fi
```
SeverinAlexB, line 18: This command is broken. I tried it myself and it messed up the Cargo.tomls.

AI output:

```
.scripts/set-version.sh:13 uses cargo set-version --version as an install/provider check, but this subcommand treats --version as the target version and mutates manifests. I reproduced this: it rewrote member Cargo.toml versions to "--version" before failing. The release script can corrupt the workspace before it even validates NEW_VERSION or asks for confirmation.
```

.scripts/set-version.sh
```diff
@@ -9,11 +9,20 @@ set -e # fail the script if any command fails
...
+
+if [[ ! "$SET_VERSION_OUTPUT" =~ ^cargo-edit-set-version[[:space:]][0-9] ]]; then
+  echo "Error: cargo set-version must be provided by cargo-edit."
+  echo "Expected 'cargo set-version --version' to return 'cargo-edit-set-version <version>'."
+  echo "Found: $SET_VERSION_OUTPUT"
+  echo "Please install it from cargo-edit by running:"
+  echo "  cargo install cargo-edit --no-default-features --features set-version"
   exit 1
```
SeverinAlexB, line 26: AI review:

```
.scripts/set-version.sh:61 is incompatible with the new version.workspace = true centralization. Running cargo set-version <version> rewrites member manifests back to explicit version = "<version>" and adds per-member dependency versions, rather than updating Cargo.toml:21 [workspace.package].version and the centralized workspace dependency versions at Cargo.toml:60-63. That defeats this PR’s centralization and can leave release metadata inconsistent.
```

It was a long time ago when I wrote this version script but I remember there was no do it all version bump helper.

SeverinAlexB (changes requested):
Version script is broken currently. I remember this was quite tricky to make it work.

SeverinAlexB (commented):
LGTM. One note: When I execute the set-version script, it will ask if I am sure (y/N). Will this script work with the Github action pipeline therefore?

Image

andrei-21:
> LGTM. One note: When I execute the set-version script, it will ask if I am sure (y/N). Will this script work with the Github action pipeline therefore?
> Image

It was like this before me:
```sh
# Ask for confirmation to update the version
read -p "Are you sure you want to set the version to $NEW_VERSION? (y/N) " -n 1 -r
```

SeverinAlexB (approved):
Then I guess it should work. LGTM

PR #420, opened 2026-06-10
feat: add list signup tokens endpoint

### LINKED ISSUE #402: Admin API: GET /signup_tokens, list signup tokens with usage info
## Summary

Per-token usage data (which signup token was consumed by which user) is needed for invite management: auditing, revocation, and tracing abuse. Today only aggregate counts are exposed via `/info`. The homeserver dashboard will surface this in the Invites tab as soon as the endpoint is available.

## Endpoint

`GET /signup_tokens?limit=<n>&cursor=<opaque>&state=<used|unused|all>`

Auth: existing `X-Admin-Password`. Pagination matches `GET /users/disabled`.

### Response 200

```json
{
  "items": [
    { "token": "abc123...", "created_at": "...", "used_at": "...", "used_by": "<pubkey>" },
    { "token": "def456...", "created_at": "...", "used_at": null, "used_by": null }
  ],
  "next_cursor": null
}
```

### Errors

`401` on auth, `500` on DB error.

## Open questions

- Single-use or reusable tokens? Affects whether `used_by` is scalar or array.
- TTL/expiration field? None enforced today.

MCarlomagno:
## Summary

Closes #402.

Adds an authenticated admin endpoint to list signup tokens with pagination and usage metadata.

## What Changed

- Added `GET /signup_tokens` to the admin API.
- Supports query params:
  - `state=all|used|unused`
  - `limit`
  - `cursor`
- Returns signup token records with:
  - `token`
  - `created_at`
  - `used_at`
  - `used_by`
- Added `used_at` tracking when a signup token is marked as used.
- Added a migration to add nullable `signup_codes.used_at`.

## Notes on open questions

Signup tokens remain single-use: `used_by` is still scalar. This PR does not add TTL or expiration behavior.

### DIFF
--- a/pubky-homeserver/src/admin_server/app.rs
+++ b/pubky-homeserver/src/admin_server/app.rs
@@ -278,6 +278,33 @@ mod tests {
         assert_eq!(body["next_cursor"], serde_json::Value::Null);
     }
 
+    #[tokio::test]
+    #[pubky_test_utils::test]
+    async fn test_list_signup_tokens_query_params_success() {
+        let context = AppContext::test().await;
+        let server = create_test_server(&context);
+
+        let response = server
+            .get("/generate_signup_token")
+            .add_header("X-Admin-Password", "test")
+            .expect_success()
+            .await;
+        let token = response.text();
+
+        let response = server
+            .get("/signup_tokens?state=unused&limit=1")
+            .add_header("X-Admin-Password", "test")
+            .expect_success()
+            .await;
+        response.assert_status_ok();
+
+        let body: serde_json::Value = response.json();
+        let items = body["items"].as_array().unwrap();
+        assert_eq!(items.len(), 1);
+        assert_eq!(items[0]["token"], token);
+        assert_eq!(body["next_cursor"], serde_json::Value::Null);
+    }
+
     fn auth_header() -> String {
         // AppState is created with password "" in create_test_server
         let auth = base64::engine::general_purpose::STANDARD.encode("admin:");
--- a/pubky-homeserver/src/admin_server/app.rs
+++ b/pubky-homeserver/src/admin_server/app.rs
@@ -188,9 +188,12 @@ mod tests {
     use axum::http::Method;
     use axum_test::TestServer;
     use base64::Engine;
+    use pubky_common::crypto::Keypair;
 
     use crate::data_directory::quota_config::BandwidthQuota;
     use crate::persistence::files::FileService;
+    use crate::persistence::sql::signup_code::{SignupCode, SignupCodeRepository};
+    use crate::shared::user_quota::UserQuota;
 
     use super::*;
 
@@ -284,13 +287,28 @@ mod tests {
         let context = AppContext::test().await;
         let server = create_test_server(&context);
 
-        let response = server
-            .get("/generate_signup_token")
-            .add_header("X-Admin-Password", "test")
-            .expect_success()
-            .await;
-        let token = response.text();
+        let token1 = SignupCode::new("0000-0000-0001".to_string()).unwrap();
+        let token2 = SignupCode::new("0000-0000-0002".to_string()).unwrap();
+        let token3 = SignupCode::new("0000-0000-0003".to_string()).unwrap();
+        let token4 = SignupCode::new("0000-0000-0004".to_string()).unwrap();
 
+        for token in [&token1, &token2, &token3, &token4] {
+            SignupCodeRepository::create(
+                token,
+                &UserQuota::default(),
+                &mut context.sql_db.pool().into(),
+            )
+            .await
+            .unwrap();
+        }
+
+        let used_by = Keypair::random().public_key();
+        SignupCodeRepository::mark_as_used(&token1, &used_by, &mut context.sql_db.pool().into())
+            .await
+            .unwrap();
+
+        // With three unused tokens, limit=1 proves the page size is applied and
+        // returns the last item in the page as the cursor.
         let response = server
             .get("/signup_tokens?state=unused&limit=1")
             .add_header("X-Admin-Password", "test")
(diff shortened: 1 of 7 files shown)


pubky-homeserver/src/admin_server/app.rs
```diff
@@ -239,15 +240,42 @@ mod tests {
 
     #[tokio::test]
     #[pubky_test_utils::test]
-    async fn test_generate_signup_token_success() {
+    async fn test_list_signup_tokens_fail() {
```
MCarlomagno, line 246: here I replaced the previous "generate token success" test by a "generate and list token success" (which implicitly tests if token generation works)

pubky-homeserver/src/persistence/sql/entities/signup_code.rs
```diff
@@ -313,6 +412,13 @@ mod tests {
 
     use super::*;
 
+    async fn current_db_timestamp(db: &SqlDb) -> sqlx::types::chrono::NaiveDateTime {
```
MCarlomagno, line 436: auxiliary method for testing used_at timestamp was correctly set after using the token

pubky-homeserver/src/persistence/sql/migrations/m20260609_add_signup_code_used_at.rs
```diff
@@ -0,0 +1,21 @@
...
+pub struct M20260609AddSignupCodeUsedAtMigration;
+
+#[async_trait]
+impl MigrationTrait for M20260609AddSignupCodeUsedAtMigration {
+    async fn up(&self, tx: &mut Transaction<'static, sqlx::Postgres>) -> anyhow::Result<()> {
+        sqlx::query("ALTER TABLE signup_codes ADD COLUMN IF NOT EXISTS used_at TIMESTAMP")
+            .execute(&mut **tx)
+            .await?;
+        Ok(())
+    }
```
MCarlomagno, line 21: here I'm wondering if we want to backfill existing codes setting some generic `used_at` when `used_by` exists? @SeverinAlexB wdyt?

pubky-homeserver/src/admin_server/routes/signup_tokens.rs
```diff
@@ -0,0 +1,286 @@
...
+
+impl SignupTokensQuery {
+    fn state(&self) -> HttpResult<SignupCodeListState> {
+        match self.state.as_deref().unwrap_or("all") {
+            "all" => Ok(SignupCodeListState::All),
+            "used" => Ok(SignupCodeListState::Used),
+            "unused" => Ok(SignupCodeListState::Unused),
+            _ => Err(HttpError::bad_request("Invalid state")),
+        }
+    }
```
SeverinAlexB: NIT: If you make `SignupCodeListState` deserializable, you can add it directly into `state: Option<SignupCodeListState>`. This would be cleaner compared to the parsing in the query.

The same goes for the cursor.`SignupCode` can be used directly if we make it deserializable.

pubky-homeserver/src/admin_server/routes/signup_tokens.rs
```diff
@@ -0,0 +1,286 @@
...
+            used_by: code.used_by.map(|pubkey| pubkey.z32()),
+        }
+    }
+}
+
+#[derive(Serialize)]
+pub(crate) struct SignupTokensResponse {
+    items: Vec<SignupTokenItem>,
+    next_cursor: Option<String>,
+}
```
SeverinAlexB, line 59: NIT: `token` and `next_cursor` could be of type SignupCode directly if you make it serializable.

pubky-homeserver/src/persistence/sql/migrations/m20260609_add_signup_code_used_at.rs
SeverinAlexB: Sounds good. Maybe we can set it to epoch=0 (earliest possible timestamp) or something like 1st Jan 2000 to make it recognizable that these have been backfilled.

pubky-homeserver/src/admin_server/app.rs
```diff
@@ -239,15 +240,42 @@ mod tests {
...
         let context = AppContext::test().await;
         let server = create_test_server(&context);
+
+        let response = server.get("/signup_tokens").expect_failure().await;
+        response.assert_status_unauthorized();
+    }
+
+    #[tokio::test]
+    #[pubky_test_utils::test]
+    async fn test_create_and_list_signup_token_success() {
```
SeverinAlexB, line 256: Can we get some simple query param test coverage as well? Doesn't need to be extensive but we should see if something breaks.

SeverinAlexB (changes requested):
Mostly NITS, otherwise LGTM

pubky-homeserver/src/persistence/sql/migrations/m20260609_add_signup_code_used_at.rs
MCarlomagno: I'm thinking that the best option is to backfill `used_at = created_at` that might make more sense

pubky-homeserver/src/persistence/sql/entities/signup_code.rs
```diff
@@ -3,18 +3,51 @@ use std::{fmt::Display, str::FromStr};
...
+pub enum SignupCodeListState {
+    All,
+    Used,
+    Unused,
+}
+
+#[derive(Debug, Clone)]
+pub struct SignupCodeListQuery {
+    pub state: SignupCodeListState,
+    pub limit: Option<u16>,
```
SHAcollision, line 33: Just passing by here :)

Found a NIT here. Current behavior appears to return an empty page with `next_cursor: null` when rows exist and `limit=0`. That could make a client conclude pagination is complete. Recommended options if you want to tackle this one:
- reject `limit=0` with `400 Bad Request`; or
- clamp to a minimum of `1`.

SHAcollision (commented):
Looks good.

pubky-homeserver/src/admin_server/app.rs
```diff
@@ -239,15 +240,69 @@ mod tests {
...
+        assert_eq!(items[0]["token"], token);
+        assert!(items[0]["created_at"].as_str().is_some());
+        assert_eq!(items[0]["used_at"], serde_json::Value::Null);
+        assert_eq!(items[0]["used_by"], serde_json::Value::Null);
+        assert_eq!(body["next_cursor"], serde_json::Value::Null);
+    }
+
+    #[tokio::test]
+    #[pubky_test_utils::test]
+    async fn test_list_signup_tokens_query_params_success() {
```
SeverinAlexB, line 286: Creating 1 signup token and then checking if limit=1 works is not a great test as it will always return 1 independent of limit is 1 or 9. 
The same goes for `unused` as there is only one unused code there anyway.
At the same time, the cursor param is not tested as far as I can see?

SeverinAlexB (changes requested):
One thing otherwise this is good to go

pubky-homeserver/src/admin_server/app.rs
MCarlomagno: Agree, I just expanded the test to cover better those cases

SeverinAlexB (approved):
Let's go!

PR #423, opened 2026-06-11
fix(sdk): resolve transport with the full _pubky.<key> qname so ICANN fallback applies

### LINKED ISSUE #422: Native SDK: ICANN fallback seems to never apply for _pubky.<key> hosts (transport resolver queries the bare key)
**Summary**

Observed on main (f9cb1a3, pkarr 6.0.1). When a homeserver advertises an unreachable direct endpoint (e.g. a private IP) plus a reachable ICANN endpoint, signup succeeds but signin/storage fail at the transport layer:

```
error sending request for url (https://_pubky.<userkey>/session)
  tcp connect error: Connection timed out (os error 110)
```

The same operation done manually over the ICANN endpoint (`POST https://<icann-domain>/session` + `pubky-host: <userkey>`) works, so the homeserver appears fine, the issue seems limited to the SDK's transport selection for `_pubky.<userkey>` hosts.

**Suspected root cause**

In `cross_request` -> `prepare_request` (`pubky-sdk/src/client/http_targets/native.rs`), the `_pubky.` prefix is stripped and the bare key is handed to `TransportResolver`, the only place ICANN fallback is implemented. But the user's HTTPS record is published at `_pubky.<key>`, not the apex (see `build_homeserver_packet` in `pkdns.rs`), so resolving the bare key returns no endpoints and the resolver defaults to `PubkyTls`, the ICANN branch is never evaluated. Checked with pkarr directly:

```
resolve_https_endpoints("<userkey>")         -> <no endpoints>
resolve_https_endpoints("_pubky.<userkey>")  -> [direct 10.21.0.23:6287, ICANN <icann-domain>]
```

This would also explain why signup works: `build_signup_url` targets `https://<homeserver_z32>/signup`, the homeserver's own apex, where its endpoints live, so there the resolver finds both endpoints and falls back correctly.

**Possible fix**

Feed the full `_pubky.<key>` qname to `TransportResolver`/`resolve_https_endpoints` (so the endpoint set includes the ICANN fallback), keeping the bare key only as the `pubky-host` header value, i.e. thread the resolution qname and header value separately through `cross_request`/`build_pubky_request`. A regression test with an unreachable direct endpoint + reachable ICANN endpoint would cover it.

SHAcollision:
Fixes #422

Confirmed via red test -> green test

### DIFF
--- a/pubky-sdk/src/client/http_targets/native.rs
+++ b/pubky-sdk/src/client/http_targets/native.rs
@@ -37,42 +37,42 @@ impl TransportResolver {
         }
     }
 
-    /// Look up the transport for `pk`, resolving via PKARR on cache miss.
-    pub(crate) async fn resolve(&self, pk: &str, pkarr: &pkarr::Client) -> ResolvedTransport {
-        if let Some(t) = self.cached(pk) {
+    /// Look up the transport for `qname`, resolving via PKARR on cache miss.
+    pub(crate) async fn resolve(&self, qname: &str, pkarr: &pkarr::Client) -> ResolvedTransport {
+        if let Some(t) = self.cached(qname) {
             return t;
         }
-        self.resolve_and_cache(pk, pkarr).await
+        self.resolve_and_cache(qname, pkarr).await
     }
 
     /// Fast path: return a cached, non-expired transport decision.
-    fn cached(&self, pk: &str) -> Option<ResolvedTransport> {
+    fn cached(&self, qname: &str) -> Option<ResolvedTransport> {
         let cache = self.cache.read().unwrap_or_else(PoisonError::into_inner);
         cache
-            .get(pk)
+            .get(qname)
             .filter(|(ts, _)| ts.elapsed() < TRANSPORT_CACHE_TTL)
             .map(|(_, t)| t.clone())
     }
 
-    /// Slow path: acquire a per-key guard, double-check the cache, resolve,
+    /// Slow path: acquire a per-qname guard, double-check the cache, resolve,
     /// and store the result.
-    async fn resolve_and_cache(&self, pk: &str, pkarr: &pkarr::Client) -> ResolvedTransport {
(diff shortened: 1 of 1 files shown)


pubky-sdk/src/client/http_targets/native.rs
```diff
@@ -169,7 +169,11 @@ impl PubkyHttpClient {
         let Some(pk) = self.prepare_request(&mut url).await? else {
             return Ok(self.request(method, &url));
         };
-        let transport = self.transport.resolve(&pk, &self.pkarr).await;
+        // Resolve the transport with the full host: for `_pubky.<pk>` hosts the
+        // endpoints live under that qname, not the bare key apex. `pk` is only
+        // the `pubky-host` header value.
+        let qname = url.host_str().unwrap_or(&pk).to_string();
+        let transport = self.transport.resolve(&qname, &self.pkarr).await;
```
SHAcollision, line 176: These 5 lines are the CORE of the fix on this PR. The rest are small fixes on variable name `pk` -> `qname`

MCarlomagno (approved):
LGTM tests pass, 

just one question/NIT (non-blocking) about the fix:

```rust
let qname = url.host_str().unwrap_or(&pk).to_string();
````

here we are silently falling back to `&pk`, which was the failing behavior, do we want to do that or rather throw an error? like

```rust
let qname = url
    .host_str()
    .ok_or_else(...SomeError("URL must include a host"))
    .to_string();
```

SHAcollision:
Thanks for the review @MCarlomagno !

That's a good question. So, in fact, this is unreachable today, we only get here when `prepare_request` returned `Some(pk)`, and it derives `pk` from `url.host_str()` without mutating the URL, so the host is always present. But the fallback `.unwrap_or(&pk)` is intentional: if a future refactor ever breaks that invariant, `&pk` degrades gracefully, exact same value for bare-key hosts, pre-#422 routing (direct PubkyTls, still works in the common case) for `_pubky.` hosts. An error here would instead hard-fail every request over a latent invariant drift. For transport selection we prefer failing soft.

I'll proceed to merge :rocket:

PR #424, opened 2026-06-11
fix: path collisions

### LINKED ISSUE #419: Forbid File/Folder Path Collisions in Homeserver Storage
## Summary

Homeserver should reject storage paths that would make an exact file and an
implicit folder share the same base path.

Examples that should not both be allowed:

```text
/pub/app/foo
/pub/app/foo/bar.json
```

The SQL listing layer can currently represent both `/pub/app/foo` as a file and
`/pub/app/foo/` as an implicit folder. However, the filesystem storage backend
cannot reliably store both a file named `foo` and a directory named `foo` at the
same location. Since homeserver supports a normal filesystem backend, allowing
this shape creates backend-dependent behavior.

## Current Behavior

The current listing model derives folders from path prefixes. Repository tests
already create and list both an exact file and an implicit folder with the same
base path, such as:

```text
/test/sub1
/test/sub1/1/1.txt
```

This works at the metadata/listing layer, but it is not portable across storage
backends.

## Required Behavior

Homeserver should reject file/folder path collisions for app-facing storage
writes.

Rules:

- `PUT /pub/app/foo` must fail if any descendant exists under `/pub/app/foo/`.
- `PUT /pub/app/foo/bar.json` must fail if `/pub/app/foo` exists as a file.
- The response should be `409 Conflict`.
- The check should happen before writing bytes to the storage backend.

## Tests

Add tests for:

- create `/pub/app/foo`, then `PUT /pub/app/foo/bar.json` returns
  `409 Conflict`;
- create `/pub/app/foo/bar.json`, then `PUT /pub/app/foo` returns
  `409 Conflict`;
- deleting `/pub/app/foo` does not delete descendants under `/pub/app/foo/`;
- after deleting the conflicting exact file, writing descendants succeeds;
- after deleting all descendants, writing the exact file succeeds;

## Audit Query

Use this query to find existing exact-file paths that also have descendants:

```sql
SELECT
  u.public_key,
  p.path AS exact_file_path,
  c.path AS descendant_path
FROM entries AS p
JOIN entries AS c
  ON c."user" = p."user"
 AND substr(c.path, 1, length(p.path) + 1) = p.path || '/'
JOIN users AS u
  ON u.id = p."user"
WHERE p.path NOT LIKE '%/'
ORDER BY u.public_key, p.path, c.path;
```

Explanation:

- `p` is the possible exact file entry, for example `/pub/app/foo`.
- `c` is a possible descendant entry, for example `/pub/app/foo/bar.json`.
- `c."user" = p."user"` compares paths only within the same user.
- `p.path || '/'` builds the folder prefix that would conflict with the exact
  file. For `/pub/app/foo`, this becomes `/pub/app/foo/`.
- `substr(c.path, 1, length(p.path) + 1)` takes the beginning of the possible
  descendant path with the same length as that folder prefix.
- If those two strings are equal, `c.path` is below the implicit folder for
  `p.path`. For example, `/pub/app/foo/bar.json` starts with `/pub/app/foo/`.
- This avoids matching sibling names such as `/pub/app/foobar.json`, because
  that path does not start with `/pub/app/foo/`.
- `p.path NOT LIKE '%/'` limits `p` to exact file-style paths.

If the query returns rows, each row is a collision candidate.

MCarlomagno:
## Summary

Closes #419.

This PR prevents new app file/folder path collisions in homeserver storage. Writes now fail with `409 Conflict` when a path would become both an exact file and an implicit folder prefix.

Examples:

- If `/pub/app/foo` exists, writing `/pub/app/foo/bar.json` returns `409 Conflict`.
- If `/pub/app/foo/bar.json` exists, writing `/pub/app/foo` returns `409 Conflict`.
- Exact overwrites remain allowed.

## Changes

- Added a `PathCollisionLayer` to the app-facing OpenDAL storage stack.
- Kept `admin_operator` unchanged so admin WebDAV/delete paths can still repair legacy data.
- Added `EntryRepository::has_file_folder_collision` for same-user collision detection.

## Notes

This does not migrate or block existing deployed collisions. The guard only prevents new app-facing collisions.

Admin maintenance paths can still access/delete legacy data. Deleting an exact file path does not delete descendants that may exist from legacy/admin-created collisions.

### DIFF
--- a/pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
+++ b/pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
@@ -7,6 +7,7 @@ use crate::{
         files::{
             entry::entry_layer::EntryLayer,
             events::{EventsLayer, EventsService},
+            path_collision_layer::PathCollisionLayer,
             user_quota_layer::UserQuotaLayer,
             write_path_layer::WritePathLayer,
         },
@@ -43,9 +44,9 @@ pub fn build_storage_operators(
     let events_layer = EventsLayer::new(db.clone(), events_service);
     // Note: Layers ordering is important:
     // Layers are applied last-to-first: write_path_layer (outermost) runs first,
-    // rejecting disallowed writes before they reach quota/entry/event layers.
-    // events_layer runs after entry_layer.close() completes, guaranteeing the
-    // file is written before the Event is created.
+    // then path_collision_layer rejects file/folder collisions
+    // before they reach storage. events_layer runs after entry_layer.close()
+    // completes, guaranteeing the file is written before the Event is created.
     let admin_operator = match &storage_config.backend {
         StorageConfigToml::FileSystem => {
             let files_dir = match data_directory.join("data/files").to_str() {
@@ -91,6 +92,7 @@ pub fn build_storage_operators(
 
     let operator = admin_operator
         .clone()
+        .layer(PathCollisionLayer::new(db.clone()))
         .layer(WritePathLayer::new(user_service));
     Ok((operator, admin_operator))
 }
--- a/pubky-homeserver/src/persistence/files/path_collision_layer.rs
+++ b/pubky-homeserver/src/persistence/files/path_collision_layer.rs
@@ -73,6 +73,7 @@ impl<A: Access> LayeredAccess for PathCollisionAccessor<A> {
     }
 
     async fn create_dir(&self, path: &str, args: OpCreateDir) -> Result<RpCreateDir> {
+        self.check_no_path_collision(path).await?;
         self.inner.create_dir(path, args).await
     }
 
--- a/pubky-homeserver/src/persistence/files/path_collision_layer.rs
+++ b/pubky-homeserver/src/persistence/files/path_collision_layer.rs
@@ -1,0 +1,113 @@
+use std::sync::Arc;
+
+use crate::persistence::files::layer_domain_error::LayerDomainError;
+use crate::persistence::sql::{entry::EntryRepository, SqlDb};
+use crate::shared::webdav::EntryPath;
+use opendal::raw::*;
+use opendal::Result;
+
+/// OpenDAL layer that rejects app-facing file/folder path collisions before
+/// opening a backend writer.
+#[derive(Clone)]
+pub struct PathCollisionLayer {
+    db: SqlDb,
+}
+
+impl PathCollisionLayer {
+    pub fn new(db: SqlDb) -> Self {
+        Self { db }
+    }
+}
+
+impl<A: Access> Layer<A> for PathCollisionLayer {
+    type LayeredAccess = PathCollisionAccessor<A>;
+
+    fn layer(&self, inner: A) -> Self::LayeredAccess {
+        PathCollisionAccessor {
+            inner: Arc::new(inner),
+            db: self.db.clone(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct PathCollisionAccessor<A: Access> {
+    inner: Arc<A>,
+    db: SqlDb,
+}
+
+impl<A: Access> PathCollisionAccessor<A> {
+    async fn check_no_path_collision(&self, path: &str) -> Result<()> {
+        let entry_path = EntryPath::parse_opendal(path)?;
+        let has_collision =
+            EntryRepository::has_file_folder_collision(&entry_path, &mut self.db.pool().into())
+                .await
+                .map_err(|e| {
+                    opendal::Error::new(
+                        opendal::ErrorKind::Unexpected,
+                        format!("Failed to check path collision for {entry_path}: {e}"),
+                    )
+                })?;
+
+        if has_collision {
+            return Err(opendal::Error::new(
+                opendal::ErrorKind::AlreadyExists,
+                format!("File/folder path collision for {entry_path}"),
+            )
+            .set_source(LayerDomainError::PathCollision));
+        }
+
+        Ok(())
+    }
+}
+
+impl<A: Access> LayeredAccess for PathCollisionAccessor<A> {
+    type Inner = A;
+    type Reader = A::Reader;
+    type Writer = A::Writer;
+    type Lister = A::Lister;
+    type Deleter = A::Deleter;
+
+    fn inner(&self) -> &Self::Inner {
+        &self.inner
+    }
+
+    async fn create_dir(&self, path: &str, args: OpCreateDir) -> Result<RpCreateDir> {
+        self.inner.create_dir(path, args).await
+    }
+
+    async fn read(&self, path: &str, args: OpRead) -> Result<(RpRead, Self::Reader)> {
+        self.inner.read(path, args).await
+    }
+
+    async fn write(&self, path: &str, args: OpWrite) -> Result<(RpWrite, Self::Writer)> {
+        self.check_no_path_collision(path).await?;
+        self.inner.write(path, args).await
+    }
+
+    async fn copy(&self, from: &str, to: &str, args: OpCopy) -> Result<RpCopy> {
+        self.check_no_path_collision(to).await?;
+        self.inner.copy(from, to, args).await
+    }
+
+    async fn rename(&self, from: &str, to: &str, args: OpRename) -> Result<RpRename> {
+        self.check_no_path_collision(to).await?;
+        self.inner.rename(from, to, args).await
+    }
+
+    async fn stat(&self, path: &str, args: OpStat) -> Result<RpStat> {
+        self.inner.stat(path, args).await
+    }
+
+    async fn delete(&self) -> Result<(RpDelete, Self::Deleter)> {
+        self.inner.delete().await
+    }
+
+    async fn list(&self, path: &str, args: OpList) -> Result<(RpList, Self::Lister)> {
+        self.inner.list(path, args).await
+    }
+
+    async fn presign(&self, path: &str, args: OpPresign) -> Result<RpPresign> {
+        self.inner.presign(path, args).await
+    }
+}
--- a/pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
+++ b/pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
@@ -226,6 +226,67 @@ impl EntryRepository {
         Ok(count > 0)
     }
 
+    /// Check if writing `path` would make an exact file path collide with an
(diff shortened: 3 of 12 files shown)


pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -226,6 +226,67 @@ impl EntryRepository {
         Ok(count > 0)
     }
 
+    /// Check if writing `path` would make an exact file path collide with an
+    /// implicit folder path for the same user.
+    pub async fn has_file_folder_collision<'a>(
```
MCarlomagno, line 231: My only concern about this is that now we need to run this query before writing storage every time. But considering this is a write path, I guess thats an acceptable tradeoff

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -226,6 +226,67 @@ impl EntryRepository {
...
+        } else {
+            format!("{path_str}/")
+        };
+        let ancestor_paths = Self::ancestor_file_paths(path_str);
+
+        // Reject both collision directions for the same user:
+        // - existing descendants under `path/`
+        // - existing exact-file ancestors of `path`
+        let con = executor.get_con().await?;
+        sqlx::query_scalar(
```
andrei-21, line 254: The problem with this approach is that it is not atomic. Two requests from the same user can bypass this check at the same time (before being recorded) and then be recorded, violating the restriction.

andrei-21 (changes requested):

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
MCarlomagno: Good catch, yes I guess the best approach is to make it as a postgres tx instead

MCarlomagno: I'm thinking that if we want to implement this in a way that prevents concurrent writes + doesnt hold a DB connection during upload, is by having a separate table for signaling writings in progress:

1. before writing storage we mark the path + the user as "reserved"
2. when the same user wants to write storage, we query the "reservations" table  
- if reserved throw "storage collision error"
- if not reserved create "reservation" entry (this should happen atomically after querying)
3. once the file was fully uploaded, errored during upload or aborted, remove or disable "reservation" entry.
4. TTL of the reservation entry is X hs (I'm thinking 24hs is fine), just as a cleanup fallback 

does it make sense?

MCarlomagno: Just pushed those changes, maybe it adds too much complexity for the concurrency problem 😕 but lmk your thoughts

andrei-21: I do not feel comfortable having a reservation queue with releases and TTLs; it creates endless possibilities for different, hard-to-catch bugs.
Wouldn't it be simpler to add this check under the SQL transaction in `WriterWrapper<R>::close()`?
Trade-off: if the object write succeeds but the metadata commit is rejected, it may leave an orphaned backend object. That is already possible today when the DB write fails after backend close. We can attempt a best-effort backend delete on collision or DB failure.
Also, as an optimization, we can check for collisions before even attempting the write (similar to what you did at the beginning).

pubky-homeserver/src/persistence/sql/entities/user.rs
```diff
@@ -76,7 +78,7 @@ impl UserRepository {
             .from(USER_TABLE)
             .columns(ALL_USER_COLUMNS)
             .and_where(Expr::col(UserIden::PublicKey).eq(public_key.z32()))
-            .lock(sea_query::LockType::Update)
+            .lock(sea_query::LockType::NoKeyUpdate)
```
MCarlomagno: Using `FOR NO KEY UPDATE`, it still serializes same user write finalization, but avoids deadlocking/blocking the nested `EntryLayer` insert that needs a foreign key lock on the same user row,

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
```diff
@@ -59,9 +63,10 @@ pub fn build_storage_operators(
             };
             let builder = opendal::services::Fs::default().root(files_dir.as_str());
             opendal::Operator::new(builder)?
-                .layer(user_quota_layer)
                 .layer(entry_layer)
+                .layer(user_quota_layer)
                 .layer(events_layer)
+                .layer(path_collision_layer)
                 .finish()
```
andrei-21, line 66: @86667 double-check the order here please.

pubky-homeserver/src/persistence/files/path_collision_layer.rs
```diff
@@ -0,0 +1,130 @@
...
+    type Writer = A::Writer;
+    type Lister = A::Lister;
+    type Deleter = A::Deleter;
+
+    fn inner(&self) -> &Self::Inner {
+        &self.inner
+    }
+
+    async fn create_dir(&self, path: &str, args: OpCreateDir) -> Result<RpCreateDir> {
+        self.inner.create_dir(path, args).await
```
andrei-21, line 77: We do not really allow creating directories, but maybe for completeness it makes sense to add the check here?

pubky-homeserver/src/persistence/files/path_collision_layer.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::persistence::files::layer_domain_error::LayerDomainError;
+use crate::persistence::sql::{entry::EntryRepository, user::UserRepository, SqlDb};
+use crate::shared::webdav::EntryPath;
+use opendal::raw::*;
+use opendal::Result;
+
+/// Preflight that rejects committed file/folder path collisions before opening
+/// a backend writer.
+#[derive(Clone)]
+pub struct PathCollisionLayer {
```
andrei-21, line 12: I see no real value in this layer if we must recheck in `UserQuotaLayer` anyway. Maybe we should merge them into something like `UserStoragePolicyLayer`?

pubky-homeserver/src/persistence/files/opendal/opendal_service.rs
MCarlomagno: More context about the layer order:

old chain was: `.layer(user_quota).layer(entry).layer(events)`
current chain: `.layer(entry).layer(user_storage_policy).layer(events)` (`path_collision` and `user_quota` layers were merged in last commit)

OpenDAL applies layers last to first, then the current runtime order is:

`Events` -> `UserStoragePolicy` -> `Entry` -> backend

Why that order: `UserStoragePolicyLayer.close()` needs to hold the user DB lock while `EntryLayer.close()` writes the metadata. If we restore the old order, `EntryLayer.close()` would wrap around the policy layer, so the quota+collision check would happen before the entry metadata write, and the user lock could be released before the metadata row exists, that reopens the concurrent collision race.

MCarlomagno:
@andrei-21 @86667 I just rolled back to path collision pre-check only (using PathCollisionLayer), no concurrency safe yet

pubky-homeserver/src/persistence/files/path_collision_layer.rs
```diff
@@ -0,0 +1,113 @@
...
+    type Writer = A::Writer;
+    type Lister = A::Lister;
+    type Deleter = A::Deleter;
+
+    fn inner(&self) -> &Self::Inner {
+        &self.inner
+    }
+
+    async fn create_dir(&self, path: &str, args: OpCreateDir) -> Result<RpCreateDir> {
+        self.inner.create_dir(path, args).await
```
andrei-21, line 77: Again, let's add the check for `create_dir()`, even if we do not expose the method explicitly.

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -226,6 +226,67 @@ impl EntryRepository {
...
+        .bind(descendant_prefix)
+        .bind(path_str)
+        .bind(ancestor_paths)
+        .fetch_one(con)
+        .await
+    }
+
+    fn ancestor_file_paths(path: &str) -> Vec<String> {
+        let path = path.trim_end_matches('/');
+        if path.is_empty() || path == "/" {
```
andrei-21: After trimming "/" is not really possible:
```suggestion
        if path.is_empty() {
```

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -226,6 +226,67 @@ impl EntryRepository {
...
+
+        path.char_indices()
+            .filter_map(|(idx, ch)| {
+                if ch == '/' && idx > 0 {
+                    Some(path[..idx].to_string())
+                } else {
+                    None
+                }
+            })
+            .collect()
```
andrei-21: Wouldn't be it mentally simpler to user `rsplit_once` in a loop:
```rust
        while let Some((parent, _)) = path.rsplit_once('/') {
            if parent.is_empty() {
                break;
            }

            paths.push(parent);
            path = parent;
        }
```

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -226,6 +226,67 @@ impl EntryRepository {
...
+        .bind(ancestor_paths)
+        .fetch_one(con)
+        .await
+    }
+
+    fn ancestor_file_paths(path: &str) -> Vec<String> {
+        let path = path.trim_end_matches('/');
+        if path.is_empty() || path == "/" {
+            return Vec::new();
+        }
```
andrei-21, line 284: Again, it shouldn't be really possible, but if a path ends with `/`, then we also should include the path itself without the trailing slash. E.g. we try to write to `/pub/test/`, we should check that `/pub/test` does not exist.

andrei-21: It is still missing, no?

andrei-21 (approved):

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
MCarlomagno: I have added the check before (line 244): 

https://github.com/pubky/pubky-core/pull/424/changes#diff-f9864dbd5cbeb10e3c25c8116e736b987752de0d25a8211e7fef32afbcd9f784R244

PR #431, opened 2026-06-17
docs(homeserver): Improve developer docs and readmes

SeverinAlexB:
- Overhaul the root README to immediately describe what the repo is and point to other documentations.
- Replaced `GETTING_STARTED.md` and `DEV_TESTING_GUIDES.md` with two guides: `INSTALL.md` for operators and `TESTING.md`  for devs.
- Improved other crate readmes

Ive attempted to keep things succinct whilst also ensuring to mention all available options and ways users may trip up. I think i found a good balance but please let me know your thoughts. 

Not included:
- Tutorial for setting up ICANN, reverse proxy for TLS, expose ports
- Other platforms - kept to Debian-based systems for now. How to manage instructions for mac, non-debian and windows can be decided later once we have some feedback on these docs.#
- Links to pubky.org


docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+The [pubky-docker](https://github.com/pubky/pubky-docker) repository provides a Docker Compose setup that runs the homeserver and PostgreSQL together using pre-built images. This is the fastest way to get a homeserver running.
+
+Requires [Docker Engine](https://docs.docker.com/engine/install/).
+
+```bash
+git clone https://github.com/pubky/pubky-docker.git
+cd pubky-docker
+cp .env-sample .env
+docker compose --profile backend up -d postgres homeserver
```
SeverinAlexB: NIT: Might be worth adding `--no-build`.

AI output:

The Docker Compose quick start says this uses pre-built images, but the command omits --no-build. In pubky-docker, the homeserver service has both image and build: ../pubky-core, so on a fresh clone Compose may try to build from a missing sibling ../pubky-core instead of pulling the published image. The upstream pubky-docker README uses docker compose --profile backend up -d --no-build.

pubky-homeserver/README.md
```diff
@@ -1,108 +1,68 @@
...
-      server.admin().listen_socket()
-  );
-  tokio::signal::ctrl_c().await?;
-
-  println!("Shutting down Homeserver");
-  Ok(())
-}
-```
+    let app = HomeserverApp::start_with_persistent_data_dir_path(
+        PathBuf::from("~/.pubky")
```
SeverinAlexB, line 35: AI:

`PathBuf::from("~/.pubky")` does not expand `~` in Rust. Users copying this library example will use a literal relative path ~/.pubky, not their home directory. Use `dirs::home_dir().unwrap().join(".pubky")` or an explicit absolute path.

README.md
```diff
@@ -1,88 +1,87 @@
-<h1 align="center"><a href="https://pubky.org/"><img alt="pubky" src="./.svg/pubky-core-logo.svg" width="200" /></a></h1>
+<h1 align="center"><a href="https://pubky.org/"><img alt="Pubky" src="./.svg/pubky-core-logo.svg" width="200" /></a></h1>
 
 <h3 align="center">
-	An open protocol for per-public-key backends for censorship resistant web applications.
+  Reference Homeserver implementation for Pubky.
```
SeverinAlexB: I am not sure we should call it "Reference". With "Reference" it feels like we want to encourage alternative implementations and I am not sure we should do this already. At the moment, we can just be reckless and dirty with changes. If we have other implementations that use us as reference, this gets more difficult.

So I think we should remove "Reference" here for at least a while. We can add it back when we are more stable IMO.

README.md
```diff
@@ -1,88 +1,87 @@
...
+  <a href="https://t.me/pubkycore"><img src="https://img.shields.io/badge/Chat-Telegram-violet" alt="Telegram Chat Group" /></a>
+  <a href="https://deepwiki.com/pubky/pubky-core"><img src="https://deepwiki.com/badge.svg" alt="Ask DeepWiki" /></a>
 </div>
 
-[![GitHub Release](https://img.shields.io/github/v/release/pubky/pkdns)](https://github.com/pubky/pubky-core/releases/latest/)
-[![Crates.io Version](https://img.shields.io/crates/v/pubky)](https://crates.io/crates/pubky)
-[![Telegram Chat Group](https://img.shields.io/badge/Chat-Telegram-violet)](https://t.me/pubkycore)
-[![GitHub License](https://img.shields.io/github/license/pubky/pubky-core)](https://github.com/pubky/pubky-core/blob/main/LICENSE)
-[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/pubky/pubky-core)
+## What Is This Repository?
```
SeverinAlexB: Unnecessary title?

```suggestion
```

README.md
```diff
@@ -1,88 +1,87 @@
...
+A Pubky Homeserver stores and serves user data. It exposes HTTP APIs for authenticated writes and public reads, and publishes DNS records so other clients can discover where a user's data lives.
 
-## Features
-- Public key based authentication.
-- Public key based 3rd party authorization.
-- Key-value store through PUT/GET/DELETE HTTP API + pagination.
+- Public-key based signup and signin
+- Third-party authorization through Pubky auth flows
+- File storage via HTTP `PUT`, `GET`, `DELETE`, and listing APIs (WebDAV-like)
+- PKDNS/Pkarr publishing for homeserver discovery
```
SeverinAlexB: Always write PKARR with capital letters. Please check the rest of the PR on this too.

```suggestion
- PKDNS/PKARR publishing for homeserver discovery
```

README.md
```diff
@@ -1,88 +1,87 @@
...
-```
-Then in a different terminal window:
-```bash
-npm test
-```
+| I want to... | Guide |
+| --- | --- |
+| Run a homeserver | [Install and Run Pubky Homeserver](./docs/INSTALL.md) |
+| Develop locally against a testnet | [Local Development](./docs/LOCAL_DEVELOPMENT.md) |
+| Run tests and CI | [Testing](./docs/TESTING.md) |
```
SeverinAlexB: This two columns feel redundant. `Run a homeserver` is very simillar to `Install and run Pubky Homeserver`. Should we turn this into a simple list?


```suggestion
- [Install and run a homeserver](./docs/INSTALL.md) |
- [Develop locally against a testnet](./docs/LOCAL_DEVELOPMENT.md) |
- [Run tests and CI](./docs/TESTING.md) |
```

README.md
```diff
@@ -1,88 +1,87 @@
...
+npm install @synonymdev/pubky
 ```
 
-Additional optional arguments can be used to run it in the background, but the most important is `--network=host`, which allows the container to access the network and provides an admin endpoint accessible from the host machine. Please refer to the Docker documentation for more detailed options.
+See the [`pubky-sdk` README](./pubky-sdk) for API details or browse the [examples](./examples).
 
 ## Links
 
+- [Pubky website](https://pubky.org/)
+- [Documentation](https://docs.pubky.org/)
```
SeverinAlexB: docs.pubky.org is now simply pubky.org.

README.md
```diff
@@ -1,88 +1,87 @@
...
+- [Pkarr](https://pkarr.org/)
```
SeverinAlexB: This is Ars domain. Please link to our Git repo https://github.com/pubky/pkarr

README.md
```diff
@@ -1,88 +1,87 @@
...
 - [Contributors Guide](./CONTRIBUTORS.md)
```
SeverinAlexB: I would not draw too much attention on the Contributors guide. External contributors are more pain than good in my experience. Can we remove this line?

README.md
```diff
@@ -1,88 +1,87 @@
...
+- [Release Process](./RELEASING.md)
+
+---
+
+May the power ⚡ be with you. Powered by [pkarr](https://github.com/pubky/pkarr).
```
SeverinAlexB: PKARR mentions are getting a bit excessive here IMO

SeverinAlexB (commented):
I only reviewed the main README so far. Looking at the overall structure, it feels incomplete/thin.

Here are some AI recommendations edited by me on how to extend the readme with:


- **No quick start command path.** The README points to guides, but a first-time reader has no immediate “try it in 2 minutes” path. Even 3 lines linking to Docker Compose or local testnet would help.
- **The repo scope is undersold.** The README does not say what pubky-core is about. It is an attempt create a new internet, free from wallet gardens (see Johns vision). At the moment, the README does not say anything about it. It feels technically dry.
- **No “who is this for?” split.** There are at least three audiences: app developers using the SDK, contributors developing locally, and operators running a homeserver. The README would benefit from making those paths explicit.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
+# Install and Run Pubky Homeserver
+
+This guide is for running a standalone Pubky homeserver. For local app development against a local testnet, see [Local Development](./LOCAL_DEVELOPMENT.md). For contributor test databases and CI setup, see [Testing](./TESTING.md).
```
SeverinAlexB: Reading this section feels bumpy. Can we do something like this?

```suggestion
This guide is for running a standalone Pubky homeserver.

> - For local app development against a local testnet, see [Local Development](./LOCAL_DEVELOPMENT.md). 
> - For contributor test databases and CI setup, see [Testing](./TESTING.md).
```

I am not even sure if my suggestion is the way to go. To put my "feeling bumpy" into words:

I ready 1 sentence what this file is about:
> This guide is for running a standalone Pubky homeserver.

And then I read two sentences that are unrelated:
>  For local app development against a local testnet, see [Local Development](./LOCAL_DEVELOPMENT.md). For contributor test databases and CI setup, see [Testing](./TESTING.md).

That confuses me. Maybe we should just remove these two sentences?

Happy to hear what you think.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+  - [Configure the Homeserver with PostgreSQL](#configure-the-homeserver-with-postgresql)
+  - [Run](#run)
+- [Configuration](#configuration)
+- [Production Notes](#production-notes)
+- [Troubleshooting](#troubleshooting)
+
+
+## Quick Start with Docker Compose
+
+The [pubky-docker](https://github.com/pubky/pubky-docker) repository provides a Docker Compose setup that runs the homeserver and PostgreSQL together using pre-built images. This is the fastest way to get a homeserver running.
```
SeverinAlexB: pubky-docker is not intended for regular homeserver running.

Image

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+### Release Binary
+
+Download the latest non-prerelease archive from the [Pubky Core releases page](https://github.com/pubky/pubky-core/releases). Choose the archive for your operating system and CPU architecture, extract it, and place `pubky-homeserver` somewhere on your `PATH`.
+
+Requires `curl`. On Ubuntu/Debian: `apt install curl`.
+
+```bash
+curl -LO https://github.com/pubky/pubky-core/releases/download/vx.x.x/pubky-core-vx.x.x-linux-amd64.tar.gz
+tar -xf pubky-core-vx.x.x-linux-amd64.tar.gz
+cp pubky-core-vx.x.x-linux-amd64/pubky-homeserver /usr/local/bin
```
SeverinAlexB: NIT: Can we do the copy in two steps? 1. cd into dir 2. cp. Not critical but feels a bit easier to comprehend.


```suggestion
curl -LO https://github.com/pubky/pubky-core/releases/download/vx.x.x/pubky-core-vx.x.x-linux-amd64.tar.gz
tar -xf pubky-core-vx.x.x-linux-amd64.tar.gz
cd pubky-core-vx.x.x-linux-amd64
cp pubky-homeserver /usr/local/bin
```

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
```
SeverinAlexB, line 51: Can we add a quick verifying section here where the user runs `pubky-homeserver --version` after the binary is copied? This would ensure that the users downloaded the correct release binaries for his architecture. This is a common mistake users.

Something like:

Run `pubky-homeserver --version`. If everything is installed correctly it should output the binary version.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+Clone the repository:
+
+```bash
+git clone https://github.com/pubky/pubky-core.git
+cd pubky-core
+git checkout vx.x.x   # Pick a version
+```
+
+#### With Cargo
```
SeverinAlexB: The "With Cargo" and "With Docker" titles feel weird. Shouldn't they be called "Build binaries" and "Builder Docker image" or so? Reading "With Cargo" makes me ask "Yes, I am building with Cargo. It's a Rust app. What is this about?".

Maybe it's even worth splitting this guide in two: One for Binary build and one for the docker image build. Hard to say.

Image

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+Build and place the binary on your `PATH`:
+
+```bash
+cargo build --release -p pubky-homeserver
+cp ./target/release/pubky-homeserver /usr/local/bin
+```
+
+#### With Docker
+
+Build the homeserver image using the [Dockerfile](../Dockerfile):
```
SeverinAlexB: Do you think this sentence is necessary or can it be dropped for the sake of being consice?

```suggestion
```

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+docker build --build-arg BUILD_TARGET=homeserver -t pubky-homeserver .
+```
+
+You can run the homeserver directly in Docker:
+
+```bash
+docker run -it --network=host pubky-homeserver
+```
+
+Use `--network=host` so the container can reach PostgreSQL on the host and expose its endpoints.
```
SeverinAlexB: You run the container here before you setup Postgres. If a users follows from top to bottom, he will fail here. You didn't do this the "with cargo" binary build before.

Image

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+See the Docker documentation for volume mounts and other options.
+
+Or copy the binary out of the image and place it on your `PATH`:
+
+```bash
+docker create --name tmp-hs pubky-homeserver
+docker cp tmp-hs:/usr/local/bin/homeserver /usr/local/bin/pubky-homeserver
+docker rm tmp-hs
+```
```
SeverinAlexB: What is this? Sounds unnecessary? The image might not even have the same compute architecture than the host.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+```bash
+docker run -it --network=host pubky-homeserver
+```
+
+Use `--network=host` so the container can reach PostgreSQL on the host and expose its endpoints.
+
+> **macOS note:** Docker Desktop for macOS does not support `--network=host`. Use `-p 6286:6286 -p 6287:6287 -p 6288:6288 -p 6289:6289` instead to map the ports manually.
+
+See the Docker documentation for volume mounts and other options.
```
SeverinAlexB: To run the docker homeserver correctly in production, one needs a volume to persist the data no? So it should be described here, shouldn't it?

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+Build the homeserver image using the [Dockerfile](../Dockerfile):
+
+```bash
+docker build --build-arg BUILD_TARGET=homeserver -t pubky-homeserver .
+```
+
+You can run the homeserver directly in Docker:
+
+```bash
+docker run -it --network=host pubky-homeserver
```
SeverinAlexB: How do I configure docker? You describe creating the config.toml and so on with the binary but it's missing with docker completely, isn't?

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+docker rm tmp-hs
+```
+
+## Initialise the Data Directory
+
+Create the data directory, default `config.toml`, and server keypair without starting the server or connecting to PostgreSQL:
+
+```bash
+pubky-homeserver init
+```
```
SeverinAlexB, line 121:
```
pubky-homeserver init
error: unexpected argument 'init' found

Usage: pubky-homeserver [OPTIONS]

For more information, try '--help'.
```

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+This creates `~/.pubky/` with a sample config and a fresh server keypair. To use a different path:
+
+```bash
+pubky-homeserver --data-dir /path/to/pubky-data init
+```
+
+## Set Up PostgreSQL
+
+The homeserver requires a running PostgreSQL instance with an empty database. It runs migrations automatically on startup. The default connection string is `postgres://localhost:5432/pubky_homeserver`.
```
SeverinAlexB: For the sake of being concise (the user wants to get this running asap), should we remove these two (unnecessary) sentences?
```suggestion
The homeserver requires a running PostgreSQL instance with an empty database.
```

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Docker
+
+Requires [Docker Engine](https://docs.docker.com/engine/install/ubuntu/).
```
SeverinAlexB, line 143: Mentioning this down here is a bit late. Could be moved up where one actually starts using docker.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+  -p 127.0.0.1:5432:5432 \
+  -v postgres-data:/var/lib/postgresql \
+  -d postgres:18
+```
+
+Verify it's running and the database exists:
+
+```bash
+docker exec pubky-postgres psql -U postgres -c '\l pubky_homeserver'
+```
```
SeverinAlexB: Is this verification step necessary? This docker image guarantees to create `POSTGRES_DB`. So it feels like it's not needed?

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+```bash
+pubky-homeserver
+```
+
+From source:
+
+```bash
+cargo run -p pubky-homeserver
+```
```
SeverinAlexB: We already built from source. To stay concise, I suggest we remove this

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+| Metrics API | `http://127.0.0.1:6289` |
+
+Standalone homeservers require signup tokens by default. Generate one through the admin API:
+
+```bash
+curl -X GET "http://127.0.0.1:6288/generate_signup_token" \
+  -H "X-Admin-Password: admin"
+```
+
+You can also open signup entirely for a private or temporary deployment by setting `signup_mode = "open"` in `config.toml` (see [Configuration](#configuration)).
```
SeverinAlexB: Not really necessary here wdyt? This is more of a side feature than something we should mention here. Default should be signup tokens.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+| `general.signup_mode` | Signup policy: `token_required` or `open`. |
+| `drive.icann_listen_socket` | Regular HTTP API listen address. |
+| `drive.pubky_listen_socket` | Pubky TLS API listen address. |
+| `storage.type` | Storage backend: `file_system`, `google_bucket`, or `in_memory`. |
+| `admin.enabled` | Enables the admin API. |
+| `admin.listen_socket` | Admin API listen address. |
+| `metrics.enabled` | Enables Prometheus metrics. |
+| `metrics.listen_socket` | Metrics API listen address. |
+| `pkdns.public_ip` | Public IP advertised for Pubky discovery. |
+| `pkdns.icann_domain` | Domain used for regular browser HTTP access. |
```
SeverinAlexB: - Metrics are very much a niche feature for devops. Wouldn't feature it here as "important settings".
- signup_mode is not really important IMO
- `admin.enabled` is not so important to be featured here IMO.
- `pkdns` might be worth having a full section on. But I think you are already on this?

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+Review the full documented sample at [`pubky-homeserver/config.sample.toml`](../pubky-homeserver/config.sample.toml).
+
+## Production Notes
+
+Before using a homeserver in production:
+
+- Use a persistent PostgreSQL instance with password authentication and back it up. Do not use trust auth in production.
```
SeverinAlexB: Why not?

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+- Back up the homeserver `secret` file and any filesystem or bucket storage.
+- Do not expose the admin or metrics APIs to the public internet.
+- Change the default admin password in `[admin].admin_password`.
+- Configure `pkdns.public_ip`, `pkdns.icann_domain`, and public ports for your deployment.
+- Put the regular HTTP API behind a reverse proxy if you need browser-compatible HTTPS.
```
SeverinAlexB: Worth having this better described, what do you think? We have two sockets (pubky tls + regular http) so it's not obvious how to do it, is it?

pubky-homeserver/README.md
86667:
```rust
impl PersistentDataDir {
    /// Creates a new data directory.
    /// `path` will be expanded to the home directory if it starts with "~".
    pub fn new(path: PathBuf) -> Self {
        Self {
            expanded_path: Self::expand_home_dir(path),
        }
    }

    /// Expands the data directory to the home directory if it starts with "~".
    /// Return the full path to the data directory.
    fn expand_home_dir(path: PathBuf) -> PathBuf {
        let path = match path.to_str() {
            Some(path) => path,
            None => {
                // Path not valid utf-8 so we can't expand it.
                return path;
            }
        };

        if path.starts_with("~/") {
            if let Some(home) = dirs::home_dir() {
                let without_home = path.strip_prefix("~/").expect("Invalid ~ prefix");
                let joined = home.join(without_home);
                return joined;
            }
        }
        PathBuf::from(path)
    }
...
```

This isnt my code and the example data dir hasnt changed.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+- Use persistent filesystem storage or a configured bucket backend, not in-memory storage.
+- Monitor logs, PostgreSQL health, disk usage, and storage backend errors.
```
SeverinAlexB:
```suggestion
```

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+### `database "pubky_homeserver" does not exist`
+
+Create the database:
+
+```bash
+createdb -h <HOST> -U <USER> pubky_homeserver
+```
+
+Or update `[general].database_url` in `~/.pubky/config.toml` to point at an existing database.
```
SeverinAlexB, line 288: Is this for the docker or fixed install case? I assume fixed install. It's not obvious when reading though.

docs/INSTALL.md
```diff
@@ -0,0 +1,303 @@
...
+
+### PostgreSQL Connection Refused
+
+Make sure PostgreSQL is running and listening on the host and port in `general.database_url`.
+
+For the Docker examples above, check the container:
+
+```bash
+docker ps --filter name=pubky-postgres
+```
```
SeverinAlexB, line 322: Does this command answer if PG is running and listening and the correct host/port?

SeverinAlexB (commented):
This is the INSTALL.md review

docs/INSTALL.md
86667: I dont think it hurts to explain what were doing before the code block, its what we do in the rest of the file for other commands.

docs/INSTALL.md
86667: changed to `Build a binary with Cargo` and `Build a Docker image`

docs/INSTALL.md
86667: added that same warning here

docs/INSTALL.md
86667: Made into a `Looking for something else?` note

docs/INSTALL.md
86667: Replaced with a `docker run --rm pubky-homeserver --version`

docs/INSTALL.md
86667: I was hoping to keep things to a single run method to make later docs simpler - ie just one `pubky-homeserver` command. This wont work though, youre right.  
Removed this option completely.

docs/INSTALL.md
86667: Added note about requiring v0.10 and what to do if on <0.10

docs/INSTALL.md
86667: This is the first mention of docker in the Postgres section. Copied  this line to the Docker build instructions.

docs/INSTALL.md
86667: removed all apart from admin.

docs/INSTALL.md
86667: Removed. This is a relic from a previous version in which I attempted to simplify things by not setting a user and password on the postgres server. Decided against that change - all docs set user and pass by default now so this no longer needs saying.

docs/INSTALL.md
86667: Improved this note and also will link to full pubky.org guides when they exist

docs/INSTALL.md
86667: Added both install method commands to each troubleshooting item. Im getting close to creating a separate doc for postgres install, or seperate docs for each of the install methods. Lets see how it grows

docs/INSTALL.md
86667: Improved this section generally

86667:
@SeverinAlexB The thinking was that intro information on the system as a whole should live only in pubky.org and linked to in the repo, which I hadnt done yet. I do however see the value in there being a short intro here aswell for those who have stumbled upon the repo without context, so i've added a paragraph with pointers to pubky.org.

Im not sure about a quick start guide on the root repo. Quick start for what? development, running a HS, building an app? IMO better to link out to docs/guides immediately. 

I liked the idea of "who is this for" section so used that as a base for linking to the new guides on this branch. 

I also stripped down this core readme - its now just an intro with links to pubky.org and repo docs.

docs/INSTALL.md
SeverinAlexB: Did you add this to INSTALL.md? This is not what I meant and feels unnecessary.

CleanShot 2026-07-07 at 08 30 02@2x

What I meant is that you recommend to run the homeserver with pubky-docker but pubky-docker is explicitly not made for this. I think we should remove pubky-docker mentions in here.

pubky-homeserver/README.md
SeverinAlexB: You are correct.

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+cd pubky-core
+git checkout v${PUBKY_CORE_VERSION}
+```
+
+#### Build a binary with Cargo
+
+Make sure you have the Rust toolchain installed and working.
+
+<details>
+<summary>Install Rust</summary>
```
SeverinAlexB: NIT

```suggestion
<summary>How to Install the Rust Toolchain</summary>
```

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+
+```bash
+docker build --build-arg BUILD_TARGET=homeserver -t pubky-homeserver .
+```
+
+Verify the image built correctly:
+
+```bash
+docker run --rm pubky-homeserver homeserver --version
+```
```
SeverinAlexB, line 113: Interesting, maybe out of scope of this PR but can't you make the homeserver the default executable? Reading this now, `pubky-homeserver homeserver` feels duplicate.

Edit: https://github.com/pubky/pubky-core/pull/483

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+| `admin.enabled` | Enables the admin API. |
+| `admin.listen_socket` | Admin API listen address. |
+
+Review the full documented sample at [`pubky-homeserver/config.sample.toml`](../pubky-homeserver/config.sample.toml).
+
+## Production Notes
+
+Before using a homeserver in production:
+
+- Use a persistent PostgreSQL instance with password authentication and back it up.
```
SeverinAlexB: Hmmm, not sure we need to say this. Additional, recommending password auth is wrong as there are many good PG auth options and passwords is not even the strongest. I would fully remove this line.

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+- Back up the homeserver `secret` file and any filesystem or bucket storage.
```
SeverinAlexB: That's a good one. Maybe we can do something like this? Maybe same same but different but this feels better

```suggestion
- Back up the homeserver
  - The keypair `.pubky/secret`
  - Any user data dending on the configured option. For example, by default files are saved in `.pubky/data/files`
  - Postgres
```

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+- Do not expose the admin or metrics APIs to the public internet.
+- Change the default admin password in `[admin].admin_password`.
+- Configure `pkdns.public_ip`, `pkdns.icann_domain`, and public ports for your deployment.
```
SeverinAlexB, line 269: I think we still need a homeserver pkdns guide, no?

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+- Use persistent filesystem storage or a configured bucket backend, not in-memory storage.
```
SeverinAlexB: I think we can remove this line. By default, the filesystem is configured and they don't need to think about this. Using buckets or in-memory storage is advanced anyway.

```suggestion
```

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+- The homeserver exposes two sockets: a **Pubky TLS** socket (`pubky_listen_socket`, default port 6287) and a regular **HTTP** socket (`icann_listen_socket`, default port 6286). Pubky TLS uses PKARR-based TLS and does not need a certificate so can be exposed directly. The HTTP socket serves browsers and should be put behind a reverse proxy if you need standard HTTPS with a domain certificate.
```
SeverinAlexB: Good you mention this but this should probably it's own pkdns/expose to the internet guide, don't you think?

docs/INSTALL.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Troubleshooting
```
SeverinAlexB, line 272: I am not sure we should provide so much PG troubleshooting. It's feels out of scope. At the same time, PG connections can fail for so many reasons, it's hard to cover them all.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+- Node.js `20` or newer for JS/WASM bindings.
+- `wasm-pack` when working on the JavaScript SDK bindings.
+
+Useful commands:
+
+```bash
+cargo check --workspace --all-features
+cargo fmt --check
+cargo clippy --workspace --all-features --exclude pubky-wasm -- -D warnings
+```
```
SeverinAlexB: Not sure this is needed here. Any contributor will figure things out as soon as they open a PR and CI fails.
At the same time, this should not be a contributors guide.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
+# Local Development
+
+This guide covers setting up a local testnet for developing and testing against Pubky. Use it when you're modifying crates in this repo or building an app that talks to a local homeserver.
```
SeverinAlexB: This should be a guide on how to build on Pubky, not build Pubky itself. Merging these two makes the doc very confused and overwhelming for new devs IMO.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+  - [Run the Testnet](#run-the-testnet)
+- [Run Examples](#run-examples) (keygen, signup, write, read)
+- [Custom Homeserver Config](#custom-homeserver-config)
+- [Troubleshooting](#troubleshooting)
+
+## Toolchain
+
+For building from source and running the local testnet:
+
+- Rust `1.89` or newer - [rustup](https://rustup.rs/).
```
SeverinAlexB: I would avoid mentioning specific versions as they can change.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+- PostgreSQL for tests and the local testnet (see [Set Up PostgreSQL](#set-up-postgresql)).
+- Node.js `20` or newer for JS/WASM bindings.
+- `wasm-pack` when working on the JavaScript SDK bindings.
```
SeverinAlexB: No js needed. They can use the libs from npm if they want to develop with js.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+
+```bash
+docker run --name pubky-postgres \
+  -e POSTGRES_USER=postgres \
+  -e POSTGRES_PASSWORD=postgres \
+  -p 127.0.0.1:5432:5432 \
+  -d postgres:18
+```
+
+For native PostgreSQL install or to use an existing instance see [INSTALL.md - Set Up PostgreSQL](./INSTALL.md#set-up-postgresql). You can skip the database creation step if you're only using the testnet.
```
SeverinAlexB: I would not mention the native PG. Docker is the easiest and they should stick with it. Too many options will just confuse devs.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+
+## Local Testnet
+
+The local testnet starts a full Pubky network from source: homeserver, local DHT, and relays. The testnet is currently ephemeral: all data is cleaned up when the process exits. Persistent testnet support is planned.
+
+### Set Up PostgreSQL
+
+The testnet requires a running PostgreSQL instance with a user that can create databases. The testnet creates temporary `pubky_test_*` databases automatically and cleans them up on exit.
```
SeverinAlexB: Persistent testnet is almost a requirement for this guide it feels like.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+| Pkarr relay | `http://127.0.0.1:15411` |
+| HTTP relay | `http://127.0.0.1:15412` |
+| Homeserver HTTP API | `http://127.0.0.1:6286` |
+| Homeserver Pubky TLS API | `127.0.0.1:6287` |
+| Homeserver admin API | `http://127.0.0.1:6288` |
+
+The testnet homeserver uses this public key:
+
+```text
+pubky8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo
```
SeverinAlexB: We need to remove the `pubky` prefix at some point so I suggest to change it to a plain pubkey

```suggestion
8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo
```

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+## Run Examples
+
+> **Prerequisites:** [Rust toolchain](https://rustup.rs/) and the [cloned source repository](./INSTALL.md#build-from-source).
+
+With the testnet running, use the examples in a separate terminal. Rust examples are in [`examples/rust`](../examples/rust).
```
SeverinAlexB: Examples should not be a replacement for a CLI. They are very bad DX anyway. I am about to change the examples heavily as well. No idea how far Vlad is with the homeserver CLI, if the operations fit into the homeserver cli even or if we need something different.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+### 4. Write data
+
+```bash
+cargo run -p pubky-core-examples --bin storage -- write --testnet \
+  /tmp/pubky-recovery.json \
+  /pub/example/hello.txt \
+  --content 'Hello from pubky!'
+```
+
+This signs in, writes the file, reads it back, and deletes it.
```
SeverinAlexB: Do we need a full testnet explorer webapp? Using the storage example is really bad as it will not just write but also read and delete after. Again, this is an example, not meant to be used as a CLI.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+npm i
+npm run build
+```
+
+To run the JS binding tests, start the testnet first (see [Run the Testnet](#run-the-testnet)), then in a separate terminal:
+
+```bash
+cd pubky-sdk/bindings/js/pkg
+npm test
+```
```
SeverinAlexB: Remove. Not important for a dev developing on Pubky.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,179 @@
...
+## Run Examples
+
+> **Prerequisites:** [Rust toolchain](https://rustup.rs/) and the [cloned source repository](./INSTALL.md#build-from-source).
+
+With the testnet running, use the examples in a separate terminal. Rust examples are in [`examples/rust`](../examples/rust).
+
+### 1. Generate a keypair
+
+```bash
+cargo run -p pubky-core-examples --bin keygen -- --output /tmp/pubky-recovery.json
```
SeverinAlexB: How much of this is AI generated? Keygen does not create a json file.

docs/TESTING.md
```diff
@@ -0,0 +1,125 @@
+# Testing
+
+This guide is for contributors running Rust tests, integration tests, or CI jobs. 
+
+> **Looking for something else?** 
+See [Local Development](./LOCAL_DEVELOPMENT.md) for running a local development testnet and [Install and Run Pubky Homeserver](./INSTALL.md) for deploying a standalone homeserver.
```
SeverinAlexB: Reading this now for the 3rd time, I don't know if this section is needed.

README.md
```diff
@@ -1,88 +1,56 @@
...
 
-### JavaScript
-If you prefer to use JavaScript in NodeJs/Browser or any runtime with Wasm support, you can either install from npm [`@synonymdev/pubky`](https://www.npmjs.com/package/@synonymdev/pubky)
-or build the bindings yourself:
-```bash
-cd pubky-sdk/bindings/js/pkg
-npm i
-npm run build
-```
+[Pubky](https://pubky.org) is an open protocol for building censorship-resistant applications where users own their identity, data, and connections. No platform lock-in, no algorithmic feeds you don't control, no losing everything when a service shuts down. Your keys are your identity, and you choose where your data lives.
```
SeverinAlexB: Can we get more space betweet the badges and this text? It's too sticky currently.

Image

SeverinAlexB: Do I control the algorithmic feed in Pubky.app? 

```suggestion
[Pubky](https://pubky.org) is an open protocol for building censorship-resistant applications where users own their identity, data, and connections. No platform lock-in, no losing everything when a service shuts down. Your keys are your identity, and you choose where your data lives.
```

README.md
```diff
@@ -1,88 +1,56 @@
...
-An alternative way to start tinkering with Pubky is to build an isolated container and run it locally. Here is an 
-example command how to build an image:
+## Who Is This For?
 
-```bash
-docker build --build-arg TARGETARCH=x86_64 -t pubky:core .
-```
+- **App developers**: Use the [SDK](https://pubky.org/explore/pubkycore/sdk/) to build apps that read and write user data on homeservers. Start with [Local Development](./docs/LOCAL_DEVELOPMENT.md) and the [examples](./examples).
+- **Operators**: [Install and run a homeserver](./docs/INSTALL.md) to host user data.
+- **Contributors** [Set up a local testnet](./docs/LOCAL_DEVELOPMENT.md) and [run the tests](./docs/TESTING.md) to develop on pubky-core itself.
```
SeverinAlexB: Please again, do not mention contributors. We really don't want to encourage them. They are more pain than good.

README.md
```diff
@@ -1,88 +1,56 @@
...
 
-A command for running it in an isolated environment with log output:
+## What Is a Homeserver?
 
-```bash
-docker run -it pubky:core
-```
+A Pubky homeserver stores and serves user data for Pubky app. Users choose which homeserver holds their data, and can move to another at any time. The homeserver exposes HTTP APIs for authenticated writes and public reads, and publishes [PKARR](https://github.com/pubky/pkarr) records so other clients can discover where a user's data lives.
```
SeverinAlexB: "Pubky app": Do you mean pubky.app or any Pubky application?

docs/INSTALL.md
SeverinAlexB: Maybe this is not good either. Maybe we should just link to the install page and avoid lengthy explanations in this regards. The docs are missing clear focus IMO (partially an initial mistake by me) so removing the section and just linking to the rustup page might be the best move.

docs/LOCAL_DEVELOPMENT.md
86667: removed completely

docs/TESTING.md
86667: removed

pubky-testnet/README.md
```diff
@@ -9,36 +9,20 @@ Two testnet types are provided:
 | [`EphemeralTestnet`] | Random | In-memory | Automated tests (`#[tokio::test]`) - parallel-safe, no port conflicts |
 | [`StaticTestnet`] | Fixed, well-known | In-memory or persistent | Interactive / CLI use - browser tests, mobile apps, manual debugging |
```
86667, line 11: This file i improved when i merged the recent `StaticTestnet` persistent data-dir changes. I made it match the guides by adding a ToC and troubleshooting section. Here I remove those in favour of keeping it as a developer-aimed crate explainer.

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+### Release Binary
+
+Download and extract the archive (requires `curl`; on Ubuntu/Debian: `apt install curl`):
+
+```bash
+curl -LO https://github.com/pubky/pubky-core/releases/download/v${PUBKY_CORE_VERSION}/pubky-core-v${PUBKY_CORE_VERSION}-linux-amd64.tar.gz
+tar -xf pubky-core-v${PUBKY_CORE_VERSION}-linux-amd64.tar.gz
+```
```
SeverinAlexB, line 38: The download url also defines the cpu architecture like `amd64`. Should this also be extracted into a variable similiar to the `PUBKY_CORE_VERSION`?

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+- [Production Notes](#production-notes)
+- [Troubleshooting](#troubleshooting)
+
+
+## Install the Homeserver
+
+Pick a version from the [Pubky Core releases page](https://github.com/pubky/pubky-core/releases). The commands below use this variable, so set it first:
+
+```bash
+PUBKY_CORE_VERSION=0.9.3
```
SeverinAlexB: AI: The guide selects version 0.9.3, then instructs users to run pubky-homeserver init. That command only exists in v0.10+, so the primary installation path fails.

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+
+With Docker:
+
+```bash
+docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserver
+```
+
+Use `--network=host` so the container can reach PostgreSQL on the host and expose its endpoints. The volume mount shares the data directory (config and keypair) with the container.
+
+> **macOS note:** Docker Desktop for macOS does not support `--network=host`. Use `-p 6286:6286 -p 6287:6287 -p 6288:6288 -p 6289:6289` instead to map the ports manually.
```
SeverinAlexB: AI review, no idea if valid. Posting here because I am not sure if you ever run it on a Mac.

docs/INSTALL.md:223-228: The macOS Docker alternative is not functional. Published ports cannot reach services bound to 127.0.0.1 inside the container, and localhost in database_url points to the container rather than the host. The guide needs either Docker Desktop host networking instructions or a complete bridge setup using host.docker.internal and 0.0.0.0 container listeners.

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+```bash
+pubky-homeserver --version
+```
+
+### Build From Source
+
+Install build dependencies (Ubuntu/Debian):
+
+```bash
+apt update && apt install -y build-essential pkg-config libssl-dev git curl
```
SeverinAlexB: Should we add `sudo` to the commands?

AI review:
docs/INSTALL.md:32-42,53-56,158-173: Ubuntu/Debian commands require root access but omit sudo, including apt, copying into /usr/local/bin, pg_ctlcluster, and su - postgres. The latter commonly fails on Ubuntu because the system account is locked.

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
+# Install and Run Pubky Homeserver
+
+How to set up and operate a Pubky homeserver.
```
SeverinAlexB: I think we need to say somewhere that this guide is intended for Ubuntu 26. Otherwise people will be confused if they want to install it on Windows. What do you think?

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+> **macOS note:** Docker Desktop for macOS does not support `--network=host`. Use `-p 6286:6286 -p 6287:6287 -p 6288:6288 -p 6289:6289` instead to map the ports manually.
+
+The default endpoints are:
+
+| Endpoint | Default |
+| --- | --- |
+| Public HTTP API | `http://127.0.0.1:6286` |
+| Pubky TLS API | `127.0.0.1:6287` |
+| Admin API | `http://127.0.0.1:6288` |
+| Metrics API | `http://127.0.0.1:6289` |
```
SeverinAlexB: AI review: The metrics endpoint is listed as a default endpoint, but metrics are disabled by default.

Should we remove metrics from this guide? It's not really that important.

docs/TESTING.md
```diff
@@ -0,0 +1,120 @@
...
+For many tests, share one Docker PostgreSQL instance with `DockerPostgres::shared()`. Each testnet still creates its own ephemeral database, so test data remains isolated:
+
+```rust
+use pubky_testnet::docker_postgres::DockerPostgres;
+use pubky_testnet::EphemeralTestnet;
+
+#[tokio::test]
+async fn test_one() {
+    let pg = DockerPostgres::shared().await;
+    let testnet = EphemeralTestnet::builder()
```
SeverinAlexB, line 81: I can't really categorize this AI review but posting it here because it's flagged as HIGH. Feel free to close instantly if it is a none issue:

docs/TESTING.md:72-88 and pubky-testnet/README.md:91-124: The shared PostgreSQL examples claim each testnet gets an isolated database. DockerPostgres::connection_string() returns a URL without ?pubky-test=true, so every testnet connects directly to the same postgres database. Concurrent tests can conflict. The connection string or examples must add the test parameter.

pubky-testnet/README.md
```diff
@@ -139,6 +123,36 @@ async fn test_two() {
...
+        "postgres://postgres:postgres@localhost:5432/postgres?pubky-test=true"
+    ).unwrap();
+
+    let testnet = EphemeralTestnet::builder()
+        .postgres(connection_string)
+        .build()
+        .await
+        .unwrap();
+}
+```
```
SeverinAlexB, line 153: This examples needs a `[pubky_testnet::test]` otherwise it will not drop the DB.

AI review:

High - pubky-testnet/README.md:137-154: The programmatic example says the ephemeral database is cleaned up when the testnet is dropped. Dropping only registers it for cleanup; deletion requires #[pubky_testnet::test] or an explicit drop_test_databases().await. As written, repeated runs leak databases.

pubky-homeserver/README.md
```diff
@@ -1,116 +1,58 @@
...
-const response = await client.fetch(url, {
-  method: "GET",
-  headers: {
-    "X-Admin-Password": "admin", // use your admin password, defaults to testnet password.
-  },
-});
-const signupToken = await response.text();
+let config = ConfigToml::default_test_config();
+let mock_dir = MockDataDir::new(config, None).unwrap();
+let app = HomeserverApp::start_with_mock_data_dir(mock_dir).await.unwrap();
```
SeverinAlexB, line 62: AI review:

pubky-homeserver/README.md:50-57: The testing example uses MockDataDir, ConfigToml::default_test_config(), and start_with_mock_data_dir, all gated behind the testing feature. The documented dependency does not enable that feature, so consumers get unresolved imports/methods. Document a dev-dependency with features = ["testing"].

README.md
```diff
@@ -1,88 +1,56 @@
...
-A command for running it in an isolated environment with log output:
+## What Is a Homeserver?
 
-```bash
-docker run -it pubky:core
-```
+A Pubky homeserver stores and serves user data. Users choose which homeserver holds their data, and can move to another at any time. The homeserver exposes HTTP APIs for authenticated writes and public reads, and publishes [PKARR](https://github.com/pubky/pkarr) records so other clients can discover where a user's data lives.
 
-Additional optional arguments can be used to run it in the background, but the most important is `--network=host`, which allows the container to access the network and provides an admin endpoint accessible from the host machine. Please refer to the Docker documentation for more detailed options.
+- Public-key based signup, signin and third-party app authorization
```
SeverinAlexB:
```suggestion
- Public-key-based signup, sign-in, and third-party app authorization.
```

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+| `admin.enabled` | Enables the admin API. |
+| `admin.listen_socket` | Admin API listen address. |
+
+Review the full documented sample at [`pubky-homeserver/config.sample.toml`](../pubky-homeserver/config.sample.toml).
+
+## Production Notes
+
+Before using a homeserver in production:
+
+- Back up homeserver's state:
```
SeverinAlexB:
```suggestion
- Back up the homeserver's state:
```

docs/INSTALL.md
```diff
@@ -0,0 +1,330 @@
...
+  - The keypair `.pubky/secret`
+  - Any user data depending on the configured option. For example, by default files are saved in `.pubky/data/files`
+  - Postgres
+- Do not expose the admin or metrics APIs to the public internet.
+- Change the default admin password in `[admin].admin_password`.
+- Configure `pkdns.public_ip`, `pkdns.icann_domain`, and public ports for your deployment.
+- The homeserver exposes two sockets: a **Pubky TLS** socket (`pubky_listen_socket`, default port 6287) and a regular **HTTP** socket (`icann_listen_socket`, default port 6286). Pubky TLS uses PKARR-based TLS and does not need a certificate so can be exposed directly. The HTTP socket serves browsers and should be put behind a reverse proxy if you need standard HTTPS with a domain certificate.
```
SeverinAlexB:
```suggestion
- The homeserver exposes two sockets: a **Pubky TLS** socket (`pubky_listen_socket`, default port 6287) and a regular **HTTP** socket (`icann_listen_socket`, default port 6286). Pubky TLS uses PKARR-based TLS and does not need a certificate, so can be exposed directly. The HTTP socket serves browsers and should be put behind a reverse proxy if you need standard HTTPS with a domain certificate.
```

README.md
```diff
@@ -1,88 +1,56 @@
...
 
-
-
-
 <div align="center">
   <h3>
-    <a href="https://pubky.github.io/pubky-core/">
-      Docs Site
-    </a>
+    <a href="https://docs.pubky.org/">Docs</a>
```
SeverinAlexB:
```suggestion
    <a href="https://pubky.org/">Docs</a>
```

SeverinAlexB (commented):
The INSTALL.md is getting pretty good! I run everything on a new Ubuntu server and it worked flawlessly. I added some comments but the most feedback I got from an AI review. I tried to only add the AI comments that are necessary but feel free to just close them if you think they are trash.

docs/INSTALL.md
86667: Hmmm true. Let me leave as is for now with a note to myself to update the suggested version to `0.10` when we release.

docs/INSTALL.md
86667: Limiting guide to Debian-based linux for simplicity. Can add other guides later if they are called for

docs/INSTALL.md
86667: i think that is best for now.

Limiting guide to Debian-based linux for simplicity. Can add other guides later if they are called for.

docs/TESTING.md
86667: Good find Mr. AI. Fixed here - https://github.com/pubky/pubky-core/pull/500

docs/INSTALL.md
86667: I decided to also remove the version set here to force the user to choose one themselves.

SeverinAlexB:
The guides LGTM. Feel free to ask somebody else (outside of core?) for a review if you think it will be of benefit. Otherwise, we can merge

I can't approve my own PR so this is my explicit PR approval

86667 (approved):

andrei-21 (approved):

PR #445, opened 2026-06-17
fix(homeserver): reject PUT and DELETE file with traling / with 400

andrei-21:


MCarlomagno (approved):
LGTM

86667 (commented):
We generally prefer this type of input validation to be done on the deserialisation of the input type itself, so eg a `WebDavFilePathAxum` for write paths which does the validation on init.

86667 (approved):

PR #447, opened 2026-06-19
docs: fix rustdoc links

andrei-21:
Replace broken/private intra-doc links with resolvable links or code
formatting, and wrap bare URLs so `cargo doc` runs cleanly.

### DIFF
--- a/pubky-sdk/src/actors/auth/cookie/legacy_api.rs
+++ b/pubky-sdk/src/actors/auth/cookie/legacy_api.rs
@@ -27,7 +27,7 @@ impl PubkySession {
 
     /// Restore a session from an `export()` string.
     ///
-    /// Delegates to [`super::secret::import_session`].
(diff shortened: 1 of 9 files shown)


pubky-sdk/src/actors/auth/cookie/legacy_api.rs
```diff
@@ -49,7 +49,7 @@ impl PubkySession {
 
     /// Restore a session from a secret token stored in a file.
     ///
-    /// Delegates to [`super::secret::session_from_secret_file`].
+    /// Delegates to `super::secret::session_from_secret_file`.
```
86667, line 52: why remove these?

andrei-21: You mean the link? Because the item is private, you cannot refer to private items in the doc.

86667 (approved):

PR #451, opened 2026-06-22
feat(cli): homeserver cli

SpontaneousOverthrow:
The initial structure for homeserver-cli. Trying to make it scalable and extendable, thats why:
 - each command in separate directory and each subcommand in separate file.
 - it is possible to use config file or use command args. The command args have greater priority.


andrei-21:
@86667 is adding OpenAPI specification for the interface https://github.com/pubky/pubky-core/pull/406.
I wonder if it makes sense to use to generate some Rust code for the cli? There are some libraries for that https://github.com/swagger-api/swagger-codegen.

homeserver-cli/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use clap::Parser;
+use cli::Cli;
+
+fn main() -> anyhow::Result<()> {
+    let cli = Cli::parse();
+
+    // Loading and parsing config only if data_dir flag is provided
+    let data_dir_flag_provided = std::env::args_os()
+        .any(|arg| arg == "--data-dir" || arg == "-d");
```
andrei-21: You should make `Cli::data_dir` `Option<PathBuf>` with no default, then you do not need to check the presence of the flag manually.

homeserver-cli/src/commands/admin/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+            config.and_then(|c| c.admin.admin_endpoint.clone())
+        })
+        .ok_or_else(|| "Error: Missing admin endpoint. Provide it via '--admin-endpoint' or in the config file.".to_string())
+    }
+}
+
+#[derive(Args, Debug)]
+pub struct AdminCmd {
+    #[arg(long)]
+    pub admin_password: Option<String>,
```
andrei-21: It is not generally safe to pass passwords or credentials as parameters or arguments, because they can be stored by the shell. If a password is missing, you can ask the user to type it in securely, for example using https://docs.rs/rpassword.

homeserver-cli/src/commands/admin/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+impl AdminCmd {
+    pub fn run(&self, config: Option<ConfigToml>) {
+        let resolved_admin_password = match self.resolve_admin_password(config.as_ref()) {
+            Ok(e) => e,
+            Err(e) => {
+                eprintln!("{}", e);
+                std::process::exit(1);
+            }
```
andrei-21: Use `anyhow::Result` and return it from everywhere.

homeserver-cli/src/commands/admin/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Subcommand, Debug)]
+pub enum AdminSubcommands {
+    GetInfo(get_info::GetInfoArgs),
+}
+
+impl AdminCmd {
+    pub fn run(&self, config: Option<ConfigToml>) {
+        let resolved_admin_password = match self.resolve_admin_password(config.as_ref()) {
+            Ok(e) => e,
```
andrei-21: We usually do not name `e` an ok result, better `Ok(password)`.

homeserver-cli/src/commands/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+use crate::cli::Cli;
+use crate::config::ConfigToml;
+use clap::Subcommand;
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    Admin(admin::AdminCmd),
+}
+
+pub fn execute(cli: Cli, _config: Option<ConfigToml>) {
```
andrei-21: The `_` prefix in symbol names prevents the compiler from reporting whether the symbol is used, and we do not need it here.

andrei-21 (dismissed):
Is this tool supposed to be used against the admin interface, or the user interface as well?

homeserver-cli/src/commands/admin/mod.rs
SpontaneousOverthrow: yeah, that was more like a mock

SpontaneousOverthrow:
> Is this tool supposed to be used against the admin interface, or the user interface as well?

It supposes to be used against the admin interface and user interface, that's why I have created separate command admin, with list of admin subcommands, like get_info or generate signup_token.

SpontaneousOverthrow:
What is still missing, will be added on next iteration:
- logs
- proper http error handling, on command lvl (instead of 404 will be printed user not found and etc)
- tests
- use quota structure from homeserver

homeserver-cli/src/helpers/http_client.rs
```diff
@@ -0,0 +1,77 @@
...
+        let url = self.url(path)?;
+        self.send(self.http.get(url.clone()), &url)
+    }
+
+    pub fn post_json<B: Serialize>(&self, path: &str, body: &B) -> Result<Response> {
+        let url = self.url(path)?;
+        self.send(self.http.post(url.clone()).json(body), &url)
+    }
+
+    #[allow(dead_code)] // used by unauthenticated commands (upcoming)
```
andrei-21: It is already used for enabling/disabling users.

homeserver-cli/src/commands/admin/user/disable.rs
```diff
@@ -0,0 +1,16 @@
...
+use pubky::PublicKey;
+
+#[derive(Args, Debug)]
+pub struct DisableArgs {
+    pub pubky: PublicKey,
+}
+
+pub fn run(context: AdminContext, args: &DisableArgs) -> Result<()> {
+    let pk = args.pubky.z32();
+    context.client.get(&format!("users/{}", pk))?;
```
andrei-21: It should `post` no?

homeserver-cli/src/helpers/http_client.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    fn send(&self, request: RequestBuilder, url: &Url) -> Result<Response> {
+        let response = self
+            .apply_auth(request)
+            .send()
+            .with_context(|| format!("request to {url} failed"))?;
+
+        if !response.status().is_success() {
+            bail!("{} returned {}", url, response.status());
+        }
```
andrei-21:
```suggestion
        ensure!(response.status().is_success(), format!("{} returned {}", url, response.status());
```

homeserver-cli/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+anyhow.workspace = true
+clap = { workspace = true, features = ["derive"] }
+serde = { version = "1", features = ["derive"] }
+url = { version = "2.5", features = ["serde"] }
+toml = "0.8"
+dirs = "6.0"
+rpassword = "7.5.4"
+reqwest = { workspace = true, features = ["blocking", "json"] }
+serde_json = "1.0.151"
+pubky = { version = "0.9.0", path = "../pubky-sdk" }
```
andrei-21: For simplicity we decided to specify only major version (unless it `0`, then `0.minor`). Like `rpassword = "7"`.

homeservercli/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct ConfigToml {
+    pub admin: AdminToml,
+}
+
+impl ConfigToml {
+    pub fn load(data_dir: Option<&Path>) -> Result<Option<Self>> {
+        let Some(dir) = data_dir else {
+            return Ok(None);
+        };
```
andrei-21, line 45: If the function returns `None` when the argument is `None` then there is no need to ask for `Option`. Simply:
```rust
pub fn load(data_dir: &Path) -> Result<Self>
```

homeserver-cli/src/helpers/quota.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let unit_start = amount_part
+            .find(|c: char| !c.is_ascii_digit())
+            .ok_or_else(|| err_msg(s))?;
+        let (number, unit) = amount_part.split_at(unit_start);
+
+        if number.is_empty() || number.parse::<u64>().is_err() {
+            return Err(err_msg(s));
+        }
+        if !matches!(unit, "b" | "kb" | "mb" | "gb") {
```
andrei-21: From AI:
> For valid server input such as `500mb/d`, the CLI rejects the quota because periods are limited to `s|m|h`, while the homeserver supports `d`; conversely this parser accepts `b`, zero, and values above `u32`, which the server rejects with 422. Reuse the server parser or mirror its nonzero-u32, `kb|mb|gb`, `s|m|h|d` grammar.

andrei-21 (commented):
Did you run `/review` command in codex? It still finds some issues.

homeserver-cli/src/commands/admin/user/disable.rs
SpontaneousOverthrow: yes

homeservercli/src/config.rs
SpontaneousOverthrow: It should be Option, later I pass that config, and if config is provided the values from config are used if no, flag values are used.

homeserver-cli/README.md
```diff
@@ -0,0 +1,150 @@
...
+```sh
+homeserver-cli --data-dir /path/to/config/dir admin info
+```
+
+Alternatively, supply credentials directly via flags (password will be prompted interactively):
+
+```sh
+homeserver-cli admin --admin-endpoint https://your-homeserver.example.com --admin-password info
+```
+
```
SeverinAlexB: I like the general approach of getting the credentials. data dir + passing it via CLI args makes sense. Do we need env variable support too? Devops people usually like env variable so you are the best person to judge this requirement :)


I do have on problem with the current cli though: Getting started is relatively high friction currently. The most common case of using the CLI will be alongside the homeserver binary within the same linux user.

So `--data-dir` should default to `~/.pubky/`, the same default the homeserver uses. It should work flawlessly with the homeserver `config.toml`, support the `listen_socket` variable, use the same defaults if they are commented out.

Basically support this file:

```toml
[admin]
admin_password = "your-admin-password"
listen_socket = "127.0.0.1:6288"
admin_endpoint = "https://your-homeserver.example.com"
```

If no `admin_password` is set then `admin` is used, same as the default homeserver admin password.
Use `admin_endpoint` if available. If it does exist, use `listen_socket` and construct the admin endpoint from it. If `listen_socket` is missing, use the default value for it.

Do I make sense? Please make sure to understand my reasoning.

homeserver-cli/README.md
```diff
@@ -0,0 +1,150 @@
...
+| Flag | Description |
+|------|-------------|
+| `-d, --data-dir <PATH>` | Directory containing `config.toml` |
+| `--admin-password` | Prompt for admin password interactively |
+| `--admin-endpoint <URL>` | Admin API base URL |
+| `-v` / `-q` | Increase / decrease log verbosity |
+
+---
+
+### `admin info`
```
SeverinAlexB: Putting everything behind the admin keyword is a mistake IMO. 90% of the commands will be behind it, making `admin` repetitive/annoying to type.

 I suggest the following structure:

```
Pubky homeserver cli.

Usage: homeserver-cli [OPTIONS] <COMMAND>

Commands:
  info    Print homeserver statistics (users, disk usage, signup codes, version)
  signup-token  Manage signup invite tokens
  user          Manage user accounts
  help   Print this message or the help of the given subcommand(s)

Options:
  -d, --data-dir <DATA_DIR>  // Needs a description here 👈
  --admin-password       // Needs a description here 👈             
  --admin-endpoint <ADMIN_ENDPOINT> // Needs a description here 👈
  -v, --verbose...           Increase logging verbosity
  -q, --quiet...             Decrease logging verbosity
  -h, --help                 Print help
  -V, --version              Print version
```

- Basically, we should make the admin command the top level.
- `quota` handles user quotas so it should be in the `user` command.
- I am not 100% sure but shouldn't `signup-token` and `user` be in plural aka `signup-tokens` and `users`? @86667 I need your native english skills.

homeserver-cli/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Print homeserver statistics.
+
+```sh
+homeserver-cli admin info
+```
```
SeverinAlexB: Should the CLI output be in JSON? This would make it easy to use the output to do admin automation with tools like `jq`. LND/Core-Lightning so it similarly. Bitcoin-Core too. @andrei-21 @SpontaneousOverthrow  what do you think?

In theory then, one theoretically does not even need to type-parse the http response with struct and can just output the response directly. This would make the CLI leaner + would reduce the work that is needed in case we add new fields to the admin api. 

Current output:
```
Homeserver Admin Info
---------------------
Users:            23 (0 disabled)
Disk used:        0.0 MB
Signup codes:     3 (1 unused)
Public key:       6ciddd8xg6upz5r5odzm7qes68jo4c8xzqz3qp3y6xp65gsr14dy
Pkarr address:    127.0.0.1:6287
Pkarr icann:      localhost:6286
Version:          0.9.3
```

homeserver-cli/src/commands/admin/signup_token/generate.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::commands::admin::signup_token::error::map_http;
+use crate::helpers::quota::{Quota, QuotaUpdate, RateLimit};
+use anyhow::{Context, Result};
+use clap::Args;
+
+#[derive(Args, Debug)]
+#[command(about = "Generate a signup invite token with optional quota overrides")]
+pub struct GenerateArgs {
+    /// Storage quota in MB for the invited user. Use a number (e.g. 500) or "unlimited".
+    /// Omit to apply the system default.
```
SeverinAlexB: When reading this the question that pops up in my head is: "What is the system default?". Not 100% sure but is it defined in the config? If yes, can you improve the text with "Omit to apply the system default which is defined in the homeserver config."?

homeservercli/src/config.rs
```diff
@@ -0,0 +1,43 @@
+use anyhow::{Context, Result};
```
SeverinAlexB, line 1: NIT: Turn this file into a config.rs without the folder. Having a folder + a mod.rs file is unnecessary.

homeserver-cli/src/commands/admin/quota/get.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::commands::admin::quota::error::map_http;
+use crate::helpers::quota::UserQuota;
+use anyhow::{Context, Result};
+use clap::Args;
+use pubky::PublicKey;
+
+#[derive(Args, Debug)]
+#[command(about = "Show the effective quota for a user")]
+pub struct GetArgs {
+    pub pubky: PublicKey,
```
SeverinAlexB: This needs a description.

Also, as a general rule, `Pubky` is the brand, `public_key` or `pubkey` is a public key. So please try to stick with one of these two. This might affect other parts of the CLI.

homeserver-cli/src/commands/admin/quota/get.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+pub fn run(context: AdminContext, args: &GetArgs) -> Result<()> {
+    let pk = args.pubky.z32();
```
SeverinAlexB: NIT: Using two letter shortcuts for variable is bad. `pubkey` or `public_key` is better. See https://medium.com/@pabashani.herath/clean-code-naming-conventions-4cac223de3c6

homeservercli/src/commands/users/quotaset.rs
```diff
@@ -0,0 +1,68 @@
...
+    ArgGroup::new("quota_fields")
+        .required(true)
+        .multiple(true)
+        .args(["storage_quota_mb", "rate_read", "rate_write", "rate_read_burst", "rate_write_burst", "allowed_write_paths"]),
+))]
+pub struct SetArgs {
+    /// Public key of the user to update (z-base-32 encoded).
+    pub pubky: PublicKey,
+
+    /// Storage quota in MB. Use a number (e.g. 500) or "unlimited".
```
SeverinAlexB: Missing default value in a lot of the descriptions.

homeserver-cli/src/commands/admin/quota/set.rs
```diff
@@ -0,0 +1,68 @@
...
+        rate_write_burst: args.rate_write_burst,
+        allowed_write_paths: args.allowed_write_paths.clone(),
+    };
+
+    context
+        .client
+        .patch_json(&format!("users/{}/quota", pk), &body)
+        .map_err(map_http)?;
+
+    println!("updated quota for user: {}", pk);
```
SeverinAlexB: I am not sure how the quota endpoint works but usually, the patch call returns the new value which then can be output in the cli.

homeservercli/src/commands/users/disable.rs
```diff
@@ -0,0 +1,21 @@
+use crate::commands::admin::context::AdminContext;
+use crate::commands::admin::user::error::map_http;
+use anyhow::Result;
+use clap::Args;
+use pubky::PublicKey;
+
+#[derive(Args, Debug)]
+#[command(about = "Disable a user account")]
+pub struct DisableArgs {
+    pub pubky: PublicKey,
```
SeverinAlexB, line 11: Missing description

homeservercli/src/helpers/errors.rs
```diff
@@ -0,0 +1,24 @@
+use thiserror::Error;
+use url::Url;
+
+#[derive(Error, Debug)]
+#[error("{url} returned {status}")]
+pub struct HttpStatusError {
+    pub status: u16,
+    pub url: Url,
+}
```
SeverinAlexB, line 10: Looks like `HttpStatusError` is omitting the response body/text which can be very helpful in figuring out the problem. Only relying on the status code is suboptimal. Can we add this here?

homeserver-cli/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "homeserver-cli"
+description = "Pubky homeserver cli."
+version = "0.0.1"
```
SeverinAlexB:
```suggestion
version = "0.1.0"
```

homeserver-cli/config.sample.toml
```diff
@@ -0,0 +1,3 @@
+[admin]
+admin_password="config"
+admin_endpoint="https://config.com"
```
SeverinAlexB: Use the same defaults as the homeserver defines so people are not going to be confused.
```suggestion
[admin]
admin_password="admin"
admin_endpoint="http://127.0.0.1:6288"
```

homeservercli/src/helpers/http_client.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        let http = Client::builder()
+            .timeout(Duration::from_secs(30))
+            .user_agent(concat!(
+                env!("CARGO_PKG_NAME"),
+                "/",
+                env!("CARGO_PKG_VERSION")
+            ))
+            .build()
+            .context("failed to build HTTP client")?;
```
SeverinAlexB, line 62: NIT: Disable redirect to prevent attacks where the password header is redirected

```
.redirect(reqwest::redirect::Policy::none())
```

homeservercli/src/commands/users/quotaset.rs
```diff
@@ -0,0 +1,68 @@
...
+    let pk = args.pubky.z32();
+
+    let body = QuotaUpdate {
+        storage_quota_mb: args.storage_quota_mb,
+        rate_read: args.rate_read.clone(),
+        rate_write: args.rate_write.clone(),
+        rate_read_burst: args.rate_read_burst,
+        rate_write_burst: args.rate_write_burst,
+        allowed_write_paths: args.allowed_write_paths.clone(),
+    };
```
SeverinAlexB, line 62: AI review, no idea if this is an issue:

Medium: Quota PATCH cannot reset overrides or configure read-only access  
homeserver-cli/src/helpers/quota.rs:86-100, homeserver-cli/src/commands/admin/quota/set.rs:51-58  
The server distinguishes:
- absent: preserve existing value
- null: restore system default or unrestricted paths
- []: make the account read-only
- value: set an override
QuotaUpdate omits every None and empty path vector, so none of the reset/read-only operations can be expressed. This can also prevent changing a rate to unlimited when a stale burst remains configured. The PATCH request needs an explicit three-state representation.

homeserver-cli/src/commands/admin/info.rs
```diff
@@ -0,0 +1,60 @@
...
+    num_users: u64,
+    num_disabled_users: u64,
+    total_disk_used_mb: f64,
+    num_signup_codes: u64,
+    num_unused_signup_codes: u64,
+    public_key: String,
+    pkarr_pubky_address: String,
+    pkarr_icann_domain: String,
+    version: String,
+}
```
SeverinAlexB: AI review, no idea if this is an actual issue:

Medium: admin info rejects valid responses with unset PKARR metadata  
homeserver-cli/src/commands/admin/info.rs:12-23  
pkarr_pubky_address and pkarr_icann_domain are required Strings, while the server returns null when either setting is absent (pubky-homeserver/src/admin_server/routes/info.rs:16-17). Such a successful response fails deserialization. These fields should be Option<String> with suitable display output.

SeverinAlexB (changes requested):
Very well structured code. I added my feedback mostly from an UX POV. Very few code issues.

homeserver-cli/README.md
SpontaneousOverthrow: I agree that we need to support the env variables too.
Related to defaults variables, I would suggest to use env_vars->flags->config not overcomplicate logic, from my experience it was hard to debug things where you have hardcoded defaults.
For example you misspelled the endpoint value in config file and instead of error on start you will spend some time to investigate why you have errors. 

What I suggest to add env variables support, add listen_socket support for config.

homeserver-cli/README.md
SpontaneousOverthrow: I wanted to make the `--format` flag  where you can provide format that you want, for example `json`, `toml` and etc
The docker has something similar, but more advanced:
`docker images repo1 --format "{{json . }}"`

homeserver-cli/src/commands/admin/signup_token/generate.rs
SpontaneousOverthrow: yes, good catch, will do that

homeservercli/src/config.rs
SpontaneousOverthrow: I was thinking that in the future we may want to make config more complex with additional structures\functions, that's why I moved it to separate dir.

homeserver-cli/Cargo.toml
SpontaneousOverthrow: will fix

homeserver-cli/config.sample.toml
SpontaneousOverthrow: yes, good catch

homeservercli/src/helpers/http_client.rs
SpontaneousOverthrow: yeah, good catch

homeservercli/src/commands/users/quotaset.rs
SpontaneousOverthrow: need a bit more investigation here, gonna do it

homeserver-cli/src/commands/admin/info.rs
SpontaneousOverthrow: hm, sounds reasonable, gonna test it and fix if needed.

homeserver-cli/src/commands/admin/quota/get.rs
SpontaneousOverthrow: yes, gonna fix it

homeserver-cli/src/commands/admin/quota/set.rs
SpontaneousOverthrow: that's correct, the new values are returned. will add them to cli output

homeservercli/src/helpers/errors.rs
SpontaneousOverthrow: yes, will add it

homeserver-cli/README.md
SpontaneousOverthrow: I am totally agree with moving quota to user command.
Related to admin, I was thinking about it, but decided to keep, because that layer adds more freedom for scale and makes config management a bit easier.

homeserver-cli/README.md
SeverinAlexB: > env_vars->flags->config 

This hierarchy sounds reasonable.

 I am not sure I understand the rest of what your saying though? Do you argue for or against my `~/.pubky/config.toml` default proposal?

homeserver-cli/README.md
SeverinAlexB: > Related to admin, I was thinking about it, but decided to keep, because that layer adds more freedom for scale and makes config management a bit easier

I feel very strongly about moving admin up. It will be a big mistake. Already testing the cli, I was annoyed needing to type admin many times. Scaling is not issue at all with this CLI IMO.

homeserver-cli/README.md
SeverinAlexB: I would keep it keep it simple and stupid (KISS). JSON only especially at this stage of development. I want to avoid the unnecessary work of us having to maintain the outputs.

homeservercli/src/config.rs
SeverinAlexB: That's over-engineering. We can create a directory in the future when this is actually going to be the case. At the moment it's not needed so it should be a file.

homeserver-cli/README.md
SpontaneousOverthrow: I am against that part:
```
If no admin_password is set then admin is used, same as the default homeserver admin password.
Use admin_endpoint if available. If it does exist, use listen_socket and construct the admin endpoint from it. If listen_socket is missing, use the default value for it.
```
We will have very sophisticated structure in that case:
env_vars->flags->config->hardcoded defaults
Also if someday for some reason we update these defaults for homeserver we should update them here too

homeservercli/src/config.rs
SpontaneousOverthrow: ok, makes sense

homeserver-cli/README.md
SpontaneousOverthrow: It seems it's not hard to support different output formats. I can create a custom type for handling the output format and support only JSON for now.

homeserver-cli/README.md
SeverinAlexB: 1. From a UX point of view, this is required.
2. This is really not that complex.
3. Yes, if we updated the default values in the homserver, we need to update them here too.

The UX goal is that the cli works automatically if the user already has a valid `~/.pubky/config.toml.

SpontaneousOverthrow: Related to 3 seems it could be easily solved using, admin-pubky-common where all shared data types will be stored.

> For example you misspelled the endpoint value in config file and instead of error on start you will spend some time to investigate why you have errors.

From my own experience this is not a good UX for cli, because of the comment above.

SpontaneousOverthrow: > The UX goal is that the cli works automatically if the user already has a valid `~/.pubky/config.toml.


Seems this is solved only using default value for config path, no need to provide defaults if there is no config, or some fields in config are missed.

homeserver-cli/src/commands/admin/quota/set.rs
SpontaneousOverthrow: in our current implementation the set quota doesn't return new values:
https://github.com/pubky/pubky-homeserver/blob/main/pubky-homeserver/src/admin_server/routes/user_quota.rs#L74
this should be implemented on homeserver lvl first

homeservercli/src/commands/users/disable.rs
SpontaneousOverthrow: what do you mean here?

SpontaneousOverthrow: oh, I have changed name to public_key so no description is needed right?

homeservercli/src/helpers/errors.rs
SpontaneousOverthrow: Not sure that it's needed, based on https://github.com/pubky/pubky-homeserver/blob/f8e74181416d75d3bfc5d189e82360c0e4d88ba9/pubky-homeserver/openapi-admin.yml#L4
we don't have any error responses that has some data in body, we can implement it when we will have such methods in the homeserver

homeservercli/src/commands/users/quotaset.rs
SpontaneousOverthrow: added, that defaults are provided in the homeserver config,

homeservercli/src/config.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct AdminToml {
+    pub admin_password: Option<String>,
+    pub admin_endpoint: Option<Url>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct ConfigToml {
+    pub admin: AdminToml,
+}
```
SeverinAlexB: In the homeserver config, the `[admin]` section is optional. Here you require it. The admin section should be optional here too so it works out of the box for most configs.

```
Error: failed to parse config file: /Users/severinbuhler/.pubky/config.toml

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | [general]
      | ^
    missing field `admin`
```

It also looks like that the default password nor the default endpoint is set. Even if I add a `[admin]` into my config, it will not take the default password.

```
cargo run -- info
   Compiling homeservercli v0.1.0 (/Users/severinbuhler/git/pubky/pubky-core/homeservercli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.98s
     Running `/Users/severinbuhler/git/pubky/pubky-core/target/debug/homeservercli info`
Error: Missing admin password. Provide it via '--admin-password', the PUBKY_HOMESERVER_ADMIN_PASSWORD environment variable, or in the config file.
```


Even if I add both fields to the config, it does not work:

Image

SeverinAlexB (changes requested):

homeservercli/src/commands/info.rs
```diff
@@ -0,0 +1,14 @@
...
+use anyhow::Result;
+use clap::Args;
+
+#[derive(Args, Debug)]
+#[command(about = "Print homeserver statistics (users, disk usage, signup codes, version)")]
+pub struct InfoArgs {}
+
+pub fn run(context: AdminContext, _args: &InfoArgs) -> Result<()> {
+    let response = context.client.get("info").map_err(map_http)?;
+    println!("{}", response.text()?);
```
SeverinAlexB: Please pretty print any received JSON before you output it. It's hard to read otherwise

homeservercli/src/commands/signup_token/generate.rs
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    let token = context
+        .client
+        .post_json("generate_signup_token", &body)
+        .map_err(map_http)?
+        .text()
+        .context("failed to read signup token response")?;
+
+    println!("invite code: {token}");
```
SeverinAlexB: Missing JSON output. Probably should just be the plain token.

homeservercli/src/config.rs
SpontaneousOverthrow: I understood it wrong, will fix

homeserver-cli/src/commands/admin/quota/set.rs
SeverinAlexB: Yeahh, just print the response. Then we can change the homeserver response dynamically.

homeservercli/src/config.rs
SeverinAlexB: Doesn't work with a missing header.

```
➜  homeservercli (feat/homeserver-cli) cargo run -- info
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `/Users/severinbuhler/git/pubky/pubky-core/target/debug/homeservercli info`
Error: failed to parse config file: /Users/severinbuhler/.pubky/config.toml

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | [general]
      | ^
    missing field `admin`
```

The `[admin]`  header also needs to be optional in the .pubky/config.toml

---

Also, it does not respect the `admin.listen_socket`.

Please read our conversation https://github.com/pubky/pubky-homeserver/pull/451#discussion_r3674810811. If you have questions, let me know.

homeservercli/src/commands/users/disable.rs
SeverinAlexB: NIT: To make things complete, I suggested to add a doc string to `pubky` here.

Example:
```suggestion
    /// Public key of the user (z-base-32 encoded).
    pub pubky: PublicKey,
```

homeservercli/src/helpers/errors.rs
SeverinAlexB: Just add it please. I am pretty sure the openapi is just not complete. At the same time, the error does not necessarily need to come from the homeserver. It can come from the reverse proxy too.

e2e/src/tests/cli.rs
```diff
@@ -0,0 +1,267 @@
+use std::path::{Path, PathBuf};
+use std::process::{Command, Output};
```
SeverinAlexB, line 2: AI review:

Medium: CLI integration tests are not executed. e2e/src/tests/cli.rs exists, but e2e/src/tests/mod.rs:1 never declares mod cli, and e2e/Cargo.toml:12 lacks its escargot and tempfile dependencies. The tests also use obsolete commands such as quota set instead of users quota-set (e2e/src/tests/cli.rs:95). Consequently, command parsing and server round trips currently have no process-level coverage.

homeservercli/src/commands/users/quotaset.rs
```diff
@@ -0,0 +1,72 @@
...
+pub fn run(context: AdminContext, args: &SetArgs) -> Result<()> {
+    let public_key = args.public_key.z32();
+
+    let body = QuotaUpdate {
+        storage_quota_mb: args.storage_quota_mb,
+        rate_read: args.rate_read.clone(),
+        rate_write: args.rate_write.clone(),
+        rate_read_burst: args.rate_read_burst,
+        rate_write_burst: args.rate_write_burst,
+        allowed_write_paths: args.allowed_write_paths.clone(),
```
SeverinAlexB, line 61: AI review:

Medium: quota overrides cannot be reset to system defaults. QuotaUpdate uses plain Option fields and omits None during serialization (homeservercli/src/helpers/quota.rs:86). The server API requires explicit JSON null to reset storage, rates, and bursts, but users quota-set can only send a value or "unlimited" (homeservercli/src/commands/users/quotaset.rs:54). Once an override is configured, administrators cannot restore inherited defaults through the CLI.

SeverinAlexB (changes requested):

homeservercli/src/commands/info.rs
SpontaneousOverthrow: it is already updated

SeverinAlexB (approved):
Let's go 🚀

Issue #432, opened 2026-06-17
Accept `/priv/` for writes


MCarlomagno:
### Context
Writes are hardcoded to /pub/ in two places. To allow `/priv/`writes we relax both and keep capability enforcement (a write cap must cover the target path) unchanged.

### Scope
Allow `PUT/DELETE` under `/priv/<scope>/...` in addition to `/pub/<scope>/....`

Reject any path whose first segment is not pub or priv (preserve today's 403 contract for other roots, e.g. /foo/...).

### Acceptance criteria
- [ ] `PUT /priv/app/x` with a cap covering `/priv/app/` → `201`.
- [ ] `PUT /priv/app/x` with only `/pub/...` caps → `403`.
- [ ] `PUT /foo/x` (neither root) → `403` with the existing "directories other than ..." message,
      updated to mention both roots.
- [ ] Existing `/pub/` write tests still pass (`authorization.rs` test module updated).

## Dependencies

Part of epic #426.

Blocked by: _nothing — foundational; unblocks the rest of the epic._

PR #454, opened 2026-06-23
feat(homeserver): filter private events from `/events` endpoint

### LINKED ISSUE #436: Stop the `GET /events/` feed leaking `/priv/` paths
### Context
`GET /events/` returns **all** events by cursor with **no user, path, or auth filter**. Once `/priv/` writes emit events into the same table, this unauthenticated endpoint dumps private paths (`PUT pubky://user/priv/...`).

### Scope
- The unauthenticated legacy feed must **exclude** events whose path is under `/priv/...`.
- Implement by filtering in the query (`get_by_cursor`) or in `feed` (path `starts_with("/priv/")`).
  Prefer the query so cursor/limit accounting stays correct after filtering.
- (Optional, can be a follow-up) allow an authenticated caller to receive their own private events
  via this endpoint — but the **default/anonymous** path must be public-only.

### Acceptance criteria
- [ ] After a `PUT /priv/app/x`, `GET /events/` (anonymous) does **not** include that path at any
      cursor position.
- [ ] `GET /events/` still returns all `/pub/...` events with correct cursor continuation.
- [ ] Cursor/limit pagination remains correct when private events are interleaved and filtered out
      (no short pages that skip the limit, no infinite loops).

## Dependencies

Part of epic #426.

Blocked by:
- #432 — Accept `/priv/` for writes (private events must exist to be filtered out)

MCarlomagno:
## Summary

Closes [#436](https://github.com/pubky/pubky-core/issues/436)

`GET /events/` is a global, unauthenticated feed that returned all events by cursor with no path or auth filter. Now that `/priv/` writes emit events into the same table ([#432](https://github.com/pubky/pubky-core/issues/432)), this endpoint would dump private paths (`PUT pubky://user/priv/...`) to any anonymous caller.

This PR makes the legacy feed public only: it returns `/pub/...` events exclusively and never exposes `/priv/...`, even to an authenticated caller — matching the Private Storage spec ("/events/ must become public-only"; private events are served only via the authenticated `/event-stream` endpoint).

### Acceptance criteria
- [x] After a `PUT /priv/app/x`, `GET /events/` (anonymous) does **not** include that path at any
      cursor position.
- [x] `GET /events/` still returns all `/pub/...` events with correct cursor continuation.
- [x] Cursor/limit pagination remains correct when private events are interleaved and filtered out
      (no short pages that skip the limit, no infinite loops).

### DIFF
--- a/pubky-homeserver/src/client_server/auth/authorization.rs
+++ b/pubky-homeserver/src/client_server/auth/authorization.rs
@@ -20,13 +20,9 @@ use pubky_common::capabilities::Action;
 
 use crate::client_server::auth::AuthSession;
 use crate::client_server::middleware::pubky_host::PubkyHost;
+use crate::constants::{PRIVATE_ROOT, PUBLIC_ROOT};
 use crate::shared::webdav::WebDavPath;
 use crate::shared::HttpError;
-
-/// `/pub/` is public-readable storage, anonymous reads are allowed.
-const PUBLIC_ROOT: &str = "/pub/";
-/// `/priv/` is private storage, reads and writes require auth.
-const PRIVATE_ROOT: &str = "/priv/";
 
 /// Storage roots a write may target.
 const STORAGE_ROOTS: [&str; 2] = [PUBLIC_ROOT, PRIVATE_ROOT];
--- a/pubky-homeserver/src/client_server/routes/events.rs
+++ b/pubky-homeserver/src/client_server/routes/events.rs
@@ -238,6 +238,10 @@ fn event_to_sse_data(entity: &EventEntity) -> String {
 
 /// Legacy text-based endpoint for fetching historical events.
 ///
+/// This feed is public and unauthenticated: it returns events under
+/// `/pub/...` exclusively and never exposes private (`/priv/...`) paths, even to
+/// an authenticated caller.
+///
 /// ## Query Parameters
 /// - `cursor` (optional): Starting cursor position. Default: "0" (beginning)
 /// - `limit` (optional): Maximum number of events to return
@@ -271,7 +275,7 @@ pub async fn feed(
     let query_start = Instant::now();
     let events = state
         .events_service
-        .get_by_cursor(Some(cursor), params.limit, &mut state.sql_db.pool().into())
+        .get_public_by_cursor(Some(cursor), params.limit, &mut state.sql_db.pool().into())
         .await?;
     state
         .metrics
--- a/pubky-homeserver/src/persistence/files/events/events_repository.rs
+++ b/pubky-homeserver/src/persistence/files/events/events_repository.rs
@@ -9,7 +9,7 @@ use sqlx::{
 };
 
 use crate::{
-    constants::{DEFAULT_LIST_LIMIT, DEFAULT_MAX_LIST_LIMIT},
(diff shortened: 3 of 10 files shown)


pubky-homeserver/src/client_server/auth/authorization.rs
```diff
@@ -20,14 +20,10 @@ use pubky_common::capabilities::Action;
...
 use crate::client_server::middleware::pubky_host::PubkyHost;
+use crate::constants::{PRIVATE_ROOT, PUBLIC_ROOT};
 use crate::shared::webdav::WebDavPath;
 use crate::shared::HttpError;
 
-/// `/pub/` is public-readable storage, anonymous reads are allowed.
-const PUBLIC_ROOT: &str = "/pub/";
-/// `/priv/` is private storage, reads and writes require auth.
-const PRIVATE_ROOT: &str = "/priv/";
-
```
MCarlomagno, line 30: I moved these to constants as they are also used in the `EventsService` now to filter out `/pub` prefixes

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -271,7 +275,7 @@ pub async fn feed(
     let query_start = Instant::now();
     let events = state
         .events_service
-        .get_by_cursor(Some(cursor), params.limit, &mut state.sql_db.pool().into())
+        .get_public_by_cursor(Some(cursor), params.limit, &mut state.sql_db.pool().into())
```
MCarlomagno, line 278: we could technically still use `get_by_cursor` instead of creating a new function, like:

`.get_by_cursor(Some(cursor), params.limit, EventVisibility::Public ,&mut state.sql_db.pool().into())`

But I liked more `get_public_by_cursor` as it's more explicit, and there is less risk of passing a wrong argument (ALL). But open to change it if I get different opinions

pubky-homeserver/src/persistence/files/events/events_repository.rs
```diff
@@ -22,6 +22,15 @@ use crate::{
...
 pub const EVENT_TABLE: &str = "events";
 
+/// Selects which events a cursor query returns, by storage-root visibility.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EventVisibility {
+    /// Only public (`/pub/...`) events.
+    Public,
+    /// All events, includes `/priv/...` events.
+    All,
+}
```
MCarlomagno, line 32: There could be a `Private` variant, but I didn't find any use case for that yet, we can add it if there is a specific case for that

pubky-homeserver/src/persistence/files/events/events_repository.rs
```diff
@@ -286,6 +297,16 @@ impl EventRepository {
...
             .to_owned();
+
+        // `All` adds no predicate; `Public` restricts to the public root.
+        if let EventVisibility::Public = visibility {
+            statement = statement
+                .and_where(
+                    Expr::col((EVENT_TABLE, EventIden::Path)).like(format!("{PUBLIC_ROOT}%")),
+                )
+                .to_owned();
+        }
```
MCarlomagno, line 308: main change of the PR, here we filter out public events based on the `visibility` argument

andrei-21 (approved):

PR #455, opened 2026-06-23
chore(deps): remove unused crate dependencies

andrei-21:
Enable unused crate dependency warnings for crates where the lint is
useful, then remove dependencies that are no longer referenced directly.

Move test-only dependencies out of normal dependency sets where possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1598,15 +1598,11 @@ dependencies = [
  "bytes",
  "eventsource-stream",
  "futures",
- "pkarr",
- "pubky-common",
  "pubky-testnet",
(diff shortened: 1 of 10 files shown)


MCarlomagno (approved):
LGTM, I ran all the tests and double checked using AI

andrei-21:
Will @86667 take an extra look to ensure none of the dependencies are used in a weird way and that they are safe to remove?
The last time I touched dependencies, I broke `pkarr`.

86667 (approved):
Looks reasonable and I double checked with AI.

Issue #438, opened 2026-06-17
SDK: subscribe to private events


MCarlomagno:
### Context
`EventStreamBuilder` builds public `/events-stream` subscriptions today. To consume private events it needs per-path filters plus the caller's session credential.

### Scope
- Add path filtering to `EventStreamBuilder` (e.g. `.path("/priv/app/")`, repeatable), mapping to
  the multiple `path` query from #429.
- Attach the session credential so the server can authorize `/priv/...` paths.
- Keep the public, anonymous subscription working unchanged.

### Acceptance criteria
- [ ] A subscription with a `/priv/app/` path + session receives that user's private events.
- [ ] The same subscription **without** a covering cap receives none (server-enforced).
- [ ] Public subscriptions (no path / `/pub/`) behave as before.

## Dependencies

Part of epic #426.

Blocked by:
- #429 — Multiple `path` parameters for `/events-stream` (server-side per-path authz)

MCarlomagno:
Design point from internal discussion that lands here.

Methods (names TBD):
- `events_for_user_pub(pubky)` → `?user=<u>` (public-only, no session)
- `events_for_user_all(pubky, session)` → attaches credential, requests the
  `/priv/...` scopes the session is capable of
- `events_for_all_users_pub()` → global public

Scope:
1. Make `EventStreamBuilder.path` repeatable + document file vs directory semantics
   (trailing `/` = dir + descendants; no slash = exact file). Today it's a single
   `path: Option<String>` documented as a "prefix", stale after #429.
2. Attach the session credential so the HS can authorize `/priv/` paths.
3. Surface HS 401/403/400 as clear typed errors, not raw HTTP.
4. Public anonymous subscriptions unchanged.

Forward compatibility: dont bake the current "one user when requesting `/priv/`" limit into the builder's shape, a likely follow up makes each key public always, private only if it's the session user), so accepting multiple keys now avoids an SDK break later.

PR #458, opened 2026-06-24
feat(homeserver): add private event stream reads

### LINKED ISSUE #429: Multiple path Parameters for /events-stream
Allow to provide multiple path parameters when subscribing to events:
```
/events-stream?user=<user>&path=/pub/&path=/priv/app1/
```

**Behavior**

- each path is authorized independently;
- public paths require no read capability;
- private paths require read capability for that exact directory/file scope;
- the response is the union of matching events;
- one cursor advances over the combined stream.

**Status codes** (same precedence as storage reads): a private `path` with no session → `401`; a
private `path` whose scope the session's caps don't cover, or a `user=` that isn't the authenticated
session user → `403`; a malformed request (e.g. a private `path` with no `user=`) → `400`.
Unauthenticated or no-`path` streams return only `/pub/...` events.

## Dependencies

Part of epic #426.

Blocked by:
- #433 — Authenticated reads for `/priv/` (shares the read-capability path authorization)

MCarlomagno:
## Summary

Closes [#429](https://github.com/pubky/pubky-core/issues/429)

This PR makes the endpoint accept multiple `path=` filters, authorizes each one independently, and serves private events only to callers that can prove read access, applying the same authorized filter set to both the historical replay and the live SSE phase, so `live=true` can't leak.

Behavior:
- `path=` is repeatable (`?user=<u>&path=/pub/&path=/priv/app/`) the response is the union of matching events over one cursor.
- A trailing slash matches a directory and its descendants (`/pub/files/` → `/pub/files/doc.txt`) without it the path is matched exactly as a file.
- Empty `path=` values are ignored, so an anonymous or unfiltered stream never exposes `/priv/`.
- `/pub/` filters need no session. Any `/priv/` filter requires a session, exactly one `user=` equal to the session user, and a read capability covering that path. If any requested private path is unauthorized, the whole subscription is rejected.

## Acceptance criteria
- [x] Repeated `path=` params return the union of matching events over a single cursor.
- [x] Unauthenticated or no-`path` streams return only `/pub/...` events — historical **and** live.
- [x] `/pub/` paths need no read capability.
- [x] `/priv/` path with a covering read cap (same tenant) → streams those events.
- [x] `/priv/` path with no session → `401`.
- [x] `/priv/` path with a `user=` that isn't the session user, or more than one `user=` → `403`.
- [x] `/priv/` path the session's caps don't cover → `403`.
- [x] Missing `user=` → `400`.
- [x] `live=true` uses the same authorized filter set as the replay and cannot leak private events.

## Note
SDK support for private paths is intentionally out of scope (follow-up [#438](https://github.com/pubky/pubky-core/issues/438)).


pubky-homeserver/src/client_server/auth/authorization.rs
```diff
@@ -40,7 +40,7 @@ const STORAGE_ROOTS: [&str; 2] = [PUBLIC_ROOT, PRIVATE_ROOT];
 /// is forbidden"` rather than axum's default 400 for a deserialization failure.
 pub fn has_write_permission(
     session: &AuthSession,
-    pubky: &PubkyHost,
+    pubkey: &PublicKey,
```
MCarlomagno, line 43: refactored to accept public key instead of the whole PubkyHost, we only need the public key internally and simplifies the auth logic

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -84,7 +91,8 @@ struct RawEventStreamQueryParams {
     reverse: bool,
     #[serde(default)]
     live: bool,
-    path: Option<String>,
+    #[serde(default)]
+    paths: Vec<String>,
```
MCarlomagno, line 95: now we accept multiple path params @andrei-21 @86667  this might be a breaking change for clients, should we keep it as `path`?

pubky-homeserver/src/persistence/files/events/path_filter.rs
```diff
@@ -0,0 +1,91 @@
...
+/// matches that directory and all of its descendants.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PathFilter {
+    /// Exact-match filter for a file path (no trailing `/`).
+    File(String),
+    /// Prefix filter for a directory path (trailing `/`) and its descendants.
+    Dir(String),
+}
+
+impl PathFilter {
```
MCarlomagno: New `PathFilter` type, file/dir matching for the live mode + `to_condition` (`LIKE` for the SQL query filter)

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -489,13 +503,67 @@ async fn resolve_user_cursors(
...
+        }
+    }
+
+    // Tenant to authorize against. For a private subscription this is the only
+    // user (enforced above) and the predicate further requires it to equal the
+    // session user.
+    let tenant = &user_cursors
+        .first()
+        .ok_or_else(|| HttpError::bad_request("user parameter is required"))?
+        .0;
```
andrei-21: It is confusing to take the first user (of many) and then use it in `has_read_permission()` ( why not the second or the last?). Maybe if we make `has_read_permission()` to accept `Option<tenat>` and pass it if `user_cursors` has exactly one item in it.
Then the function becomes:
```rust
    // No path requested: default to public-only.
    if paths.is_empty() {
        return Ok(vec![PathFilter::Dir(PUBLIC_ROOT.to_string())]);
    }

    let tenant = match user_cursors {
        [(tenant, _)] => Some(tenant),
        _ => None,
    };

    let mut filters = Vec::with_capacity(paths.len());
    for path in paths {
        has_read_permission(session, tenant, path)?;
        filters.push(PathFilter::from_path(path));
    }
    Ok(filters)
```

pubky-homeserver/src/persistence/files/events/path_filter.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Build a filter from a request path.
+    pub fn from_path(path: &WebDavPath) -> Self {
```
andrei-21: You can implement `From<WebDavPath> for PathFilter` and use `path.into()` where you need.

pubky-homeserver/src/persistence/files/events/path_filter.rs
```diff
@@ -0,0 +1,91 @@
...
+///
+/// A file filter matches only its exact path, while a directory filter (trailing `/`)
+/// matches that directory and all of its descendants.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PathFilter {
+    /// Exact-match filter for a file path (no trailing `/`).
+    File(String),
+    /// Prefix filter for a directory path (trailing `/`) and its descendants.
+    Dir(String),
+}
```
andrei-21: I would just wrap `WebDavPath` like:

```suggestion
pub struct PathFilter(WebDavPath);
```

andrei-21 (changes requested):

pubky-homeserver/src/client_server/routes/events.rs
andrei-21: I guess we are going to have more serious breaking changes because of the new auth, I think it is fine to add this.
@86667 what will you say?

pubky-homeserver/src/client_server/routes/events.rs
MCarlomagno: Yes agree, I think this is functionally correct, because `/priv` reads only accept one user, but agree with the refactor

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -489,13 +503,62 @@ async fn resolve_user_cursors(
...
+    let has_private = paths.iter().any(|p| p.as_str().starts_with(PRIVATE_ROOT));
+
+    // A private path needs a session. Surface that as a 401 up front so an
+    // anonymous private request fails the same way regardless of how the
+    // requested paths are ordered.
+    if has_private && session.is_none() {
+        return Err(HttpError::unauthorized_with_message(
+            "Authentication required to subscribe to private events",
+        ));
+    }
```
andrei-21: This check is redundant, because we validate it in `has_read_permission()`.

pubky-homeserver/src/persistence/files/events/events_repository.rs
```diff
@@ -203,20 +205,17 @@ impl EventRepository {
...
-                    .replace('%', "\\%");
-                let like_pattern = format!("{}%", escaped_prefix);
-                statement = statement
-                    .and_where(
-                        Expr::col((EVENT_TABLE, EventIden::Path))
-                            .like(LikeExpr::new(like_pattern).escape('\\')),
-                    )
-                    .to_owned();
+            // Restrict to the authorized path filters, ORed together so an event
+            // matching any one of them is returned. The list is non-empty in
```
andrei-21: I was confused with `ORed`, maybe something like:
Restrict results to events whose path matches **at least one** authorized path filter.

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -489,13 +503,62 @@ async fn resolve_user_cursors(
...
+}
+
+/// Filter events in live mode based on user IDs, cursors, and the authorized
+/// path filters.
 fn should_include_live_event(
     event: &EventEntity,
     user_ids: &[i32],
     user_cursor_map: &HashMap<i32, Option<EventCursor>>,
-    path_filter: Option<&WebDavPath>,
+    path_filters: &[PathFilter],
```
andrei-21: `path_filters` is not really a filter, but allow list or something.
I would also rename other instances similar, then `authorized_path_filters()` returning default `/pub/` will become more intuitive and natural.

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -318,10 +324,17 @@ pub async fn feed(
 /// ```
 pub async fn feed_stream(
     State(state): State<AppState>,
+    session: Option<AuthSession>,
```
andrei-21, line 327: AI claims that `session` might be empty for cookie-based auth because current cookie middleware does not look at `user` param.
@SeverinAlexB, @86667 is it ok if we deprecate cookie-based auth?

SeverinAlexB: No. Cookie based auth needs to work for at least another 6 to 12 months.
If it is a bug then it needs to be fixed.

pubky-homeserver/src/client_server/routes/events.rs
86667: I dont think this needs to be a break - it can work the same as `user`. Change `path: Option<String>` to be `path: Vec<String>` and push into an array each `path` param.

andrei-21 (approved):

andrei-21 (approved):

PR #463, opened 2026-06-26
feat(homeserver): events admin endpoint

### LINKED ISSUE #427: Admin Events Endpoint
Add an admin-only event stream for operational tooling, backup, debugging, and maintenance. It is mounted on the existing admin HTTP server.

The full requirements doc: https://docs.google.com/document/d/1XYL7_pFYtdPJC_rKL7K5iFe2ZH2mEnTUf_oKGY_nNkE/edit?tab=t.eo542lh3h2q.

## Dependencies

Part of epic #426.

Blocked by _(soft — can be developed in parallel)_:
- #432 — Accept `/priv/` for writes (needs private events to exist to be worth streaming)

MCarlomagno:
## Summary

Closes [#427](https://github.com/pubky/pubky-core/issues/427)

Adds an admin-server counterpart to the client `GET /events-stream` SSE endpoint for operational tooling, moderation and debugging. Gated by the admin password (`X-Admin-Password`), it streams **all** events — public **and** private (`/priv/...`) — over Server-Sent Events.

It mirrors the client `/events-stream` contract (`limit`, `reverse`, `live`, repeatable `path`), with two admin differences: `user` is an *optional* filter (omit to stream every user) and a single global `cursor` paginates the combined feed. The streaming logic is shared via `EventsService::all_events_stream`. Because it surfaces private paths, the response is `Cache-Control: no-store`.

## Acceptance criteria
- [x] `GET /events-stream` with the admin password streams all events, including `/priv/...`, in batch and live mode.
- [x] `user` is optional (omit = all users); repeatable `path` filters union (file-vs-directory matching).
- [x] A single global `cursor` resumes the feed; `reverse` returns newest-first (batch only).
- [x] Missing or wrong admin password → `401`.
- [x] Malformed cursor → `400`.
- [x] Response is `Cache-Control: no-store`.
- [x] The public/client `GET /events-stream` stays unchanged (no regression).

### DIFF
--- a/pubky-homeserver/openapi.yml
+++ b/pubky-homeserver/openapi.yml
@@ -676,7 +676,7 @@ paths:
           description: Maximum total events to send.
           schema:
             type: integer
-            minimum: 0
+            minimum: 1
             maximum: 65535
         - name: reverse
           in: query
--- a/pubky-homeserver/openapi.yml
+++ b/pubky-homeserver/openapi.yml
@@ -626,6 +626,13 @@ paths:
         data: cursor: 42
         data: content_hash: r0NJufX5oaagQE3qNtzJSZvLJcmtwRK3zJqTyuQfMmI=
         ```
+
+        **Admin server variant.** The admin server (`http://localhost:6288`) exposes the same
+        `GET /events-stream` SSE protocol, but authenticated with the `X-Admin-Password` header
+        instead of a session. It streams **all** events, including private (`/priv/...`) paths, and
+        responds with `Cache-Control: no-store`. There, `user` is an *optional* filter (omit it to
+        stream every user's events) and a single global `cursor` paginates the combined feed;
+        `limit`, `live`, `reverse`, and `path` behave as below.
       operationId: getEventStream
       parameters:
         - name: user
@@ -702,53 +709,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/AdminInfoResponse"
-        "401":
-          description: Missing or invalid admin password
-
-  /events:
-    get:
-      tags: [Admin]
-      summary: All events feed (plain text, includes private paths)
-      description: |
-        Same plain-text, cursor-paginated format as the public `GET /events/` feed, but exposes
-        all events including private (`/priv/...`) paths. Response is `Cache-Control: no-store`.
-      operationId: getAdminEventFeed
-      servers:
-        - url: http://localhost:6288
-      security:
-        - adminPassword: []
-      parameters:
-        - name: cursor
-          in: query
-          description: Starting cursor position. Defaults to `"0"` (beginning) when omitted.
-          schema:
-            type: string
-        - name: limit
-          in: query
-          description: Maximum number of events to return.
-          schema:
-            type: integer
-            minimum: 1
-            maximum: 65535
-      responses:
-        "200":
-          description: Event feed (public and private events)
-          headers:
-            Cache-Control:
-              schema:
-                type: string
-                example: no-store
-          content:
-            text/plain:
-              schema:
-                type: string
-              example: |
-                PUT pubky://o1gg96ewuo.../pub/example.txt
-                PUT pubky://o1gg96ewuo.../priv/app/secret.txt
-                DEL pubky://o1gg96ewuo.../pub/old.txt
-                cursor: 12345
-        "400":
-          description: Invalid cursor
         "401":
           description: Missing or invalid admin password
 
--- a/pubky-homeserver/openapi.yml
+++ b/pubky-homeserver/openapi.yml
@@ -664,6 +664,13 @@ paths:
               value: [o1gg96ewuojmopcjbz8895478wdtxtzzuxnfjjz8o8e77csa1ngo]
             withCursor:
               value: ["o1gg96ewuojmopcjbz8895478wdtxtzzuxnfjjz8o8e77csa1ngo:42"]
+        - name: cursor
+          in: query
+          description: |
+            Resume from a single global cursor (admin server). On the client server, append
+            `:cursor` to each `user` value instead.
+          schema:
+            type: string
         - name: limit
           in: query
           description: Maximum total events to send.
--- a/pubky-homeserver/openapi.yml
+++ b/pubky-homeserver/openapi.yml
@@ -627,24 +627,32 @@ paths:
         data: content_hash: r0NJufX5oaagQE3qNtzJSZvLJcmtwRK3zJqTyuQfMmI=
         ```
 
-        **Admin server variant.** The admin server (`http://localhost:6288`) exposes the same
-        `GET /events-stream` SSE protocol, but authenticated with the `X-Admin-Password` header
-        instead of a session. It streams **all** events, including private (`/priv/...`) paths, and
-        responds with `Cache-Control: no-store`. There, `user` is an *optional* filter (omit it to
-        stream every user's events) and a single global `cursor` paginates the combined feed;
-        `limit`, `live`, `reverse`, and `path` behave as below.
+        This path is served by **two servers with different contracts** (select the server below):
+        - **Client server** (`http://localhost:6287`): session-authenticated. `user` is **required**;
+          `/priv/...` paths require an authorized session capability (otherwise `401`/`403`).
+        - **Admin server** (`http://localhost:6288`): authenticated with the `X-Admin-Password`
+          header. Streams **all** events including private (`/priv/...`) paths and responds with
+          `Cache-Control: no-store`. `user` is **optional** (omit to stream every user); a single
+          global `cursor` paginates the combined feed.
       operationId: getEventStream
+      servers:
+        - url: http://localhost:6287
+          description: Client server (session auth)
+        - url: http://localhost:6288
+          description: Admin server (X-Admin-Password)
       security:
         - {}
         - bearerAuth: []
         - cookieAuth: []
+        - adminPassword: []
       parameters:
         - name: user
           in: query
-          required: true
+          required: false
           description: |
             One or more user public keys (z-base-32). Repeat for multiple users.
             Optionally append `:cursor` to start from a specific position.
+            **Required** on the client server; **optional** on the admin server (omit to stream all users).
           schema:
             type: array
             items:
@@ -677,7 +685,11 @@ paths:
             default: false
         - name: path
           in: query
-          description: Path filter. /priv/ paths require an authorized session.
+          description: |
+            Path filter (repeatable; union — an event matches if it satisfies any). A trailing slash
+            matches a directory and its descendants; otherwise it matches that exact file. On the
+            client server, `/priv/...` paths require an authorized session capability; the admin
+            server returns all paths.
           schema:
             type: array
             items:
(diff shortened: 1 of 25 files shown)


pubky-homeserver/src/admin_server/routes/admin_events.rs
```diff
@@ -0,0 +1,63 @@
...
+/// PUT pubky://user_pubkey/priv/app/secret.txt
+/// DEL pubky://user_pubkey/pub/old.txt
+/// cursor: 12345
+/// ```
+pub async fn feed(
+    State(state): State<AppState>,
+    params: ListQueryParams,
+) -> HttpResult<impl IntoResponse> {
+    // Treat a missing or empty cursor as "from the beginning", matching the public feed.
+    let cursor_str = params.cursor.unwrap_or_else(|| "0".to_string());
```
andrei-21: `get_all_by_cursor()` takes optional cursor, so you do not need to default here, just use `None`.

pubky-homeserver/openapi.yml
```diff
@@ -705,6 +705,53 @@ paths:
...
+          description: Starting cursor position. Defaults to `"0"` (beginning) when omitted.
+          schema:
+            type: string
+        - name: limit
+          in: query
+          description: Maximum number of events to return.
+          schema:
+            type: integer
+            minimum: 1
+            maximum: 65535
```
andrei-21: As per https://github.com/pubky/pubky-core/blob/feat%2Fadmin-events-endpoint/pubky-homeserver/src/constants.rs#L3:
```suggestion
            maximum: 1000
```

pubky-homeserver/openapi.yml
```diff
@@ -705,6 +705,53 @@ paths:
...
+        all events including private (`/priv/...`) paths. Response is `Cache-Control: no-store`.
+      operationId: getAdminEventFeed
+      servers:
+        - url: http://localhost:6288
+      security:
+        - adminPassword: []
+      parameters:
+        - name: cursor
+          in: query
+          description: Starting cursor position. Defaults to `"0"` (beginning) when omitted.
```
andrei-21: We should not limit ourselves to the underlying structure of the cursor:
```suggestion
          description: Starting cursor position. Defaults to the beginning when omitted.
```

pubky-homeserver/src/admin_server/routes/admin_events.rs
```diff
@@ -0,0 +1,63 @@
...
+use super::super::app_state::AppState;
+use crate::{
+    client_server::{query_params::ListQueryParams, routes::events::format_events_feed},
+    shared::{HttpError, HttpResult},
+};
+
+/// Admin-only historical events feed returning **all** events (public and private).
+///
+/// ## Query Parameters
+/// - `cursor` (optional): Starting cursor position. Default: `"0"` (beginning).
```
andrei-21: Again, no need to be specific about the cursor structure:
```suggestion
/// - `cursor` (optional): Starting cursor position. Default to the beginning.
```

pubky-homeserver/src/admin_server/routes/admin_events.rs
```diff
@@ -0,0 +1,63 @@
+//! Admin-only historical events feed.
+//!
+//! it exposes all events — public (`/pub/...`) and private (`/priv/...`) alike — for operational tooling,
+//! debugging, and maintenance.
+//!
+//! The public feed (/events) is public-only and never leaks private paths, this endpoint is the one
+//! place private paths are surfaced over the events feed, hence the admin auth requirement
+//! and the `Cache-Control: no-store` header.
```
andrei-21: Something is off with this comment. Should it mention `/events/`?

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -216,13 +216,31 @@ impl TryFrom<RawEventStreamQueryParams> for EventStreamQueryParams {
...
     // reconstruction here.
     format!("pubky://{}{}", entity.user_pubkey.z32(), entity.path.path())
 }
 
+/// Render a batch of events as the plain-text feed body used by `GET /events/`.
+///
+/// One line per event (`<TYPE> pubky://<user>/<path>`), followed by a trailing
+/// `cursor: <id>` line pointing at the last event so the caller can resume.
+/// Returns an empty string when there are no events (and therefore no cursor line).
+pub(crate) fn format_events_feed(events: &[EventEntity]) -> String {
```
86667: This doesnt belong here anymore since its used by both client_server and admin_server modules. Perhaps it could live with `EventEntity` itself? Or move to `/shared`

pubky-homeserver/src/client_server/mod.rs
```diff
@@ -8,7 +8,7 @@ mod app;
 pub(crate) mod app_state;
 pub(crate) mod auth;
 mod middleware;
-mod query_params;
+pub(crate) mod query_params;
```
86667: Same here, move into shared please as this is now depended on by the admin_server module too.

We're aiming at moving this code base to a module layering architecture. Its a little messy now but were on our way to cleaning it up. If this means nothing to you then have a read online and ping me for more context - it took me a few chats with Sev before I fully understood the vision!

pubky-homeserver/src/admin_server/routes/admin_events.rs
```diff
@@ -0,0 +1,63 @@
...
+/// ```text
+/// PUT pubky://user_pubkey/pub/example.txt
+/// PUT pubky://user_pubkey/priv/app/secret.txt
+/// DEL pubky://user_pubkey/pub/old.txt
+/// cursor: 12345
+/// ```
+pub async fn feed(
+    State(state): State<AppState>,
+    params: ListQueryParams,
+) -> HttpResult<impl IntoResponse> {
```
86667, line 184: This logic looks as though it lives in the `EventsService`, especially since its almost identical to the `feed()` fn in client_server.

86667 (commented):
Nice, just a few bits to move around.

I am wondering whether the admin `/events` should follow the approach taken by `/events-stream` rather than copying the deprecated existing `/events` endpoint. This would make it more flexible, support SSE, plus mean that Events returned are in the same format as `/events-stream`. The existing `/events` is deprecated, afterall.

I also think in this case that the naming should match `/events-stream`, but with an option for pulling all user Events for the admin version.

The reason we still have `/events` would then be only for Nexus (since moderation would use this new admin one). 

For context, were still deciding what the best route is for nexus - whether to always fetch events in batches of users via `/events-stream` or whether they do truly need a "fetch all user events" functionality - its not clear if the get-events-by-user-batches pattern is sufficient due to the number of requests they would need to make. Either they do, and we'll add an `/events-stream-all` type endpoint, or they dont and we can just remove `/events`.

andrei-21 (changes requested):

pubky-homeserver/src/admin_server/routes/admin_events.rs
MCarlomagno: Agree, moved the streaming into `EventsService::all_events_stream(...)`

pubky-homeserver/src/client_server/routes/events.rs
MCarlomagno: Went with `EventEntity` and `ConnectionGuard` moved next to `Metrics`

pubky-homeserver/openapi.yml
```diff
@@ -649,12 +664,19 @@ paths:
...
+            `:cursor` to each `user` value instead.
+          schema:
+            type: string
         - name: limit
           in: query
           description: Maximum total events to send.
           schema:
             type: integer
-            minimum: 1
+            minimum: 0
```
andrei-21: Why changing to `0`? `0` valued does not make much sense. We probably should require minimum `1` and validate the limit in the API.

pubky-homeserver/openapi.yml
```diff
@@ -626,18 +626,33 @@ paths:
...
+      servers:
+        - url: http://localhost:6287
+          description: Client server (session auth)
+        - url: http://localhost:6288
+          description: Admin server (X-Admin-Password)
       security:
         - {}
         - bearerAuth: []
         - cookieAuth: []
+        - adminPassword: []
```
andrei-21, line 647: We are hitting the limits of one file. The security is not optional and supports only admin password for admin endpoint.
I created an issue to address it https://github.com/pubky/pubky-core/issues/476.

pubky-homeserver/openapi.yml
MCarlomagno: yeah I wasnt sure about that, like 0 is not useful but it's harmless, so allowing it would save us the extra validation in the API, but I can revert that

MCarlomagno: fixed

pubky-homeserver/openapi.yml
MCarlomagno: so do you want me to do some change here? I would tackle this in a different PR

andrei-21: Sure we should address it in a different PR.

pubky-homeserver/src/admin_server/routes/admin_events.rs
```diff
@@ -0,0 +1,204 @@
...
+    users: Vec<PublicKey>,
+    /// Optional starting cursor (global). None = from the beginning.
+    cursor: Option<String>,
+    /// Maximum total events to send before closing.
+    limit: Option<u16>,
+    /// Live streaming mode (cannot be combined with `reverse`).
+    live: bool,
+    /// Reverse (newest-first) ordering; batch-only.
+    reverse: bool,
+    /// Path filters (repeatable; union — an event matches if it satisfies any). A trailing slash
```
andrei-21: Minor: if some options are incompatible, then an enum can be used like:
```rust
enum Mode {
    Forward,
    ForwardLive,
    Reverse,
}
```

andrei-21 (dismissed):

pubky-homeserver/src/persistence/files/events/events_service.rs
```diff
@@ -141,6 +195,137 @@ impl EventsService {
...
+                if caught_up {
+                    if !live {
+                        return;
+                    }
+                    break;
+                }
+            }
+
+            // Phase 2: live mode (only `Mode::ForwardLive` reaches here).
+            if live {
```
andrei-21, line 283: One of `if live` condition is useless.
The comment also looks weird, it states that only live mode can reach, but then we test again.

pubky-homeserver/src/persistence/files/events/events_service.rs
```diff
@@ -141,6 +195,137 @@ impl EventsService {
...
+        let half_capacity = self.channel_capacity() / 2;
+
+        async_stream::stream! {
+            // Cleanup + connection metrics on any exit path.
+            let _guard = ConnectionGuard::new(metrics.clone());
+
+            let mut last_cursor = filter.start_cursor;
+            let mut total_sent: usize = 0;
+            let reverse = filter.mode == Mode::Reverse;
+            let live = filter.mode == Mode::ForwardLive;
```
andrei-21: Maybe it can be encapsulated into `Mode` like: `mode.live()` and `mode.reverse()`.

pubky-homeserver/src/persistence/files/events/events_service.rs
MCarlomagno: >One of if live condition is useless.

If I'm following corretly, I think is not, if there is an error in `get_all_events` you might still want to stream live events only when `live=true`

andrei-21 (approved):

PR #466, opened 2026-06-29
fix: docker build pipeline

SpontaneousOverthrow:
Updated secret name using github ui for username(we have the same naming convention for all our repos),  added security check for tag.

In the future I will make this pipeline a part of release pipeline.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,18 +24,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Enforce Tag is on Main Branch
-        run: |
-          # Get the exact commit hash of the tag
-          TAG_COMMIT=$(git rev-parse HEAD)
-
-          # Check if that commit exists in the remote main branch history
-          if ! git merge-base --is-ancestor "$TAG_COMMIT" origin/main; then
-            echo "❌ SECURITY ERROR: Tag does not belong to the main branch history."
-            exit 1
-          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -25,6 +24,18 @@ jobs:
...
+        run: |
+          # Get the exact commit hash of the tag
+          TAG_COMMIT=$(git rev-parse HEAD)
+
+          # Check if that commit exists in the remote main branch history
+          if ! git merge-base --is-ancestor "$TAG_COMMIT" origin/main; then
+            echo "❌ SECURITY ERROR: Tag does not belong to the main branch history."
+            exit 1
+          fi
+          echo "✅ Security check passed. Tag is a valid ancestor of main."
```
SeverinAlexB: It doesn't really need to be on the main branch. The alpha version I published today was on a side branch too. Hot fixes might be on a side branch as well.

Or what is the intend of this check?

SeverinAlexB (changes requested):
Great. One thing, otherwise good to go

.github/workflows/docker.yml
SpontaneousOverthrow: yes, the main thing was to make sure that the release is built from main branch

SpontaneousOverthrow: ok let me remove it

SpontaneousOverthrow: done

SeverinAlexB (approved):

PR #467, opened 2026-06-29
feat: Add init cli command for generating config.toml and secrets

86667:
Create config.toml without having to run Homeserver. This will be replaced by the [Homeserver CLI](https://github.com/pubky/pubky-core/pull/451) work but is useful now for the [new docs](https://github.com/pubky/pubky-core/pull/431).

### DIFF
--- a/pubky-homeserver/src/main.rs
+++ b/pubky-homeserver/src/main.rs
@@ -46,10 +46,8 @@ async fn main() -> Result<()> {
             let data_dir = PersistentDataDir::new(args.data_dir);
             data_dir.init()?;
             println!(
-                "Data directory initialized at {}.\n\
-                 Edit {} to configure PostgreSQL and other settings, then run `pubky-homeserver` to start.",
-                data_dir.path().display(),
-                data_dir.get_config_file_path().display(),
+                "Data directory initialized at {}.",
+                data_dir.path().display()
             );
         }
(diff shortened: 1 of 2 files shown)


pubky-homeserver/src/main.rs
```diff
@@ -21,49 +23,73 @@ fn validate_config_dir_path(path: &str) -> Result<PathBuf, String> {
...
 
-    tracing::info!("Use data directory: {}", args.data_dir.display());
-    let server = HomeserverApp::start_with_persistent_data_dir_path(args.data_dir).await?;
+    match args.command {
+        Some(Command::Init) => {
+            let data_dir = PersistentDataDir::new(args.data_dir);
+            data_dir.init()?;
+            println!(
+                "Data directory initialized at {}.\n\
+                 Edit {} to configure PostgreSQL and other settings, then run `pubky-homeserver` to start.",
```
andrei-21: If I run `$ cargo r -- --data-dir /tmp/pubky init` it prints me:
> [...] run `pubky-homeserver` to start.

But it should print: run `pubky-homeserver --data-dir /tmp/pubky` to start.

86667: removed that line of the message

andrei-21 (approved):

PR #470, opened 2026-06-30
feat(sdk): subscribe to private events

### LINKED ISSUE #438: SDK: subscribe to private events

MCarlomagno:
Closes #438.

### Summary

Extends the SDK's `EventStreamBuilder` so apps can subscribe to a user's private
(`/priv/...`) events, not just public ones. The homeserver side already landed in
#429 (repeatable `path=` params + per-path authorization), this PR is the SDK/bindings
half that lets callers actually drive it.

Two builder additions:
- **`.path()` is now repeatable** — call it once per scope to receive the union
  (e.g. `/pub/` plus a private `/priv/app/`). Previously it was a single overwrite-on-recall `Option<String>`.
- **`.session(&PubkySession)`** — attaches the session credential so the homeserver can authorize each private path against the session's read capabilities.

### API

```rust
let mut stream = pubky
    .event_stream_for_user(&user, None)
    .session(&session)        // new: authenticate the subscription
    .path("/pub/")            // repeatable
    .path("/priv/app/")       // private scope, authorized server-side
    .live()
    .subscribe()
    .await?;
```

### Auth behavior
- Public-only streams stay anonymous, even when a session is provided.
- Private streams attach a session credential only for a single-user subscription where the session belongs to that user.
- Session credentials are only attached to their bound target homeserver, so the SDK fails early instead of sending cookies or bearers to the wrong homeserver.
- On the homeserver, a valid bearer authenticates normally. If an `Authorization: Bearer ...` header is present, the event-stream cookie fallback is disabled, even when the bearer is invalid.
- Cookie auth is still supported for single-tenant private streams by resolving the cookie for the requested user= tenant.

### Acceptance criteria
- [x] A subscription with a `/priv/app/` path + session receives that user's private events.
- [x] The same subscription **without** a covering cap receives none (server-enforced).
- [x] Public subscriptions (no path / `/pub/`) behave as before.

### DIFF
--- a/e2e/src/tests/events/stream_private.rs
+++ b/e2e/src/tests/events/stream_private.rs
@@ -375,3 +375,121 @@ async fn events_stream_live_authorized_receives_private() {
         "authorized live subscriber should receive the in-scope private event"
     );
 }
+
+/// Sign up a fresh cookie user; return its pubkey, the (kept-alive) session, and
+/// the raw `name=value` Cookie header for its session secret.
+async fn cookie_user(
+    testnet: &pubky_testnet::EphemeralTestnet,
+) -> (
+    pubky_testnet::pubky::PublicKey,
+    pubky_testnet::pubky::PubkySession,
+    String,
+) {
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+    let signer = pubky.signer(Keypair::random());
+    let session = signer
+        .signup_cookie(&server.public_key(), None)
+        .await
+        .unwrap();
+    // `export_secret()` returns `<pubkey_z32>:<secret>`; the wire cookie is
+    // `<pubkey_z32>=<secret>`.
+    let token = session.as_cookie().unwrap().export_secret().unwrap();
+    let (name, value) = token.split_once(':').unwrap();
+    (signer.public_key(), session, format!("{name}={value}"))
+}
+
+/// Raw `/events-stream?user=A&path=/priv/app/` with A's session cookie is
+/// authenticated (200): the server resolves the cookie by the single `user=`
+/// tenant even though the endpoint is homeserver-addressed.
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_cookie_same_tenant_is_authorized() {
+    let testnet = build_full_testnet().await;
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+    let server_host = server.public_key().z32();
+
+    let (a, session, cookie) = cookie_user(&testnet).await;
+    session
+        .storage()
+        .put("/priv/app/secret.txt", vec![1])
+        .await
+        .unwrap();
+
+    let url = format!(
+        "https://{}/events-stream?user={}&path=/priv/app/&limit=1",
+        server_host,
+        a.z32()
+    );
+    let response = pubky
+        .client()
+        .request(Method::GET, &url)
+        .header("Cookie", cookie)
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(response.status(), StatusCode::OK);
+}
+
+/// A valid cookie value for A, presented under user B's cookie name for
+/// `user=B`, does not authenticate: the session belongs to A, not B → 401.
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_cookie_cross_user_secret_is_unauthenticated() {
+    let testnet = build_full_testnet().await;
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+    let server_host = server.public_key().z32();
+
+    let (_a, _session, a_cookie) = cookie_user(&testnet).await;
+    let a_secret = a_cookie.split_once('=').unwrap().1.to_string();
+    let b = Keypair::random().public_key();
+
+    let url = format!(
+        "https://{}/events-stream?user={}&path=/priv/app/",
+        server_host,
+        b.z32()
+    );
+    let response = pubky
+        .client()
+        .request(Method::GET, &url)
+        .header("Cookie", format!("{}={}", b.z32(), a_secret))
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+}
+
+/// A valid cookie plus an invalid `Authorization: Bearer` is rejected: a
+/// presented bearer disables the cookie fallback (conservative precedence).
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_cookie_not_used_when_bearer_present() {
+    let testnet = build_full_testnet().await;
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+    let server_host = server.public_key().z32();
+
+    let (a, session, cookie) = cookie_user(&testnet).await;
+    session
+        .storage()
+        .put("/priv/app/secret.txt", vec![1])
+        .await
+        .unwrap();
+
+    let url = format!(
+        "https://{}/events-stream?user={}&path=/priv/app/",
+        server_host,
+        a.z32()
+    );
+    let response = pubky
+        .client()
+        .request(Method::GET, &url)
+        .header("Cookie", cookie)
+        .header("Authorization", "Bearer not-a-real-token")
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+}
--- a/e2e/src/tests/events/stream_sdk.rs
+++ b/e2e/src/tests/events/stream_sdk.rs
@@ -2,7 +2,7 @@ use super::*;
 use futures::StreamExt;
 use pubky_testnet::pubky::errors::{Error, RequestError};
 use pubky_testnet::pubky::{ClientId, EventCursor, PubkySession, PublicKey};
-use tokio::time::{Duration, timeout};
+use tokio::time::{timeout, Duration};
 
 /// Sign up a fresh user and return its public key plus an authenticated
 /// (root-capability) grant session.
--- a/e2e/src/tests/events/stream_sdk.rs
+++ b/e2e/src/tests/events/stream_sdk.rs
@@ -1,4 +1,32 @@
 use super::*;
+use pubky_testnet::pubky::errors::{Error, RequestError};
+use pubky_testnet::pubky::{ClientId, PubkySession, PublicKey};
+
+/// Sign up a fresh user and return its public key plus an authenticated
+/// (root-capability) grant session.
+async fn signed_in_user(
+    testnet: &pubky_testnet::EphemeralTestnet,
+    client_id: &str,
+) -> (PublicKey, PubkySession) {
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+    let signer = pubky.signer(Keypair::random());
+    signer.signup(&server.public_key(), None).await.unwrap();
+    let session = signer
+        .signin(ClientId::new(client_id).unwrap())
+        .await
+        .unwrap();
+    (signer.public_key(), session)
+}
+
+/// Extract the HTTP status from a typed server-rejection error, so tests assert
+/// against `RequestError::Server { status, .. }` rather than a raw string.
+fn server_status(err: &Error) -> Option<StatusCode> {
+    match err {
+        Error::Request(RequestError::Server { status, .. }) => Some(*status),
+        _ => None,
+    }
+}
 
 /// Test the SDK builder API: `event_stream_for()` and `add_users()`
 /// This tests the high-level SDK interface rather than raw HTTP requests.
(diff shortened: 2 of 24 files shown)


pubky-sdk/src/actors/event_stream.rs
```diff
@@ -283,12 +319,35 @@ impl EventStreamBuilder {
...
+    /// (cookie or bearer token) is attached to the request so the homeserver can
+    /// authorize each private [`path`](Self::path) against the session's read
+    /// capabilities. Public subscriptions don't need this.
+    ///
+    /// The homeserver only authorizes private paths when the subscription names
+    /// exactly the session's own user (a single `user=`); otherwise it responds
+    /// with `403 Forbidden`.
+    #[must_use]
+    pub fn session(mut self, session: &PubkySession) -> Self {
+        self.credential = Some(Arc::clone(session.credential()));
```
andrei-21, line 350: We probably should validate the session owner before attaching credentials here. We fail early and do not reveal sessions to wrong homeservers by mistake.

MCarlomagno: Sounds good, just one tradeoff about it, the session is homeserver-scoped, not user-scoped, so to validate the owner here we would need an extra pkarr call to resolve the owner homeserver and compare it against the target. It's doable, but just flagging the tradeooff

MCarlomagno: just commited that change, let me know if that makes sense

pubky-sdk/bindings/js/src/actors/event_stream.rs
```diff
@@ -150,18 +160,39 @@ impl EventStreamBuilder {
         EventStreamBuilder(self.0.reverse())
     }
 
-    /// Filter events by path prefix.
+    /// Filter events by path. Call once per path to receive the
```
86667, line 163: Nit: Its still the prefix isnt it? Or did that change?

pubky-sdk/src/actors/event_stream.rs
```diff
@@ -344,19 +416,33 @@ impl EventStreamBuilder {
...
+            .credential
+            .as_ref()
+            .map(|credential| credential.info().public_key().clone());
+
+        // Resolve the homeserver we will send the request to.
+        //
+        // When a session credential is attached it must ONLY ever be transmitted
+        // to its owner's homeserver. We resolve the owner's homeserver and, if the
+        // caller also pinned one explicitly, require the two to match.
+        let homeserver = if let Some(owner) = &session_owner {
```
86667: I dont think this is necessary. It limits the usage of the instance and what is the benefit? slightly smaller payload?

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +327,210 @@ async fn events_stream_sdk_builder_api() {
...
+        .add_users([(&user, None)])
+        .unwrap()
+        .path("/priv/app/")
+        .subscribe()
+        .await;
+
+    let err = match result {
+        Ok(_) => panic!("private path without a session must be rejected"),
+        Err(e) => e,
+    };
```
86667:
```suggestion
    let err = result.expect_err("private path without a session must be rejected");

```

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +327,210 @@ async fn events_stream_sdk_builder_api() {
...
+        event.resource.path
+    );
+}
+
+/// A mixed `/pub/` + `/priv/app/` subscription returns the union
+/// and never an unrequested private scope.
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_private_union_excludes_unrequested() {
+    use futures::StreamExt;
```
86667: nit: move to top

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +327,210 @@ async fn events_stream_sdk_builder_api() {
         err
     );
 }
+
+/// An authenticated owner subscribing via the SDK builder with
+/// `.session()` + a `/priv/` path receives that user's private events, scoped
+/// to the requested filter.
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_private_authorized_receives() {
```
86667: this test and the one below seem very similar

86667 (commented):
Nice looks good. A couple comments.

pubky-sdk/src/actors/event_stream.rs
86667: Lol just seen above comment - https://github.com/pubky/pubky-core/pull/470#discussion_r3512335886

andrei-21: We should do it a bit differently, otherwise we risk overriding the requested homeserver:
```rust
pubky.event_stream_for_user(&other_user, None)
    .session(&my_session)
    .path("/pub/")
    .subscribe()
```
This should still target `other_user`'s homeserver, not `my_session` owner's homeserver.

So first resolve the intended target:
```rust
let homeserver = if let Some(hs) = &self.homeserver {
    hs.clone()
} else {
    self.resolve_homeserver(&self.users[0].0).await?
};
```
Then decide whether the session credential can be attached.

pubky-sdk/bindings/js/src/actors/event_stream.rs
MCarlomagno: Yes, but not in all cases, a directory path is a prefix match, but a file path is now exact, so "prefix" alone is wrong, which is why I changed the wording, does that make sense?

pubky-sdk/src/actors/event_stream.rs
MCarlomagno: done, the target now comes from the subscription (explicit or first user), not the session

e2e/src/tests/events/stream_sdk.rs
MCarlomagno: merged

e2e/src/tests/events/stream_sdk.rs
MCarlomagno: I did it a bit differently, but following the same idea

```rust
    let err = result
        .err() // we need to extract the error first
        .expect("must refuse to send the session credential to a foreign homeserver");
```

pubky-sdk/bindings/js/src/actors/event_stream.rs
86667: Makes sense thanks. I didnt realise this had changed

pubky-sdk/src/actors/event_stream.rs
andrei-21: For explicit homeserver streams, skip PKARR validation when attaching a session. If the caller used `event_stream_for(&homeserver).session(&session)`, treat the explicit homeserver as intentional and attach the credential directly. Keep the PKARR-based check only for implicit `event_stream_for_user(...)` streams, where resolution is already part of selecting the target homeserver.

pubky-sdk/src/actors/event_stream.rs
```diff
@@ -366,19 +445,34 @@ impl EventStreamBuilder {
...
 
+        let credential = self
+            .credential
+            .as_ref()
+            .filter(|credential| self.should_attach_credential(credential.as_ref()));
+        let can_attach_credential = match credential {
+            Some(credential) => credential.can_attach_to(&homeserver).await,
+            None => false,
+        };
+        if credential.is_some() && !can_attach_credential {
```
andrei-21, line 456: This condition will not catch when `.session()` is provided for a private path but the target is the wrong user or a multi-user. I think we should fail the request instead of ignoring the session.

e2e/src/tests/events/stream_private.rs
```diff
@@ -375,3 +375,121 @@ async fn events_stream_live_authorized_receives_private() {
...
+        .request(Method::GET, &url)
+        .header("Cookie", format!("{}={}", b.z32(), a_secret))
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+}
+
+/// A valid cookie plus an invalid `Authorization: Bearer` is rejected: a
+/// presented bearer disables the cookie fallback (conservative precedence).
```
MCarlomagno, line 465: it avoids surprising behavior where a bad bearer accidentally still succeeds due to an ambient browser cookie

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
         err
     );
 }
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_private_authorized_scoping() {
```
MCarlomagno, line 322: ensures that if I subscribe to `/priv/app/` I receive `/priv/app/` only events,  and not `/priv/other/z.txt` event

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
...
+    );
+    drop(stream);
+
+    let mut stream = pubky
+        .event_stream_for(&server.public_key())
+        .add_users([(&user, None)])
+        .unwrap()
+        .session(&session)
+        .path("/pub/")
+        .path("/priv/app/")
```
MCarlomagno, line 374: ensures that if I subscribe to `/priv/app/` AND `/pub/` I receive `/priv/app/` and `/pub/` events, and not `/priv/other/` event

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
...
+            break;
+        }
+    }
+    assert!(paths.iter().any(|p| p.contains("/pub/a.txt")));
+    assert!(paths.iter().any(|p| p.contains("/priv/app/secret.txt")));
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_session_does_not_override_target_homeserver() {
```
MCarlomagno, line 399: If I subscribe to other user, and I also pass my session, the SDK must still target other users homeserver, not my session homeserver

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
...
+    assert!(
+        event.resource.path.as_str().contains("/pub/hello.txt"),
+        "expected other's public event, got: {}",
+        event.resource.path
+    );
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_signin_cookie_private_authorized() {
```
MCarlomagno, line 455: a legacy cookie is still allowed to listen to private events

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
...
+            .as_str()
+            .contains("/priv/app/signin-cookie.txt"),
+        "expected the signin-cookie private event, got: {}",
+        event.resource.path
+    );
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_cookie_bound_homeserver_is_enforced() {
```
MCarlomagno, line 501: A cookie session created on one homeserver must not be sent to another homeserver for a private stream, cookies must be bound to a single HS

e2e/src/tests/events/stream_sdk.rs
```diff
@@ -299,3 +316,273 @@ async fn events_stream_sdk_builder_api() {
...
+    };
+    assert!(
+        message.contains("cannot attach session credential"),
+        "got: {message}"
+    );
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn events_stream_sdk_cookie_live_private_receives_in_scope() {
```
MCarlomagno, line 540: A live private event stream using a cookie session receives only private events inside the requested path

pubky-common/src/storage.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Classifies a user-supplied path filter after WebDAV-style normalization.
+pub fn is_private_path_filter(path: &str) -> bool {
+    let Some(path) = normalize_path_filter(path) else {
+        return false;
+    };
+    is_private_path(&path)
+}
+
+fn normalize_path_filter(path: &str) -> Option<String> {
```
MCarlomagno, line 22: takes a user-provided path string and turns it into a normalized absolute path, this catches cases where the client might create weird routing to access unexpected paths using WebDAV valid routes like `.` or `..`

```
"/pub/../priv/secret/" -> "/priv/secret/"
"priv/x"               -> "/priv/x"
"/a/./b"               -> "/a/b"
"/a//b"                -> "/a/b"
"/../../priv/secret/"  -> None
""                     -> None
```

pubky-homeserver/src/client_server/auth/grant/middleware.rs
```diff
@@ -13,40 +13,13 @@
...
-    let Some(raw_token) = value.strip_prefix("Bearer ") else {
-        return BearerTokenExtraction::Invalid;
-    };
-
-    match SessionBearer::parse(raw_token) {
-        Ok(bearer) => BearerTokenExtraction::Present(bearer),
-        Err(_) => BearerTokenExtraction::Invalid,
-    }
-}
-
```
MCarlomagno, line 49: refactor: moved to `bearer.rs`

pubky-homeserver/src/client_server/auth/middleware/authentication.rs
```diff
@@ -128,4 +146,26 @@ mod tests {
         let resp = svc.oneshot(req).await.unwrap();
         assert_eq!(resp.status(), StatusCode::OK);
     }
+
+    #[tokio::test]
+    #[pubky_test_utils::test]
+    async fn invalid_bearer_falls_through_to_user_addressed_cookie() {
```
MCarlomagno, line 152: when doing a regular request, invalid Bearer can fallback to Cookies, because the request intentionally addresses the target user (the client is specifying the user as part of headers). So we allow the fallback.

We prevent the fallback in events stream because the request is addressed to the homeserver, not a single user (the client just specify the HS), then we can't allow fallback to cookies because that could be unexpected behaviour, bad authorization bearer but still authorized.

This should be fixed once we get rid of the hybrid model

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -82,6 +82,36 @@ pub struct EventStreamQueryParams {
     pub paths: Vec<WebDavPath>,
 }
 
+#[derive(Clone, Copy)]
+enum EventStreamTenantScope<'a> {
```
MCarlomagno, line 87: Allowed scopes are Public paths or Private paths for a single user/tenant

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -325,15 +355,24 @@ pub async fn feed(
...
     let params =
         parse_query_params(raw_query.0.as_deref().unwrap_or("")).map_err(HttpError::from)?;
+    let tenant_scope = EventStreamTenantScope::from_query(&params.paths, &params.user_cursors);
+
+    // Bearer auth disables the homeserver-addressed cookie fallback.
+    let session = match session {
+        Some(session) => Some(session),
+        None if !has_bearer_auth(&headers) => {
+            resolve_tenant_cookie_session(&state, &cookies, tenant_scope).await
+        }
```
MCarlomagno, line 350: falls back to cookies if there is no bearer auth

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -502,39 +541,41 @@ async fn resolve_user_cursors(
...
+    extract_bearer_token(headers).has_bearer_scheme()
+}
+
+async fn resolve_tenant_cookie_session(
+    state: &AppState,
+    cookies: &Cookies,
+    scope: EventStreamTenantScope<'_>,
+) -> Option<AuthSession> {
+    let EventStreamTenantScope::PrivateSingleTenant(tenant) = scope else {
+        return None;
```
MCarlomagno, line 533: no session needed if public paths

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -502,39 +541,41 @@ async fn resolve_user_cursors(
...
+        )
+        .into()]);
     }
 
-    // The tenant to authorize each path against. A private read is single-tenant,
-    // so a tenant only exists when the subscription names exactly one user.
-    let tenant = match user_cursors {
-        [(pubkey, _)] => Some(pubkey),
-        _ => None,
-    };
```
MCarlomagno, line 511: this is now handled by `EventStreamTenantScope`

pubky-sdk/src/actors/auth/cookie/credential.rs
```diff
@@ -54,6 +56,8 @@ pub struct CookieCredential {
     /// Cookie secret captured from `Set-Cookie`. `None` only on browser
     /// WASM where the value is hidden by the fetch spec.
     cookie: Option<String>,
+    /// Homeserver this cookie may attach to.
+    homeserver: Arc<RwLock<Option<PublicKey>>>,
```
MCarlomagno, line 60: Before, cookie auth mostly used user-addressed requests like:
```
https://_pubky.<user>/session
```

This PR adds private event streams that are homeserver-addressed:
```
https://<homeserver>/events-stream?user=<user>&path=/priv/app/
```

then we need to remember the homeserver pk and the SDK may not immediately know which homeserver issued that cookie

pubky-sdk/src/actors/auth/cookie/credential.rs
```diff
@@ -160,6 +176,26 @@ impl CookieCredential {
     }
 }
 
+fn session_resource(user: &PublicKey) -> String {
+    format!("pubky{}{}", user.z32(), SESSION_PATH)
+}
+
+async fn session_request(
```
MCarlomagno, line 183: now that homeserver requests are authenticated too, we need 2 different kinds of session requests, homeserver requests or pubky (single user) requests

pubky-sdk/src/actors/auth/cookie/credential.rs
```diff
@@ -201,26 +243,46 @@ impl SessionCredential for CookieCredential {
...
+            None => {
+                // Browser WASM keeps the secret in the cookie jar.
+                #[cfg(target_arch = "wasm32")]
+                {
+                    Ok(rb.fetch_credentials_include())
+                }
+                #[cfg(not(target_arch = "wasm32"))]
+                {
+                    Ok(rb)
+                }
```
MCarlomagno, line 261: this fixes a legacy issue, browser JS cannot reliably read `Set-Cookie`, so cookie credentials may have `cookie: None` and must rely on the browser cookie jar

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -332,10 +332,13 @@ impl GrantCredential {
         .await?;
         let body = serde_json::json!({ "grant": &state.grant_jws, "pop": pop_jws });
 
-        let url = format!("pubky{}/auth/grant/session", state.grant_claims.iss.z32());
-        let resolved = resolve_pubky(&url)?;
         let resp = client
-            .cross_request(Method::POST, resolved)
+            .cross_request_via_homeserver(
```
MCarlomagno, line 336: we always prefer to revalidate against the original homeserver rather than using the pubky resolution, as the session is bound to a homeserver

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -365,14 +382,9 @@ impl SessionCredential for GrantCredential {
...
-        // Hit the auth endpoint directly and attach the bearer ourselves —
-        // `/auth/grant/session` is not a storage URL.
-        let user_pk = self.state.lock().await.grant_claims.iss.clone();
-        let url = format!("pubky{}/auth/grant/session", user_pk.z32());
-        let resolved = resolve_pubky(&url)?;
         let bearer = self.current_bearer().await;
-        let response = client
-            .cross_request(Method::DELETE, resolved)
+        let response = self
+            .grant_session_request(client, Method::DELETE)
```
MCarlomagno, line 387: same as before

pubky-sdk/src/actors/auth/grant/credential.rs
```diff
@@ -396,20 +408,18 @@ impl SessionCredential for GrantCredential {
...
-        // attach the bearer ourselves — `/auth/grant/session` is not a
-        // storage URL.
-        let user_pk = self.state.lock().await.grant_claims.iss.clone();
-        let url = format!("pubky{}/auth/grant/session", user_pk.z32());
-        let resolved = resolve_pubky(&url)?;
         let bearer = self.current_bearer().await;
-        let response = client
-            .cross_request(Method::GET, resolved)
+        let response = self
+            .grant_session_request(client, Method::GET)
```
MCarlomagno, line 422: same

pubky-sdk/src/actors/event_stream.rs
```diff
@@ -317,14 +376,38 @@ impl EventStreamBuilder {
...
+
+    fn auth_scope(&self) -> EventStreamAuthScope<'_> {
+        if !self.has_private_path_filter() {
+            return EventStreamAuthScope::PublicOnly;
+        }
+
+        match self.users.as_slice() {
+            [(user, _)] => EventStreamAuthScope::PrivateSingleUser(user),
+            _ => EventStreamAuthScope::PrivateUnsupported,
+        }
```
MCarlomagno, line 401: return PrivateSingleUser only if there is a private path

pubky-sdk/src/actors/event_stream.rs
MCarlomagno: Yes but in that case fails server side, see later in this same function:

```rust
// Surface homeserver rejections (e.g. 401/403/400 for private-path
// authorization) as a typed `RequestError::Server` carrying the status
// and the server's message, rather than a hand-rolled string.
let response = check_http_status(response).await?;
```

andrei-21 (approved):

pubky-homeserver/src/client_server/routes/events.rs
86667: Could you improve the comment to explain why were doing this here rather than in the middleware. Or maybe on `resolve_tenant_cookie_session()`. 

I do think that doing this authentication work in the service here is the least bad approach considering the existing state and that cookie auth is being deprecated anyway, but we should clearly document why to avoid someone attempting a refactor.

pubky-common/src/storage.rs
86667: This is a useful comment. Can you add as docs to the fn

pubky-homeserver/src/client_server/routes/events.rs
86667: comment this too. Either as code comment or in fn docs

pubky-sdk/src/actors/auth/cookie/credential.rs
86667: Could be a better named fn, also please add docs

86667:
I made a few comments asking for more docs/comments. Overall i think the hackiness is fine considering Cookie is a soon deprecated flow and the alternative route would be a large change.

- https://github.com/pubky/pubky-core/pull/470#discussion_r3570143666
- https://github.com/pubky/pubky-core/pull/470#discussion_r3570049127
- https://github.com/pubky/pubky-core/pull/470#discussion_r3570157269
- https://github.com/pubky/pubky-core/pull/470#discussion_r3570183916

One potential issue is that an authenticated SSE stream whose session gets revoked will continue to pull priv events. Im not sure how big of a concern that is but we should at least acknowledge it.

PR #472, opened 2026-07-01
feat(testnet): `StaticTestnet` data dir persist mode

### LINKED ISSUE #185: Add configurable data dir to config for pubky-testnet

86667:
`StaticTestnet` uses an ephemeral temp directory, so all homeserver state (user data, files, config) is lost on shutdown. There is no way to persist a testnet environment, users have been using docker-compose presumably with production mainline - [discussion in issue](https://github.com/pubky/pubky-core/issues/185#issuecomment-3138343305). 

This is annoying for users and a gap which needs to be closed. Weve seen it brought up in telegram a couple times. 

An added bonus is that i've wanted to improve `StaticTestnet` for a while so used this opportunity to sneak in a refactor to make it easy to expand. I also improved docs to make it clearer how `EphemeralTestnet` and `StaticTestnet` differ and what they should each be used for.

Note that guides will be coming soon for how `pubky-testnet` can be used for development.  In the readme I intentionally stick to discussing only the crate's behaviour.

Here we:
- feat: Add a `persist <data_dir>` subcommand to the pubky-testnet binary allowing it to store homeserver state on disk across restarts
- feat: Seed a deterministic keypair on first run so the homeserver identity is stable and matches the in-memory default
- refactor: Introduce `StaticTestnetBuilder` with `.persistent(path)` and `.homeserver_config(path)` for programmatic use
- refactor: Other cleanup on `StaticTestnet` and `EphemeralTestnet`


pubky-testnet/src/static_testnet.rs
```diff
@@ -2,73 +2,223 @@ use crate::pubky::Pubky;
...
+            }
+            StorageMode::Persistent(data_dir) => {
+                testnet
+                    .run_persistent_homeserver(data_dir, self.homeserver_config.as_deref())
+                    .await
+                    .map_err(|e| anyhow::anyhow!("Failed to run persistent homeserver: {}", e))?;
+                true
+            }
+        };
+        testnet.persistent = persistent;
```
MCarlomagno: I feel `persistent` for a bool var is a bit confusing, I would rather use `is_persistent`

pubky-testnet/src/static_testnet.rs
```diff
@@ -175,41 +326,426 @@ impl StaticTestnet {
...
 
         let homeserver = HomeserverApp::start_with_mock_data_dir(mock).await?;
         self.testnet.homeservers.push(homeserver);
         Ok(())
     }
 }
+
+/// Copy the source config file into the persistent data directory.
+/// Errors if a `config.toml` already exists at the destination.
+fn seed_config(source: &Path, persistent_dir: &PersistentDataDir) -> anyhow::Result<()> {
```
MCarlomagno: nit: I usually try to clean up standalone functions by making them part of higher level structs/enums, in this case I would make it  a private function of `StaticTestnet`, but this is personal taste

pubky-testnet/src/static_testnet.rs
```diff
@@ -175,41 +326,426 @@ impl StaticTestnet {
...
+    let config_path = persistent_dir.get_config_file_path();
+    if config_path.exists() {
+        anyhow::bail!(
+            "config.toml already exists at {}. Remove --homeserver-config to use the existing config, \
+             or delete the file to replace it.",
+            config_path.display()
+        );
+    }
+    std::fs::create_dir_all(persistent_dir.path())?;
+    std::fs::copy(source, &config_path)?;
```
MCarlomagno: nit: also something I usually avoid is to leave full import chains inline like this, I prefer something like `fs::copy(source, &config_path)` instead, but very personal too

MCarlomagno:
other than the nits, the README tells users to omit `?pubky-test=true` in persistent mode, but nothing enforces it. Then the persistent mode could route into the epheremeral path and create a fresh `pubky_test_{uuid}` database each run

pubky-testnet/src/static_testnet.rs
86667: Changed to `is_persistent`

86667:
> the README tells users to omit `?pubky-test=true` in persistent mode, but nothing enforces it.

@MCarlomagno good point. Ive added a check for the flag when in persistent mode

MCarlomagno (dismissed):
Thanks @86667, there is still an edge case 

what if `config.toml` itself contains `database_url = "...?pubky-test=true"`? I think the check should run against the final `config.general.database_url`

I mark the PR as approved since this is not a functional blocker

86667:
@MCarlomagno good catch again - moved check later so that it applies to all configuration routes

MCarlomagno (approved):

PR #473, opened 2026-07-01
test(homeserver): private storage user quotas tests

MCarlomagno:
Closes #440.

### Summary

The verification + tests half of `/priv/` quota support. Storage accounting and `allowed_write_paths` enforcement were already path-agnostic — the quota pre-check, the `used_bytes` accounting in `UserQuotaLayer`, and `is_write_path_allowed` never special-case `/pub/` — so no production changes were needed. This PR adds the tests that prove `/priv/` behaves identically to `/pub/` and locks that in against regressions.

Coverage added:
- `used_bytes` accounting + quota rejection for `/priv/` writes (unit + e2e),
  including a test that `/priv/` and `/pub/` share one storage bucket.
- `allowed_write_paths: ["/priv/app/"]` enforcement at both the matching-logic
  and OpenDAL-layer level.
- The Content-Length quota pre-check on a `/priv/` path.

### Acceptance criteria
- [x] A `/priv/` write increments `used_bytes` and is rejected at quota like a `/pub/` write.
- [x] `allowed_write_paths: ["/priv/app/"]` permits `/priv/app/x`, denies `/priv/other/x` and `/pub/...`.
- [x] Verification + new tests; confirmed no code assumes `/pub/`.

PR #474, opened 2026-07-01
test(homeserver): e2e for private data

### LINKED ISSUE #441: e2e authz for private data
### Context
Private data is a security boundary and needs an explicit cross-cutting test matrix in the `e2e` crate.

### Scope
Matrix of **actor** × **operation** × **expected**:
- actors: owner-with-cap, owner-without-covering-cap, other tenant, anonymous.
- operations: `PUT`, `GET`, `HEAD`, `DELETE`, directory `LIST`, `/events/` feed, `/events-stream`.
- assert the **status-by-auth-tier** contract, invariant to existence: anonymous → `401`,
  authed-but-under-scoped or wrong-tenant → `403`, authorized → `200`/`404`. The same status must be
  returned for an existing and a nonexistent private resource at each non-authorized tier.
- assert no `/priv/...` path appears in the anonymous `/events/` feed or in unauthenticated /
  no-`path` `/events-stream` output.

### Affected files
- `e2e/` (new test module, e.g. `tests/private_data.rs`).
- May reuse helpers from `pubky-testnet`.

### Acceptance criteria
- [ ] All cells of the matrix asserted and green.
- [ ] Existence-oracle regression: identical status for existing vs nonexistent private resources at
      the anonymous (`401`) and under-scoped (`403`) tiers.
- [ ] Leak regression: no private path via the anonymous `/events/` feed or no-`path` stream.

## Dependencies

Part of epic #426.

Blocked by:
- #432, #433, #434 — write / read / list behaviors under test
- #436 — public-feed filtering
- #429 — `/events-stream` authz

MCarlomagno:
Closes #441.

### Summary

Cross-cutting e2e authorization matrix for private (`/priv/`) data. This adds an explicit actor × operation × expected grid in a new `e2e/src/tests/private_data.rs` module. Tests-only, no production changes.

Each of the four actor tiers hits the owner's resource URL with only its credential varied, asserting the status-by-auth-tier contract, invariant to existence:
- **anonymous** → `401`
- **owner with an under-scoped cap** / **other tenant** → `403`
- **owner with a covering cap** → `200`/`404`

### Acceptance criteria
- [x] All cells of the matrix asserted and green.
- [x] Existence-oracle regression: identical status for existing vs nonexistent private resources at the anonymous (`401`) and under-scoped/wrong-tenant (`403`) tiers.
- [x] Leak regression: no private path via the anonymous `/events/` feed (no-`path` stream covered by `events::stream_private`).

### DIFF
--- a/e2e/src/tests/private_data.rs
+++ b/e2e/src/tests/private_data.rs
@@ -1,0 +1,300 @@
+//! Authorization matrix for private (`/priv/`) resources: each actor tier
+//! (anonymous, under-scoped owner, other tenant, authorized owner) against
+//! every storage action.
+//!
+//! `/events-stream` authorization is covered in `events::stream_private`.
+
+use super::build_full_testnet;
+use pubky_testnet::pubky::{
+    AuthFlowKind, ClientId, IntoPubkyResource, Keypair, Method, PubkyGrantAuthFlow,
+    PubkyHttpClient, PubkySession, PubkySigner, StatusCode,
+};
+use pubky_testnet::pubky_common::capabilities::Capabilities;
+use pubky_testnet::pubky_common::crypto::PublicKey;
+use pubky_testnet::EphemeralTestnet;
+
+const SECRET: &str = "/priv/app/secret.txt";
+const ABSENT: &str = "/priv/app/absent.txt";
+const DIR: &str = "/priv/app/";
+
+/// Approve a grant with `caps` for `signer` and return the resulting session.
+async fn grant_session(
+    testnet: &EphemeralTestnet,
+    signer: &PubkySigner,
+    caps: Capabilities,
+) -> PubkySession {
+    let pubky = testnet.sdk().unwrap();
+    let auth = PubkyGrantAuthFlow::builder(
+        &caps,
+        AuthFlowKind::signin(),
+        ClientId::new("test.app").unwrap(),
+    )
+    .relay(testnet.http_relay().local_link_url())
+    .client(pubky.client().clone())
+    .start()
+    .unwrap();
+    signer
+        .approve_auth(&auth.authorization_url())
+        .await
+        .unwrap();
+    auth.await_approval().await.unwrap()
+}
+
+/// Transport URL for `path` in `owner`'s namespace.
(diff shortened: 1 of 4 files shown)


e2e/src/tests/private_data.rs
```diff
@@ -0,0 +1,300 @@
...
+    );
+    assert_eq!(
+        req_status(client, Method::HEAD, &absent, bearer, None).await,
+        StatusCode::NOT_FOUND
+    );
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn anonymous_events_feed_excludes_private() {
```
86667: should this live with events-stream tests

e2e/src/tests/private_data.rs
```diff
@@ -0,0 +1,300 @@
...
+    )
+    .await;
+    let token = tenant.as_grant().unwrap().current_bearer().await;
+
+    assert_all_verbs_denied(pubky.client(), &owner, Some(&token), StatusCode::FORBIDDEN).await;
+}
+
+#[tokio::test]
+#[pubky_testnet::test]
+async fn owner_with_cap_priv_access_is_authorized() {
```
86667: This one feels like it should live with auth tests

86667 (commented):
Hmm I see the benefits now to including these tests and being confident that priv data is working as expected. However I worry that if we keep adding feature-specific tests like this, especially e2e ones, we'll end up with 100s of tests full of redundancy and which are difficult to get a global view of. 

I may be wrong here so please let me know your thoughts. My intuition is that `/priv` auth concerns should live with other auth tests, events/events-stream `/priv` filtering with other events tests, and storage PUT/DEL/READ also have their own `objects.rs` tests, etc.

MCarlomagno:
@86667 yes I agree with you as commented [here](https://github.com/pubky/pubky-core/pull/473#issuecomment-4876685515), I would just extend existing tests for now, if there is some kind of functional change between `/priv` and `/pub` in the future other than authorization we should split them, but for now we can keep them as shared tests

[UPDATE] I would keep the `private_data.rs` file for the matrix described in the issue, and just move the tests you flagged

86667 (approved):

PR #477, opened 2026-07-02
fix(sdk)!: Surface errors on `get_homeserver_of`

### LINKED ISSUE #408: [SDK] `get_homeserver_of` should return error if they happen not swallowing it
`get_homeserver_of -> Option<SignedPacket>` currently swallows errors by returning `None`. This should not be the case though.
The underlying `pkarr.resolve` call does the same though not sure this is resolvable there.

https://github.com/pubky/pubky-core/blob/baefac72ec95ffc2c66c0088e28518730ce669ee/pubky-sdk/src/pubky.rs#L216

This should be fixed.

MCarlomagno:
Closes #408.

## Summary

Pkarr v7 distinguishes a genuinely missing record (`ResolveError::NotFound`) from operational resolution failures. This PR updates homeserver resolution to preserve that distinction instead of treating every failure as “no homeserver.”

## Behavior change

| Outcome | Before | After |
|---|---|---|
| Valid `_pubky` target | `Some(pk)` | `Ok(Some(pk))` |
| Record not found | `None` | `Ok(None)` |
| Packet without `_pubky` | `None` | `Ok(None)` |
| Operational resolution failure | `None` | `Err(Error::Pkarr(PkarrError::Resolve(_)))` |
| Malformed `_pubky` target | `None` | `Err(Error::Pkarr(PkarrError::InvalidRecord(_)))` |

Operational resolution errors are retryable through `PkarrError::is_retryable()`.

The publish path now applies the same validation before reusing an existing `_pubky` target, so it will not republish a target the read path rejects.

## Breaking changes

- `Pubky::get_homeserver_of` and `Pkdns::get_homeserver_of` change from `Option<PublicKey>` to `Result<Option<PublicKey>>`.
- JavaScript `getHomeserverOf()` still resolves to `PublicKey | undefined` on success, but now rejects on operational resolution failures or malformed records.
- Existing `Result`-returning APIs have a wider runtime error surface despite no signature change:
  - `Pkdns::get_homeserver`
  - `signin()` and `signin_blocking()`
  - event-stream subscriptions without an explicit homeserver
  - grant credential exchange
  - `signin_cookie()` and `signin_cookie_blocking()`
- `require_homeserver_of` now returns `RequestError::Validation` only for genuine absence. Operational failures propagate as `Error::Pkarr` instead of being reported as “user has no homeserver.”
- Unbound cookie `signout()` and `revalidate()` retain their best-effort pubky-URL fallback for operational resolution failures, while malformed records remain visible as `Error::Pkarr(PkarrError::InvalidRecord(_))`.

Consumers matching specifically on `RequestError::Validation` to detect an unsigned-up user should handle `Error::Pkarr` separately.

## Validation

- Added coverage for `NotFound`, operational resolution errors, and malformed records.
- Added a regression test preventing malformed `_pubky` targets from being republished.
- Added coverage for the unbound-cookie operational-error fallback and malformed-record propagation.
- `cargo test -p pubky --lib` (164 passed).
- Native clippy with warnings denied, allowing the existing `clone_on_ref_ptr` and restored `single_match_else` lints.
- `wasm32-unknown-unknown` check.


86667 (changes requested):
Good PR description.

Please mark the conventional commit message as `BREAKING CHANGE:`

86667 (dismissed):

86667:
> Please mark the conventional commit message as BREAKING CHANGE:

Sorry @MCarlomagno i was mixing up what we did in a previous team. Ive set the title to the correct format.

pubky-sdk/src/actors/pkdns.rs
```diff
@@ -139,45 +139,57 @@ impl Pkdns {
...
+    ) -> Result<Option<PublicKey>> {
         cross_log!(
             info,
             "Resolving homeserver for public key {} via PKARR",
             user_public_key
         );
-        let packet = self.client.pkarr().resolve(user_public_key).await?;
-        let s = extract_host_from_packet(&packet)?;
-        let result = PublicKey::try_from_z32(&s).ok();
+        let Some(packet) = self.client.pkarr().resolve(user_public_key).await else {
```
andrei-21: We need to wait for a new pkarr release where `resolve()` will be returning `Result<SignedPacket, ResolveError>`.

andrei-21 (changes requested):

pubky-sdk/src/actors/pkdns.rs
```diff
@@ -139,38 +140,62 @@ impl Pkdns {
...
+
+    /// Convert a Pkarr resolution into the public homeserver lookup result.
+    ///
+    /// A missing packet is a valid absence; all operational resolution failures remain errors.
+    fn homeserver_pubkey_from_resolution(
+        resolution: std::result::Result<SignedPacket, ResolveError>,
+    ) -> Result<Option<PublicKey>> {
+        match resolution {
+            Ok(packet) => homeserver_pubkey_from_packet(&packet),
+            Err(ResolveError::NotFound) => Ok(None),
```
andrei-21, line 188: Shouldn't we treat `ResolveError::InvalidSignedPacket { seq }` as `Ok(None)` in this context?
The doc says:
> `Ok(None)` when no Pkarr record exists, or the record carries no `_pubky` target

MCarlomagno: I think this should remain as an error. "not found" would be misleading when the actual result is "exists but cannot be validated", plus the purpose of the PR is surfacing any resolution failures.

The cases for "not found" (no pkarr record or a record without `_pubky` target) are the actual cases of unreachable homeserver

But maybe I'm missing something here, why do you think this error should be supressed?

SeverinAlexB: My gut tells me that is should be an error but not 100% sure.You guys decide.

andrei-21 (approved):

PR #482, opened 2026-07-06
feat(homeserver|sdk)!: signup deeplinks auth

MCarlomagno:
Supersedes #300.

### Summary

Adds first-class support for direct signup deep links via a new `pubkyauth://direct_signup` intent.

Direct signup links carry only the target homeserver, plus an optional signup token:

- `pubkyauth://direct_signup?hs=<homeserver>` — approving registers the account directly on that homeserver via `PubkySigner::signup`.
- `pubkyauth://direct_signup?hs=<homeserver>&st=<token>` — same, for token-gated homeservers.

No relay round-trip is involved: approving the link performs the signup immediately, and there is no completion callback to the initiating app. Existing relayed signup/cookie flows are unchanged and still require `caps`, `relay`, and `secret`.

### ⚠️ Breaking change

This PR adds a `DirectSignup` variant to the public `DeepLink` enum, which is not `#[non_exhaustive]`, so downstream exhaustive `match`es on `DeepLink` will fail to compile. The next release should be a `v0.10` bump, not a patch.

### Acceptance criteria

- [x] `pubkyauth://direct_signup` links parse and round-trip with and without `st`.
- [x] Approving a direct signup link creates a homeserver account and the user can sign in afterward.
- [x] Direct signup works on homeservers that require a signup token.
- [x] Direct signup URLs are rejected by `resume_cookie_auth_flow`.
- [x] Relayed signup/cookie flows remain supported.
- [x] JS bindings cover direct signup parsing and approval.


SeverinAlexB (changes requested):
Looking at the code, I think you are merging two things together that should not necessarily be together.

I feel like the two uses cases should be two separate deep links:
- The existing `signup` is for signup and auth at the same time.
- The new signup (`simple-signup`? Maybe you can think of a better intent name) should only be the signup.

The reason I am saying this is because 
- the SignupParams struct looks very awkward. In theory, I can create an invalid signup link by setting a secret but no relay. This is usually a sign one needs to split up the struct.
- This change is breaking (not mentioned in the PR title by the way). If possible, I would like to avoid this.
- The `is_direct_signup()` method is another hint that this should be two separate structs.

```rust
/// Typed parameters for signup deep links.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct SignupParams {
    /// Capabilities requested by the app.
    pub capabilities: Capabilities,
    /// Base HTTP relay URL. `None` for a direct signup link.
    pub relay: Option<Url>,
    /// Secret used to derive the encrypted relay channel. `None` for a direct
    /// signup link.
    pub secret: Option<[u8; 32]>,
    /// Homeserver public key.
    pub homeserver: PublicKey,
    /// Optional signup token.
    pub signup_token: Option<String>,
}

impl SignupParams {
    /// Returns `true` when this is a direct signup link (no relay/secret).
    #[must_use]
    pub fn is_direct_signup(&self) -> bool {
        self.relay.is_none() && self.secret.is_none()
    }
}
```

Does this make sense?

MCarlomagno:
@SeverinAlexB agree,

I would call them `signup` and `direct_signup`, where:

- signup requires `caps`, `relay`, `secret`, `hs` and optional `st`
- direct_signup requires `hs` and optional `st`

e2e/src/tests/auth/signup_tokens.rs
```diff
@@ -184,3 +184,138 @@ async fn get_signup_token() {
...
+    let testnet = build_full_testnet().await;
+    let server = testnet.homeserver_app();
+    let pubky = testnet.sdk().unwrap();
+
+    let signer = pubky.signer(Keypair::random());
+    let deeplink = format!("pubkyauth://direct_signup?hs={}", server.public_key().z32());
+
+    // Approving a direct signup link registers the account directly on the
+    // homeserver.
+    signer.approve_auth(&deeplink).await.unwrap();
```
SeverinAlexB: This is not the correct method for this. You do not approve any authentication with this new deep link. You just signup.
We can talk about a general `approve_deeplink()` method if you want. Let me know if you have another idea.

SeverinAlexB (changes requested):
Nice work. The only thing that is suboptimal is the reuse of approve_auth. Otherwise good to go.

e2e/src/tests/auth/signup_tokens.rs
MCarlomagno: I agree, I would use `handle_deeplink`

SeverinAlexB (approved):

PR #487, opened 2026-07-10
docs(examples): Added grant auth examples + reordered/refactored examples

SeverinAlexB:
This PR refreshes a lot of rust/js examples:
- Uses Grant auth in the auth flow examples.
- Adds a js auth flow example
- Adds rust/js session management example
- Renumbers the examples so the important once are on top (number 1,2, and so on)
- Refactors the examples so less CLI arguments are needed. For example, it now uses a default keypair so one doesn't need to create one for running with testnet.


examples/rust/keygen.rs
```diff
@@ -26,9 +26,6 @@ fn main() -> Result<()> {
     // 2) Encrypt and save the recovery file
     println!("Enter a passphrase to encrypt the recovery file:");
     let passphrase = rpassword::read_password()?;
-    if passphrase.is_empty() {
-        anyhow::bail!("Passphrase cannot be empty");
-    }
```
MCarlomagno, line 31: Is this intentional? I think that could lead to unencrypted keys due to empty passphrase

examples/rust/recovery.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
+use pubky::Keypair;
+use std::path::{Path, PathBuf};
+
+pub const SAMPLE_RECOVERY_FILE: &str = "sample_recovery.key";
+
+pub fn sample_recovery_file() -> PathBuf {
```
MCarlomagno: Just as an extra security measure, we could enforce testnet when sampling a recovery file just to make sure it's not used in prod


something like:
```rust
pub fn resolve_recovery_file(file_path: Option<PathBuf>, testnet: bool) -> Result<PathBuf> {
    match file_path {
        Some(path) => Ok(path),
        None if testnet => Ok(sample_recovery_file()),
        None => Err("The sample key is only for testnet!" ),
    }
}
```

examples/rust/1-signup/README.md
```diff
@@ -0,0 +1,26 @@
+# Authentication examples
```
MCarlomagno: nit: I think this title is too generic, maybe "Signup-only example"

examples/javascript/2-authenticator.mjs
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env node
+// Approve a Pubky Auth URL (QR/deeplink) using a recovery file.
+// If --testnet, we first ensure the user exists by signing up locally.
+import { Pubky, Keypair, PublicKey } from "@synonymdev/pubky";
+import { args, promptHidden, readFileUint8 } from "./_cli.mjs";
+
+const TESTNET_HOMESERVER = "8pinxxgqs41n4aididenw5apqp1urfmzdztr8jt4abrkdn435ewo";
```
MCarlomagno: nit: I've seen this constant being declared across multiple places, maybe we can expose it from some shared location, same in rust examples

examples/javascript/2-authenticator.mjs
```diff
@@ -0,0 +1,58 @@
...
+
+// 3) If testnet, ensure we exist (signup once)
+if (a.testnet) {
+  const homeserver = PublicKey.from(TESTNET_HOMESERVER);
+  try {
+    await signer.signup(homeserver);
+    console.log("Signed up to the testnet homeserver.");
+  } catch {
+    console.log("Testnet user already exists, continuing...");
+  }
```
MCarlomagno, line 60: here in the catch we might be ignoring other errors

examples/rust/keygen.rs
SeverinAlexB: It is intentional yes. The goal is to make these examples as easy as possible to run. Any password prompt is friction so I tried to avoid it.

I added a new warning though to prevent empty password in production. This is hopefully enough to prevent people using this in prod.

```rust
if passphrase.is_empty() {
    println!("Warning: You entered an empty passphrase. This is not recommended for a production environment.");
}
```

examples/javascript/2-authenticator.mjs
SeverinAlexB: I added a log & throw in case it is not a "user already exists error"

MCarlomagno (dismissed):

86667 (changes requested):
These examples require a running testnet but nowhere is this mentioned. Could we add a "How to use these examples" section?

Also nit: this is not a `chore`, should be `feat` or even `docs`

examples/rust/README.md
```diff
@@ -4,15 +4,17 @@ Minimal examples for different flows and functions you might need to implement u
...
 
-0. [**Logging**](./0-logging/README.md): configure tracing and watch the SDK emit debug output during a storage roundtrip.
-1. [**Testnet**](./1-testnet/README.md): shows how to build a pubky app offline against a local ephemeral homeserver.
-2. [**Authentication**](./2-signup/README.md): shows how to signup, signin or signout to and from a homeserver.
-3. [**Authorization Flow**](./3-auth_flow/README.md): shows how to setup Pubky authz for a 3rd party application and how to implement an authenticator to sign in such app.
-4. [**Storage**](./4-storage/README.md): public reads and authenticated writes on homeserver storage.
-5. [**Request**](./5-request/README.md): shows how to make direct HTTP requests to Pubky URLs.
-6. [**Signup Authorization Flow**](./6-auth_flow_signup/README.md): shows how to setup Pubky authz for a 3rd party application and how to implement an authenticator to sign up such app.
-7. [**Event Stream**](./7-events_stream/README.md): subscribe to Server-Sent Events from a user's homeserver.
+1. [**Authentication**](./1-signup/README.md): shows how to signup, signin or signout to and from a homeserver.
```
86667: The names here should match the names of the example files. Eg `Authentication` here points to `signup` which is confusing.

Also there are two examples numbered 6

SeverinAlexB: Update. I can't find the 6 number duplication though. Can you help me what you meant?

86667: Thanks. And 6 duplication was my mistake, all good

86667 (approved):

PR #493, opened 2026-07-13
chore(sdk): v0.10 migration guide

SeverinAlexB:
Adds a section to the migration guide about grant authentication.

Depends on #487.


docs/migration-v0.10.md
```diff
@@ -1,14 +1,26 @@
 # Migrating to v0.10
 
-This guide is for applications upgrading from `v0.9.x` that already use cookie-based SDK authentication and want that behavior to keep working.
+This guide covers applications upgrading from `v0.9.x`. It has a minimal cookie-auth compatibility path and a recommended grant-auth path for applications that can update their authentication model.
 
 Cookie auth still exists in v0.10, but the SDK now names the cookie-compatible APIs explicitly. The main migration is to replace the old generic signer methods with the new `*Cookie` methods where your app expects a cookie-backed session.
```
86667:
```suggestion
Cookie auth still exists in v0.10, but the SDK now names the cookie-compatible APIs explicitly. If you do not want to upgrade to grant-auth then the main migration is to replace the old generic signer methods with the new `*Cookie` methods where your app expects a cookie-backed session.
```

docs/migration-v0.10.md
```diff
@@ -1,14 +1,26 @@
...
 Some cookie-specific APIs are documented or marked as deprecated because cookie auth is now the compatibility path. That is expected for this migration.
 
+## Recommended: Grant Authentication
+
+For new applications, and existing applications able to make an intentional authentication-model change, use grant authentication. It provides capability-scoped, self-refreshing sessions and is the recommended replacement for cookie auth in v0.10.
+
+Grant auth is not a mechanical rename of the cookie APIs. A third-party app starts a grant flow, displays its authorization URL as a QR code or deep link, and receives a grant-backed session after the user's authenticator approves the request.
+
+- JavaScript: use `pubky.startGrantAuthFlow(...)` and persist browser sessions with `pubky.browserSessionStore`.
+- Rust: use `pubky.start_grant_auth_flow(...)` or `PubkyGrantAuthFlow::builder(...)`.
```
86667: These first two bullet points are not useful. Either give a walkthrough here or just point to the code docs or  examples.

docs/migration-v0.10.md
86667: Can we mention at the top that there are two steps to the migration: code change to new API and check/update permission paths. This section may never be seen by some readers but is crucial to know about.

docs/migration-v0.10.md
86667: I think a short walk-through would be useful here, or elsewhere and pointed to. If i were a dev reading this i'd likley not understand what grant-auth means and just do the easier cookie path upgrade. Ideally though cookie upgrade is just for those with little time, we should encourage the grant-auth route and make that as easy as possible to upgrade to.

docs/migration-v0.10.md
86667: Also mention that the Cookie auth is now considered deprecated

86667 (commented):
As for the examples ive pushed some changes as it was easier to just do than explain here. 

Should `auth-flow-signup` go before `auth-flow`? Also needs updating to grant auth. Maybe both of these examples could be combined, its not immediately clear how they differ.

SeverinAlexB:
> Should auth-flow-signup go before auth-flow?

Signing up is a a rare operation with the auth flow IMO. That's why it put the regular auth-flow first.

> Also needs updating to grant auth. 

Can you point me to which one is not already grant auth?

> Maybe both of these examples could be combined, its not immediately clear how they differ.

I am tempted to delete the signup auth flow even. This will be such a rare case that I am not sure it's worth having. At the same time, it's very similar to the regular auth flow so it can easily be reproduced by looking at the regular auth flow.

docs/migration-v0.10.md
SeverinAlexB: Not 100% sure what you mean with permission paths?

docs/v0.10-migration/README.md
```diff
@@ -0,0 +1,63 @@
+# Migrating to v0.10
+
+This guide covers applications upgrading from `v0.9.x`. `v0.10` includes multiple breaking changes and most importantly, a new and more secure grant authentication system.
+
+It is strongly recommended to upgrade to **Grant authentication. Cookie authentication is now deprecated** and will be removed in future versions. It is [insecure](https://github.com/pubky/pubky-core/issues/520). Cookie auth still exists in v0.10, but the SDK now names the cookie-compatible APIs explicitly. If you do not want to upgrade to grant-auth then the main migration is to replace the old generic sign-in/up methods with the new `*Cookie` methods.
+
```
86667, line 6:
```suggestion

You will need to consider whether each of the breaking changes listed below will effect your code: 
```

docs/v0.10-migration/README.md
```diff
@@ -0,0 +1,63 @@
...
+let homeserver = &params.homeserver;
+let signup_token = params.signup_token.as_deref();
+```
+
+If your code matches on `DeepLink`, make sure it has a fallback or handles all variants. v0.10 adds extra variants, so exhaustive matches written against `v0.9.x` may fail to compile.
+
+
+## Capability Path Matching
+
+v0.10 tightens capability validation in various ways, for example a trailing slash is now significant.
```
86667: It is only actually the trailing slash behaviour which changed? Or if not, should the other changes be listed too?

docs/v0.10-migration/grant-auth.md
```diff
@@ -0,0 +1,163 @@
...
+| Supports the upcoming homeserver mirroring                 | Doesn't support multiple homeservers                         |
+| Recommended in Pubky v0.10+                                | Deprecated and scheduled for removal                         |
+
+The cookie belongs to the homeserver domain rather than the third-party application. Consequently, website B can receive the same homeserver cookie used by website A and potentially exercise A's permissions. Grant authentication prevents this by using bearer tokens, binding each authorization to an app-specific key, and explicitly scoped capabilities.
+
+## Client ID
+
+A new client ID is introduced when creating a grant authentication session. It is an identifier that is saved alongside the session in the homeserver. When a users lists all session with the new session management api, the client ID is listed.
+
+A client name can be any domain like string and should identify your application.
```
86667:
```suggestion
A new client ID is introduced when creating a grant authentication session. It is an identifier that is saved alongside the session in the homeserver. When a user lists all session with the new session management api, the client ID is listed.

A client name can be any domain-like string and should identify your application. 
```

docs/v0.10-migration/grant-auth.md
```diff
@@ -0,0 +1,163 @@
...
+
+A client name can be any domain like string and should identify your application. 
+
+Examples:
+- `pubkyapp.synonym.to`
+- `example-app`
+
+## Auth Flow / QR Login
+
+Grant auth flows add a required client ID and produce a grant-backed, self-refreshing session.
```
86667:
```suggestion
Grant auth flows require a client ID and produce a grant-backed, self-refreshing session.
```

docs/v0.10-migration/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```text
+/pub/app/:rw covers /pub/app/file.txt
+/pub/app:rw only covers /pub/app
+```
+
+If your app intends to grant access to everything under an app directory, use a trailing slash.
+
+```rust
+let caps = Capabilities::builder()
```
86667, line 55: written as:
```
let caps = Capabilities::builder()
    .read_write("/pub/my-cool-app/")
    .finish();
```
elsewhere

docs/v0.10-migration/cookie-auth.md
```diff
@@ -1,14 +1,11 @@
-# Migrating to v0.10
+# Not Recommended: Keep Cookie Authentication
 
-This guide is for applications upgrading from `v0.9.x` that already use cookie-based SDK authentication and want that behavior to keep working.
+Cookie authentication is deprecated and insecure. Prefer the new [Grant authentication system](./grant-auth.md). This guide shows how stick to Cookie authentication while migrating from `v0.9` to `v0.10`.
```
86667:
```suggestion
Cookie authentication is deprecated and insecure. Prefer the new [Grant authentication system](./grant-auth.md). This guide shows how to stick to Cookie authentication while migrating from `v0.9` to `v0.10`.
```

86667 (commented):
>  I am tempted to delete the signup auth flow even

Im happy with this, its easy to see that its possible to do from the fn docs. Could show it as an option in the auth-flow example with commented code.

86667:
I am wondering about the usefulness of 8-logging and 9-testnet. logging could be shown in each of the other examples, and 9-testnet is specifically about using an ephemeral testnet, is that actually very useful for devs?

docs/migration-v0.10.md
86667: I wanted it to be clear from the start of the file that there are multiple things which have changed, one being your `Capability Path Matching` section. The first suggestion in the next review below is what im looking for.

docs/v0.10-migration/README.md
SeverinAlexB: There are too many changes to list them all. The most important one is the trailing slash. This has a real effect on how things behave. The rest is more nuances or more like bug fixes that I don't think need to be mentioned.

docs/v0.10-migration/README.md
SeverinAlexB: This is a new change coming from https://github.com/pubky/pubky-core/pull/514 which has just been merged recently. 

Or what do you mean with "elsewhere"?

SeverinAlexB:
> I am wondering about the usefulness of 8-logging and 9-testnet. logging could be shown in each of the other examples, and 9-testnet is specifically about using an ephemeral testnet, is that actually very useful for devs?

Good question but out of scope for this PR IMO. Happy to have it removed but it's also not that big of a deal.


Incorporated your suggestions. I also let AI grammar check the docs which changed a few things.

docs/migration-v0.10.md
SeverinAlexB: Added the suggestion below.

docs/v0.10-migration/README.md
86667: in `v0.10-migration/README.md` you have an `unwrap()` on `.read_write("/pub/my-cool-app/")`, but in  `v0.10-migration/cookie-auth.md` you do not.

examples/rust/2-auth_flow/README.md
```diff
@@ -1,42 +1,56 @@
...
+# with a custom client ID or capabilities
 cargo run --bin auth_client -- --testnet \
   --client-id my-app.example \
   --capabilities /pub/my-app/:rw
 ```
 
 Copy the Pubky Auth URL from the client output. It should use the `signin_grant` intent and include `cid` and `cpk` query parameters.
 
-Finally run the authenticator in another terminal to approve it:
+To create an account on the local testnet and authorize the app in one flow, start the client with `--signup`:
```
86667: This is a little confusing. You dont explicitly mention before the command on line 27 that the user must be already signed up. Maybe split into "Signin an existing user", pointing to the signup example to do signup before the signin part. Then a "Signup and Signin in one step" part?

examples/rust/2-auth_flow/README.md
```diff
@@ -1,42 +1,56 @@
...
 
 For the browser version of the third-party app, see the JavaScript [2-auth-flow](../../javascript/2-auth-flow/README.md) example.
 
+## Prerequisites
+
+Examples using `--testnet` require a running local testnet. See the [examples README](../README.md#quick-start) for setup.
+
 ## Recovery File
 
 This example defaults to `../../sample_recovery.key`. You may supply a custom recovery file.
```
86667, line 20: I ran into `Sample recovery file is only available for testnet` error when running against non-testnet homeserver. Can we make this non-testnet version more clear, ie show how to generate a keypair, how to pass into `authenticate` command

86667 (commented):
Looking gooood. A couple points on clarity, and please update the numbers on examples 6,7,8,9. 

The JS examples probably need review at some point too. 

Edit: shall we remove the numbers? We could list a recommended order to run them in the README. Please create a ticket for review of:
- Rust examples - are logging and testnet useful? could they be folded into existing examples
- Compare JS examples with Rust examples and bring in-line
- Consider removing numbering in favour of recommended ordering in readme
- Anything else you think could be improved

examples/rust/2-auth_flow/README.md
SeverinAlexB: I improved the panic message but this is out of scope for this PR. I am also very cautious how far we should go with making these example mainnet ready. They are examples, not applications you should use on mainnet.

examples/rust/2-auth_flow/README.md
SeverinAlexB: Restructured the readme. If I knew how thorough you want this I would have just removed the --signup example instead of adding it to the auth flow. The signup auth flow is really not that important and I wouldn't be surprised that usage will completely disappear as soon as we have a fullly working Pubky Passport aka removing the signup auth flow from PubkyApp.

SeverinAlexB:
I renumbered the folder. I think the numbering is important. The order indicates how important an example is. That's why the numbers of logging and testnet is so high.

I added this issue https://github.com/pubky/pubky-core/issues/532. I am personally very cautious with extending the examples. Examples are there to teach devs how code works. I am way more concerned with clean and easy to read code than any README or mainnet support. Lots of other repos do not even have a readme for an example. Us using clap to make it easy to run is already an overkill IMO. Us abusing the examples to do mainnet operations was bad. 
My intuition is be to simplify the examples heavily and definitely not extend them. I tried to go a middle way so we have something working for mainnet until we have the homeserver-cli. After that, I will urge to simply the examples.

Does this make sense?

examples/rust/2-auth_flow/README.md
```diff
@@ -1,42 +1,62 @@
...
 
 ## Usage
 
-First you need to be running a local testnet Homeserver, in the root of this repo run
-
-```bash
-cargo run -p pubky-testnet
-```
+### 1a) Signin in
```
86667:
```suggestion
### 1a) Signing in
```

examples/rust/2-auth_flow/README.md
```diff
@@ -1,42 +1,62 @@
...
   --client-id my-app.example \
   --capabilities /pub/my-app/:rw
 ```
 
 Copy the Pubky Auth URL from the client output. It should use the `signin_grant` intent and include `cid` and `cpk` query parameters.
 
-Finally run the authenticator in another terminal to approve it:
+### 1b) Signing up + Signing in
+
+If you do not have signed up to the homeserver yet you can sign up and sign in at the same time with the auth flow. Start the client with `--signup`:
```
86667:
```suggestion
If you have not signed up to the homeserver yet you can sign up and sign in at the same time with the auth flow. Start the client with `--signup`:
```

86667 (commented):
> I am personally very cautious with extending the example..

I'm not against anything you say, i have no strong views on examples apart from them being usable as a cli replacement for now. My main concern is that were changing things only partially here (which may be my fault) and its clear from discussion that they need review. Especially the rust and js examples no longer mirroring eachother.

Anyway, 2 minor grammar points then lets close this out.

86667 (approved):

PR #495, opened 2026-07-13
refactor(homeserver): move PKARR retries into dedicated republisher

andrei-21:
Keep `Publisher` and `Republisher` focused on a single publish
attempt. Use `BackON` for capped exponential backoff with jitter,
retry only recoverable failures, and report attempt counts.


pubky-homeserver/src/republishers/pkarr_republisher/batch_republisher.rs
```diff
@@ -1,13 +1,21 @@
 use super::republish_summary::{RepublishResult, RepublishSummary};
-use super::republisher::RepublisherSettings;
-use super::resilient_client::{ResilientClient, ResilientClientBuilderError};
+use super::republisher::{Republisher, RepublisherSettings};
+use super::retrying_republisher::RetryingRepublisher;
```
MCarlomagno, line 3: I see some confusing in naming here, `multi_replublisher` and `retrying_republisher` sounds like the same thing with different naming, maybe we can make `multi_replublisher` something more generic and call it `backoff.rs` or somethink like that, where the responsability is to take an action and a config then do the backoff work (in this case the action would be republishing)

andrei-21: Renamed `MultiRepublisher` to `BatchRepublisher`.
Left `RetryingRepublisher` as is; the point is that it retries. *Backoff* is an implementation detail here.

MCarlomagno: Sounds good, yes after checking with AI it suggested the same, I guess I misunderstood the scope

MCarlomagno (approved):

PR #498, opened 2026-07-15
refactor(homeserver): restructure PKARR republishing

andrei-21:
Separate batch, retry, and single-key settings by ownership.
Return typed outcomes and summarize non-publish results explicitly.
Propagate PublishError directly through retry handling.


MCarlomagno (approved):
LGTM

Issue #486, opened 2026-07-09
Build: Improve build pipeline (and other pipelines?)


86667:
In `Build Artifacts` we can reuse cache and build in parallel to speed things up. 

Can we improve the other workflows, also?

PR #499, opened 2026-07-15
ci: cache npm dependencies

ok300:
This PR partially addresses #486.

It focuses on the `wasm-test` job's npm dependency path to make CI faster and more cache-friendly.

### Changes
- Configured npm download caching via `actions/setup-node@v4`:
  - `cache: npm`
  - `cache-dependency-path: pubky-sdk/bindings/js/pkg/package-lock.json`
- Added a dedicated `Install JS dependencies` step:
  - `npm ci --prefer-offline --no-audit --no-fund`
- Removed the redundant per-step `npm install` from `Build WASM`, since dependencies are now installed once in a dedicated step.

### Expected impact
Repeated CI runs can reuse npm's package download cache, reducing network fetches during `npm ci` while preserving deterministic lockfile-based installs. This should improve npm install time without relying on a `node_modules` cache that `npm ci` would remove before installing.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -225,10 +225,16 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "22"
+          cache: npm
+          cache-dependency-path: pubky-sdk/bindings/js/pkg/package-lock.json
 
       - uses: taiki-e/install-action@v2
         with:
           tool: wasm-pack
+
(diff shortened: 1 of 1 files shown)


ok300:
The `wasm-test` CI job:
- before: 10m
- after: 6m

andrei-21 (approved):
"Squash and merge" please to make the git history clean.

PR #500, opened 2026-07-16
fix(testnet/homeserver)!: DockerPostgres::shared() to share Postgres instance but isolate databases and remove `pubky-test` URL query param

86667:
The idea behind `DockerPostgres::shared()` is that the work of downloading, building and starting up Postgres shouldn't need to be done for each testnet which needs to use the Postgres instance. For example, in a test suite in which each test requires its own testnet.

The existing behaviour was that the Postgres instance and database was being shared. The fix here is to share the Postgres instance but give each test a fresh, ephemeral database.

These changes motivated a re-write of how databse urls are set, as the logic was spread across different places and behaviour not well-defined.

Previously we configured ephemeral database by a `?pubky-test=true` query parameter in the connection URL, a fragile convention that is easy to forget and adds confusion. We replaced it with a `DatabaseMode` enum that is resolved at startup based on compile-time cfg flags. In test builds you always get `EphemeralTest`, in production you always get `Direct`. The URL is now just a server address, it no longer carries behavioral intent.

We also change `database_url` from `ConnectionString` to `Option<ConnectionString>`. Test configs set it to `None`, deferring resolution to the env var or built-in default. Explicit urls (e.g. Docker Postgres) are `Some` and used as-is. Production builds throw and error if `None`.

The priority for database URL resolution in test builds is:

- Explicit URL set in config (e.g. from Docker Postgres or config.toml)
- TEST_PUBKY_CONNECTION_STRING env var
- Built-in test default

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -93,7 +93,7 @@ jobs:
 
     env:
       # Your code can read these, or just use the defaults above
-      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5432/postgres?pubky-test=true
+      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5432/postgres
 
     strategy:
       matrix:
@@ -193,7 +193,7 @@ jobs:
           --health-retries=5
     env:
       # Your code can read these, or just use the defaults above
-      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5433/postgres?pubky-test=true
+      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5433/postgres
 
     runs-on: ubuntu-latest
     steps:
--- a/pubky-homeserver/src/app_context.rs
+++ b/pubky-homeserver/src/app_context.rs
@@ -13,7 +13,7 @@ use crate::{
     observability::{Metrics, MetricsInitError},
     persistence::{
         files::{events::EventsService, FileIoError, FileService},
-        sql::{DatabaseMode, Migrator, PgEventListener, SqlDb},
+        sql::{Migrator, PgEventListener, SqlDb},
     },
     ConfigToml, DataDir,
 };
@@ -36,12 +36,9 @@ pub enum AppContextConversionError {
     /// Failed to open SQL DB.
     #[error("Failed to open SQL DB: {0}")]
     SqlDb(sqlx::Error),
-    /// Failed to resolve the database mode (e.g. invalid TEST_PUBKY_CONNECTION_STRING).
+    /// Failed to resolve the database mode (e.g. missing URL or invalid TEST_PUBKY_CONNECTION_STRING).
     #[error("Failed to resolve database mode: {0}")]
     DatabaseResolution(anyhow::Error),
-    /// No database URL was configured (production builds require an explicit URL).
-    #[error("No database_url configured. Set [general].database_url in config.toml.")]
-    NoDatabaseUrl,
     /// Failed to run migrations.
     #[error("Failed to run migrations: {0}")]
     Migrations(anyhow::Error),
@@ -157,7 +154,9 @@ impl AppContext {
             .read_or_create_keypair()
             .map_err(AppContextConversionError::Keypair)?;
 
-        let db_mode = Self::resolve_database_mode(&conf)?;
+        let db_mode = dir
+            .resolve_database_mode(&conf)
+            .map_err(AppContextConversionError::DatabaseResolution)?;
         let sql_db = SqlDb::connect(db_mode)
             .await
             .map_err(AppContextConversionError::SqlDb)?;
@@ -208,28 +207,6 @@ impl AppContext {
 }
 
 impl AppContext {
-    /// Resolve the [`DatabaseMode`] from config.
-    ///
-    /// - Production: `database_url` must be `Some`; always [`DatabaseMode::Direct`].
-    /// - Test builds: delegates to [`DatabaseMode::resolve_test`] which resolves
-    ///   `None` via env var or default, always returning [`DatabaseMode::EphemeralTest`].
-    fn resolve_database_mode(conf: &ConfigToml) -> Result<DatabaseMode, AppContextConversionError> {
-        #[cfg(any(test, feature = "testing"))]
-        {
-            DatabaseMode::resolve_test(conf.general.database_url.clone())
-                .map_err(AppContextConversionError::DatabaseResolution)
-        }
-
-        #[cfg(not(any(test, feature = "testing")))]
-        {
-            conf.general
-                .database_url
-                .clone()
-                .map(DatabaseMode::Direct)
-                .ok_or(AppContextConversionError::NoDatabaseUrl)
-        }
-    }
-
     /// Build the pkarr client builder based on the config.
     fn build_pkarr_builder_from_config(config_toml: &ConfigToml) -> pkarr::ClientBuilder {
         let mut builder = pkarr::ClientBuilder::default();
--- a/pubky-homeserver/src/app_context.rs
+++ b/pubky-homeserver/src/app_context.rs
@@ -237,9 +237,7 @@ impl AppContext {
     }
 
     /// Connect to the SQL database.
-    ///
-    /// In test builds with `?pubky-test=true`, `SqlDb::connect` automatically
-    /// creates an ephemeral database (see [`SqlDb::connect`]).
+    /// In test builds with `?pubky-test=true`, creates an ephemeral database.
     async fn connect_to_sql_db(
         config_toml: &ConfigToml,
     ) -> Result<SqlDb, AppContextConversionError> {
--- a/pubky-homeserver/src/app_context.rs
+++ b/pubky-homeserver/src/app_context.rs
@@ -35,7 +35,7 @@ pub enum AppContextConversionError {
     Keypair(anyhow::Error),
     /// Failed to open SQL DB.
     #[error("Failed to open SQL DB: {0}")]
-    SqlDb(sqlx::Error),
+    SqlDb(crate::persistence::sql::SqlDbConnectError),
     /// Failed to run migrations.
     #[error("Failed to run migrations: {0}")]
     Migrations(anyhow::Error),
@@ -151,7 +151,9 @@ impl AppContext {
             .read_or_create_keypair()
             .map_err(AppContextConversionError::Keypair)?;
 
-        let sql_db = Self::connect_to_sql_db(&conf).await?;
+        let sql_db = SqlDb::connect(conf.general.database_url.clone())
+            .await
+            .map_err(AppContextConversionError::SqlDb)?;
         Migrator::new(&sql_db)
             .run()
             .await
@@ -203,7 +205,14 @@ impl AppContext {
     fn build_pkarr_builder_from_config(config_toml: &ConfigToml) -> pkarr::ClientBuilder {
         let mut builder = pkarr::ClientBuilder::default();
         #[cfg(any(test, feature = "testing"))]
-        if config_toml.general.database_url.is_test_db() {
+        // None (no explicit URL) or a `?pubky-test=true` URL both indicate a test
+        // environment where we must avoid contacting the public DHT.
+        if config_toml
+            .general
+            .database_url
+            .as_ref()
+            .is_none_or(|url| url.is_test_db())
+        {
             builder
                 .no_default_network()
                 // Keep the client buildable without contacting the public DHT.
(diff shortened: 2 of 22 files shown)


pubky-homeserver/src/app_context.rs
```diff
@@ -163,11 +169,36 @@ impl AppContext {
...
 impl AppContext {
+    /// Resolve the [`DatabaseMode`] from config.
+    ///
+    /// - Production: `database_url` must be `Some`; always [`DatabaseMode::Direct`].
+    /// - Test builds: delegates to [`DatabaseMode::resolve_test`] which resolves
+    ///   `None` via env var or default, always returning [`DatabaseMode::EphemeralTest`].
+    fn resolve_database_mode(conf: &ConfigToml) -> Result<DatabaseMode, AppContextConversionError> {
+        #[cfg(any(test, feature = "testing"))]
+        {
+            Ok(DatabaseMode::resolve_test(
```
MCarlomagno: I think `Persistent` mode should use `DatabaseMode::Direct`, `pubky-testnet` always enables the testing feature, so persistent mode also gets an ephemeral database, and so each restart loses the db state and leaves another database behind.

.github/workflows/pr-check.yml
```diff
@@ -93,7 +93,7 @@ jobs:
 
     env:
       # Your code can read these, or just use the defaults above
-      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5432/postgres?pubky-test=true
+      TEST_PUBKY_CONNECTION_STRING: postgres://test_user:test_pass@localhost:5432/postgres
```
ok300, line 96: NIT: Not sure if intentional, but the `?pubky-test=true` suffix is still present in:
- `examples/rust/README.md`
- `pubky-testnet/README.md`

Technically both should still work, because the change is backwards compatible.

pubky-homeserver/src/persistence/sql/database_mode.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    #[test]
+    fn resolve_falls_back_to_default() {
+        let result = DatabaseMode::resolve_test_inner(None, None);
+        assert_eq!(result.connection_string().as_str(), DEFAULT_TEST_SERVER);
+        assert!(matches!(result, DatabaseMode::EphemeralTest(_)));
+    }
+
+    #[test]
+    fn resolve_invalid_env_var_falls_back_to_default() {
```
ok300: NIT: maybe this should fail instead?

If someone sets an env variable, but it's silently ignored, it might get tricky to troubleshoot issues later on.

pubky-homeserver/src/app_context.rs
```diff
@@ -163,11 +169,36 @@ impl AppContext {
...
+                conf.general.database_url.clone(),
+            ))
+        }
+
+        #[cfg(not(any(test, feature = "testing")))]
```
ok300: > `#[cfg(not(any(test, feature = "testing")))]`

This essentially means: when neither `cfg(test)` nor `feature = "testing"` is enabled.

However, in `pubky-testnet/Cargo.toml:27` , the `pubky-homeserver` dependency is included with the "testing" feature enabled.

So this means `AppContext::resolve_database_mode` will fallback to L178-183 above

```rust
        #[cfg(any(test, feature = "testing"))]
        {
            Ok(DatabaseMode::resolve_test(  
                conf.general.database_url.clone(),
            ))
        }
```

which means every `pubky-testnet` call will use an ephemeral DB URL.

Assuming I understand this correclty -- is this intended behavior?

I see `pubky-testnet` has a `Command::Persist`, which looks like it would also use an ephemeral DB URL, and therefore not persist the DB, no?

ok300: Basically same comment as https://github.com/pubky/pubky-homeserver/pull/500#discussion_r3753097030 , marking as resolved since its a duplicate.

86667:
@MCarlomagno @ok300 

Thanks both for review and finding that issue. My bad - I must have made the removal of `?pubky-testnet=true` change after testing the persistent `StaticTestnet`, then not re-tested after.

I have removed the "ephemeral db" decision from the testing cfg and instead drive it off of `DataDir`. Ie, If a data dir is persisted then so is the db, if datat dir is not persisted then neither is the db. 

Im going to look into refactoring the whole `DataDir` trait with `MockDataDir` and `PersistentDataDir` implementors. I dont find it a clean abstraction. I didnt want to make this PR any larger though so used what i had.

.github/workflows/pr-check.yml
86667: Fixed

pubky-homeserver/src/persistence/sql/database_mode.rs
86667: True, good call. updated

pubky-homeserver/src/app_context.rs
```diff
@@ -203,7 +211,9 @@ impl AppContext {
     fn build_pkarr_builder_from_config(config_toml: &ConfigToml) -> pkarr::ClientBuilder {
         let mut builder = pkarr::ClientBuilder::default();
         #[cfg(any(test, feature = "testing"))]
-        if config_toml.general.database_url.is_test_db() {
+        // In test builds, no explicit database_url means we're in a test environment
+        // where we must avoid contacting the public DHT.
+        if config_toml.general.database_url.is_none() {
```
ok300, line 216: > we must avoid contacting the public DHT

This only happens if the `database_url` is unset. So if a `MockDataDir` is created with an explicit `database_url`:

```rust
let mut config = ConfigToml::minimal_test_config();
config.general.database_url = Some(postgres_url);

let dir = MockDataDir::new(config, None)?;
HomeserverApp::start_with_mock_data_dir(dir).await?;
```

then the test can publish test records to the public DHT, and / or fail the test if the environment is sandboxed and outbound connections are denied.

I guess in our test suites we wouldn't do that, but the `README` documents how library users might write their own tests, so others could inadvertently trigger this edge-case:

https://github.com/pubky/pubky-homeserver/blob/b4100e7849d2db18cc2c8800ad57f6aa0b9f724a/pubky-homeserver/README.md?plain=1#L77-L79

One solution could be to pass `db_mode: &DatabaseMode` as an extra arg to `build_pkarr_builder_from_config` then internally check it directly:

```diff
- if config_toml.general.database_url.is_none()
+ if matches!(db_mode, DatabaseMode::EphemeralTest(_))
```

pubky-testnet/src/testnet.rs
```diff
@@ -39,8 +39,7 @@ impl Testnet {
             http_relays: vec![],
             homeservers: vec![],
             temp_dirs: vec![],
-            postgres_connection_string: Self::extract_postgres_connection_string_from_env_variable(
-            ),
+            postgres_connection_string: None,
```
ok300, line 42: This change means the persistent command from `pubky-testnet/README.md`

```sh
TEST_PUBKY_CONNECTION_STRING='postgres://postgres:postgres@localhost:5432/postgres' \
  cargo run -p pubky-testnet -- persist ./my-testnet-data
```

now ignores `TEST_PUBKY_CONNECTION_STRING` and instead uses the generated config default

```toml
 database_url = "postgres://localhost:5432/pubky_homeserver"
```

from `config.sample.toml`, which doesn't have the user/pass or the right DB name for the Docker DB setup.

pubky-homeserver/src/persistence/sql/sql_db.rs
```diff
@@ -83,87 +89,56 @@ impl Drop for TestDbDropper {
...
+        let admin_con = Self::connect_inner(admin_con_string).await?;
+        let test_db_name = format!("pubky_test_{}", Uuid::new_v4().as_simple());
+        let query = format!("CREATE DATABASE \"{}\"", test_db_name);
+        sqlx::query(&query).execute(admin_con.pool()).await?;
 
-        // If no connection string is passed, use the default test connection string.
-        ConnectionString::new(DEFAULT_TEST_CONNECTION_STRING)
-            .expect("Default test connection string is valid")
+        let mut test_db_url = admin_con_string.clone();
+        test_db_url.set_database_name(&test_db_name);
```
ok300, line 132: `set_database_name()` only modifies the URL path.

If the `admin_con_string` contains a `?dbname=` query parameter (which is valid PostgreSQL connection string syntax), the resulting URL after `set_database_name()` will have:
- a new path (e.g., `/pubky_test_xxx`)
- but the original `?dbname=postgres` query parameter remains unchanged

Example:

```
// Input:
postgres://user:pass@localhost:5432/postgres?dbname=postgres

// After set_database_name("pubky_test_abc123"):
postgres://user:pass@localhost:5432/pubky_test_abc123?dbname=postgres
```

SQLx 0.8 parses the path first and then parses query parameters. When it sees `dbname=postgres`, that query parameter overrides `pubky_test_abc123`.

pubky-homeserver/src/persistence/sql/database_mode.rs
```diff
@@ -0,0 +1,115 @@
...
+/// and "create a fresh ephemeral database for this test" explicit.
+#[derive(Debug, Clone)]
+pub enum DatabaseMode {
+    /// Connect directly to the database identified by the URL.
+    /// Used in production and persistent testnets.
+    Direct(ConnectionString),
+
+    /// Create an ephemeral `pubky_test_{uuid}` database on the server
+    /// identified by the URL, then connect to it. The database is dropped
+    /// when the [`SqlDb`](super::SqlDb) is dropped.
```
ok300: > The database is dropped when the [`SqlDb`](super::SqlDb) is dropped.

Dropping `SqlDb` only queues cleanup; deletion requires `#[pubky_testnet::test]` or an explicit `drop_test_databases()` call. Callers without either leak databases on external PostgreSQL instances.

Simplest fix is probably to change these doc lines to say that dropping registers the DB for cleanup and that callers must use the `#[pubky_testnet::test]` macro or call `drop_test_databases()`. Ideally the `README` examples[^1] should include or mention that too.

[^1]: https://github.com/pubky/pubky-homeserver/blob/b4100e7849d2db18cc2c8800ad57f6aa0b9f724a/pubky-testnet/README.md?plain=1#L146-L174

pubky-homeserver/src/data_directory/data_dir.rs
```diff
@@ -20,6 +21,13 @@ pub trait DataDir: std::fmt::Debug + DynClone + Send + Sync {
     /// Reads the secret file from the data directory.
     /// Creates a new secret file if it doesn't exist.
     fn read_or_create_keypair(&self) -> anyhow::Result<pubky_common::crypto::Keypair>;
+
+    /// Resolve how the homeserver should connect to its database.
+    ///
+    /// Each implementation decides its own strategy:
+    /// - [`PersistentDataDir`](super::PersistentDataDir): [`DatabaseMode::Direct`] — connects to the configured URL.
+    /// - [`MockDataDir`](super::MockDataDir): [`DatabaseMode::EphemeralTest`] — creates a temporary database.
```
ok300: L29 links unconditionally to `MockDataDir` and `DatabaseMode` which are not normally visible by default as they are cfg-gated behind

```
#[cfg(any(test, feature = "testing"))]
```

None of those flags are set when generating default rustdocs, which means running

```sh
cargo doc -p pubky-homeserver --no-deps
```

fails with an error.

Simplest solution would be to remove these 3 rustdoc lines from here and instead mention the relevant`DatabaseMode` for the relevant impls, for example:

```diff
  /// Resolve how the homeserver should connect to its database.
  ///
- /// Each implementation decides its own strategy:
- /// - [`PersistentDataDir`](super::PersistentDataDir): [`DatabaseMode::Direct`] — connects to the configured URL.
- /// - [`MockDataDir`](super::MockDataDir): [`DatabaseMode::EphemeralTest`] — creates a temporary database.
+ /// Each implementation selects the appropriate database lifecycle.
  fn resolve_database_mode(&self, conf: &ConfigToml) -> anyhow::Result<DatabaseMode>;
```

then

```diff
impl DataDir for MockDataDir {
+  /// Creates a temporary database with [`DatabaseMode::EphemeralTest`]
  fn resolve_database_mode()
}
```

and

```diff
impl DataDir for PersistentDataDir {
+    /// Connects to the configured URL with [`crate::persistence::sql::DatabaseMode::Direct`]
    fn resolve_database_mode(
}
```

pubky-homeserver/src/app_context.rs
86667: This is equal in fragility as before, when a provide database_url without `?pubky-test=true` would not isolate pkarr. I'll fix this properly when i refactor the whole data dir/db/file system config situation, if that okay with you.

pubky-testnet/src/testnet.rs
86667: this also is pre-existing and also will be addressed in the follow-up refactor

pubky-homeserver/src/persistence/sql/sql_db.rs
86667: Fixed by removing dbname param

pubky-homeserver/src/persistence/sql/database_mode.rs
86667: Updated doc here and also added missing `[crate::test]`!

pubky-homeserver/src/data_directory/data_dir.rs
86667: Updated.

pubky-homeserver/src/persistence/sql/database_mode.rs
86667: I'll also look into making it clearer where/when this is needed, since it was missed on some tests already.

ok300 (approved):

MCarlomagno (approved):
LGTM

PR #501, opened 2026-07-16
fix(homeserver): private events on session revocation

### LINKED ISSUE #492: Stop private event streams after session revocation
## Description

Private `/events-stream` subscriptions currently authorize the requested private paths when the SSE connection is opened. After that, live events are filtered using the already-computed allowed path list. If the session or grant used to open the stream is revoked while the connection is still alive, the stream may continue receiving matching private events until the client disconnects.

Update private live event streams so revocation is respected for already-open connections. A revoked cookie session or grant session should no longer receive private SSE events after revocation. Public-only streams should continue to work anonymously as they do today.

## Acceptance criteria

- [ ] Add a regression/e2e test that opens a live private event stream, revokes the session/grant, writes a new matching private event through another valid session, and verifies the revoked stream does not receive it.
- [ ] Active private SSE streams re-check session validity before yielding private live events, or otherwise close promptly after revocation.
- [ ] Cookie-session revocation and grant-session revocation are both considered, or any intentionally unsupported case is documented.
- [ ] Public event streams remain unaffected and do not require session revalidation.
- [ ] The implementation does not widen private path access beyond the paths authorized at subscription time.
- [ ] Add comments/docs explaining the chosen revocation behavior for long-lived SSE streams.

MCarlomagno:
Closes #492.

### Summary

Private `/events-stream` subscriptions authorize their private paths when the connection opens, then keep using that authorization for the lifetime of the SSE stream. Before this change, revoking the cookie session or grant that opened the stream did not stop an already open connection from receiving subsequent private events.

This adds committed postgresql revocation notifications (thru LISTEN/NOTIFY) and local SSE signal so every homeserver instance closes affected private streams promptly after cookie signout or grant revocation.

Behavior:

- Cookie signout and grant revocation publish a Postgres `NOTIFY` in the same transaction as the database mutation.
- Each homeserver instance listens for revocations and forwards them to its local private SSE streams.
- A matching revocation closes the private stream before it can receive later matching events.
- If the auth listener disconnects or cannot guarantee no notification was missed, all local private streams close and new private subscriptions are rejected until reconnection succeeds.
- A lagged local revocation receiver revalidates only its own credential instead of disconnecting unrelated private streams.
- Public only streams do not subscribe to auth revocations and remain unaffected.
- Private stream path authorization remains fixed at subscription time, this does not widen access beyond the originally authorized paths.
- Credential expiry and bearer rotation are enforced when opening or reconnecting a stream, they do not terminate an open stream.

### Acceptance criteria

- [x] A live private stream opened with a grant closes after that grant is revoked and does not receive a later matching private event.
- [x] A live private stream opened with a cookie closes after that cookie session signs out and does not receive a later matching private event.
- [x] Cookie and grant revocations propagate across homeserver instances through PostgreSQL.
- [x] Auth-listener disconnection fails closed for existing and new private streams until reconnection.
- [x] Public event streams remain open and anonymous after unrelated auth revocations.
- [x] Private path access remains limited to the paths authorized when the subscription opened.
- [x] The long-lived SSE revocation and reconnect behavior is documented.

### DIFF
--- a/pubky-homeserver/src/client_server/auth/grant/service.rs
+++ b/pubky-homeserver/src/client_server/auth/grant/service.rs
@@ -162,17 +162,6 @@ impl GrantAuthService {
         self.revoke_grant(&session.grant_id).await
     }
 
-    /// Recheck the grant status immediately before a private long-lived stream
-    /// starts. Subsequent revocation is handled by the notification listener.
-    pub(crate) async fn validate_active_session(
-        &self,
-        session: &GrantSession,
-    ) -> Result<(), AuthServiceError> {
-        self.validate_active_grant_session(session.token_expires_at, &session.grant_id)
-            .await?;
-        Ok(())
-    }
-
     /// Resolve an opaque bearer into a `GrantSession`.
     ///
     /// Hashes the bearer, looks up the matching session row, loads the
@@ -222,9 +211,11 @@ impl GrantAuthService {
 
     // ── Private helpers ─────────────────────────────────────────────────
 
-    /// Shared active-session validation used both when resolving a bearer and
-    /// immediately before a long-lived private stream begins.
-    async fn validate_active_grant_session(
+    /// Validate a grant session and return its active backing grant.
+    ///
+    /// Used both when resolving a bearer and immediately before a private
+    /// long-lived stream begins.
+    pub(crate) async fn validate_active_grant_session(
         &self,
         token_expires_at: u64,
         grant_id: &GrantId,
--- a/pubky-homeserver/src/client_server/auth/grant/service.rs
+++ b/pubky-homeserver/src/client_server/auth/grant/service.rs
@@ -27,7 +27,7 @@ use super::crypto::{
 };
 use super::persistence::{
     grant::{GrantEntity, GrantRepository, NewGrant},
-    grant_session::{GrantSessionRepository, NewGrantSession},
+    grant_session::{GrantSessionEntity, GrantSessionRepository, NewGrantSession},
     pop_nonce::{PopNonceError, PopNonceRepository},
 };
 use super::service_error::AuthServiceError;
@@ -177,15 +177,14 @@ impl GrantAuthService {
             AuthServiceError::SessionNotFound,
         )?;
 
-        let grant = self
-            .validate_active_grant_session(session.expires_at as u64, &session.grant_id)
-            .await?;
+        let grant = self.validate_active_grant_session_entity(&session).await?;
 
         Ok(GrantSession {
             user_key: grant.user_pubkey.clone(),
             capabilities: grant.capabilities,
             grant_id: session.grant_id,
             token_expires_at: session.expires_at as u64,
+            token_hash: session.token_hash,
         })
     }
 
@@ -211,21 +210,38 @@ impl GrantAuthService {
 
     // ── Private helpers ─────────────────────────────────────────────────
 
-    /// Validate a grant session and return its active backing grant.
-    ///
-    /// Used both when resolving a bearer and immediately before a private
-    /// long-lived stream begins.
+    /// Recheck that this exact bearer session row still exists before opening
+    /// a private long-lived stream.
     pub(crate) async fn validate_active_grant_session(
         &self,
-        token_expires_at: u64,
-        grant_id: &GrantId,
+        session: &GrantSession,
+    ) -> Result<GrantEntity, AuthServiceError> {
+        let persisted = map_not_found(
+            GrantSessionRepository::get_by_token_hash(
+                &session.token_hash,
+                &mut self.sql_db.pool().into(),
+            )
+            .await,
+            AuthServiceError::SessionNotFound,
+        )?;
+
+        if persisted.grant_id != session.grant_id {
+            return Err(AuthServiceError::SessionNotFound);
+        }
+
+        self.validate_active_grant_session_entity(&persisted).await
+    }
+
+    async fn validate_active_grant_session_entity(
+        &self,
+        session: &GrantSessionEntity,
     ) -> Result<GrantEntity, AuthServiceError> {
         let now = Utc::now().timestamp();
-        if token_expires_at <= now as u64 {
+        if session.expires_at <= now {
             return Err(AuthServiceError::SessionExpired);
         }
 
-        let grant = self.get_grant(grant_id).await?;
+        let grant = self.get_grant(&session.grant_id).await?;
         grant.require_active(now)?;
         Ok(grant)
     }
@@ -864,6 +880,41 @@ mod tests {
 
     #[tokio::test]
     #[pubky_test_utils::test]
+    async fn validate_active_grant_session_rejects_rotated_bearer() {
+        let service = test_service().await;
+        let (user_kp, _) = create_test_user(&service).await;
+        let client_kp = Keypair::random();
+        let (grant_jws, pop_jws, raw_grant) =
+            sign_grant(&user_kp, &client_kp, &service.homeserver_public_key());
+
+        let response = service
+            .create_grant_session(&grant_jws, &pop_jws)
+            .await
+            .unwrap();
+        let session = service
+            .resolve_grant_session_by_bearer(&SessionBearer::parse(&response.token).unwrap())
+            .await
+            .unwrap();
+
+        service
+            .validate_active_grant_session(&session)
+            .await
+            .unwrap();
+
+        service
+            .mint_session(&raw_grant, &mut service.sql_db.pool().into())
+            .await
+            .unwrap();
+
+        let err = service
+            .validate_active_grant_session(&session)
+            .await
+            .unwrap_err();
+        assert!(matches!(err, AuthServiceError::SessionNotFound));
+    }
+
+    #[tokio::test]
+    #[pubky_test_utils::test]
     async fn resolve_grant_session_after_revoke() {
         let service = test_service().await;
         let (user_kp, _) = create_test_user(&service).await;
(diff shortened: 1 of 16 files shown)


pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -433,8 +514,33 @@ pub async fn feed_stream(
...
+                                tracing::warn!("auth revocation receiver lagged and stream session is no longer valid; closing private event stream");
+                                return;
+                            }
+                            Err(broadcast::error::RecvError::Closed) => {
+                                tracing::warn!("auth revocation receiver closed; closing private event stream");
+                                return;
+                            }
+                        }
+                    }
+                    event = rx.recv() => match event {
```
andrei-21: I think making `next_revocation()`, the body of `match revocation`, `stream_auth_is_valid()` and `pending_revocation()` as methods of `StreamAuth` will make the code easier and more robust. Also moving to it's own file.

pubky-homeserver/src/client_server/auth/revocation.rs
```diff
@@ -0,0 +1,446 @@
...
+        let listener = Self::connect(pool).await?;
+        service.mark_healthy();
+
+        let cancel = CancellationToken::new();
+        let handle = {
+            let pool = pool.clone();
+            let cancel = cancel.clone();
+            tokio::spawn(async move {
+                Self::listen_loop(pool, service, listener, cancel).await;
+            })
```
andrei-21: Minor:
```suggestion
            tokio::spawn(Self::listen_loop(pool, service, listener, cancel))
```

pubky-homeserver/src/client_server/auth/revocation.rs
```diff
@@ -0,0 +1,446 @@
...
+        }
+    }
+}
+
+impl Drop for PgAuthRevocationListener {
+    fn drop(&mut self) {
+        self.cancel.cancel();
+        if let Some(handle) = self.handle.take() {
+            handle.abort();
+        }
```
andrei-21: Cancellation is almost useless if you abort the task immediately. We need to wait a while to let the task exit gracefully or just drop the cancellation logic.

andrei-21 (changes requested):

pubky-homeserver/src/client_server/auth/revocation.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+/// Local fan-out service for committed authentication revocations.
+#[derive(Clone, Debug)]
+pub(crate) struct AuthRevocationService {
+    notifications: broadcast::Sender<AuthRevocation>,
+    listener_healthy: Arc<AtomicBool>,
+}
+
+impl AuthRevocationService {
```
andrei-21: I think the revocation listener would benefit from a watchdog/actor design.

`AuthRevocationService::subscribe()` would send a subscription request to a watchdog. The watchdog forwards it to the current listener actor. If that actor has stopped—indicated by a closed mailbox—the watchdog creates a new actor and retries the request.
The actor would own both the PostgreSQL listener and the broadcast channel. It would process subscription requests and PostgreSQL notifications in the same event loop. If the connection is lost or another fatal listener error occurs, the actor exits and drops its broadcast sender. This closes every existing subscriber, preserving the required fail-closed behavior because notifications may have been missed.
The next subscription request causes the watchdog to create a fresh actor, effectively reconnecting. Existing subscribers are never migrated to the new actor; they must reconnect themselves.

This should remove the explicit health flag, `AuthRevocation::All` signal, reconnection loop, cancellation token, and manual task abortion. Task shutdown can instead follow channel ownership: dropping the service closes the watchdog mailbox, which drops the actor mailbox and lets both tasks exit naturally.

pubky-homeserver/src/client_server/routes/events.rs
```diff
@@ -433,8 +471,21 @@ pub async fn feed_stream(
...
+                        match signal {
+                            RevocationSignal::Continue => {}
+                            RevocationSignal::Close => return,
+                            RevocationSignal::Revalidate => {
+                                if !stream_auth.revalidate_session(&state.auth_state).await {
+                                    return;
+                                }
+                            }
+                        }
+                    }
```
andrei-21: We can even do:
```rust
  check = stream_auth.next_check(&state.auth_state) => {
      if !check.await {
          return;
      }
  }
```

pubky-homeserver/src/client_server/auth/revocation.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+/// Serializable form of an authentication revocation.
+///
+/// Cookie secrets and bearer tokens are deliberately not present in this
+/// payload. Postgres channel consumers only need stable database identifiers.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(tag = "kind", content = "id", rename_all = "snake_case")]
+enum WireAuthRevocation {
```
andrei-21: Can't `WireAuthRevocation` and `AuthRevocation` be merged into one struct?

pubky-homeserver/src/client_server/auth/revocation/listener.rs
```diff
@@ -0,0 +1,378 @@
...
+impl AuthRevocationService {
+    /// Start the supervisor with a listener already receiving, so the app never
+    /// starts serving private streams without its revocation feed.
+    pub(crate) async fn start(pool: &PgPool) -> Result<Self, sqlx::Error> {
+        let supervisor = Supervisor {
+            pool: pool.clone(),
+            actor: Some(ListenerActor::spawn(pool).await?),
+            replacement_cooldown: ReplacementCooldown::default(),
+        };
+        let (requests, mailbox) = mpsc::channel(SUPERVISOR_MAILBOX_CAPACITY);
```
andrei-21: I do not think that we need that complication, we can simply wrap the supervisor like `Arc<tokio::sync::Mutex<Supervisor>>` and call methods directly.

pubky-homeserver/src/client_server/auth/revocation/listener.rs
```diff
@@ -0,0 +1,378 @@
...
+        if let Some(handle) = self.actor.as_ref() {
+            if let Some(receiver) = handle.subscribe().await {
+                return Ok(receiver);
+            }
+            self.actor = None;
+        }
+        self.replace().await
+    }
+
+    async fn replace(&mut self) -> SubscriptionResult {
```
andrei-21: This method is a bit confusing, it not only reconnects, it also subscribes. I think it should only replace.

pubky-homeserver/src/client_server/auth/revocation/listener.rs
```diff
@@ -0,0 +1,378 @@
...
+        tracing::info!("started a replacement auth revocation listener");
+        self.actor = Some(replacement);
+        Ok(receiver)
+    }
+}
+
+struct ListenerActor {
+    listener: PgListener,
+    notifications: broadcast::Sender<AuthRevocation>,
+    mailbox: mpsc::Receiver<ActorRequest>,
```
andrei-21: Usually senders and receivers are named like `tx`, `rx` and prefixed with something meaningful if necessary like `notifications_tx`.
See examples of https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html.

pubky-homeserver/src/client_server/auth/stream_auth.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        loop {
+            match revocation_rx.try_recv() {
+                Ok(revocation) if revocation.matches(session) => return RevocationSignal::Close,
+                Ok(_) => continue,
+                Err(broadcast::error::TryRecvError::Empty) => return RevocationSignal::Continue,
+                // Revalidate this stream's own credential instead of
+                // disconnecting every private stream after an unrelated burst.
+                Err(broadcast::error::TryRecvError::Lagged(_)) => {
+                    return RevocationSignal::Revalidate
```
andrei-21: Consider resubscribing here (and in `next_revocation_signal()`) to drop remaining notifications, since we are going to revalidate anyway:
```rust
				*revocation_rx = revocation_rx.resubscribe();
```

pubky-homeserver/src/client_server/auth/revocation/listener.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+impl ListenerActor {
```
andrei-21, line 113: Consider the following idea:
Remove the request channel between `Supervisor` and `ListenerActor`. Let the actor own the only strong broadcast sender, while the supervisor holds a `WeakSender` for creating subscriptions directly. Use a one-shot lifetime channel to stop the actor when its supervisor handle is dropped. If the actor fails, dropping its sender closes all subscribers; the next subscription detects the dead weak sender and creates a replacement actor.
`ListenerActorHandle` will simply becomes:
```rust
  struct ListenerActorHandle {
      notifications: broadcast::WeakSender<AuthRevocation>,
      _lifetime: oneshot::Sender<()>,
  }

  impl ListenerActorHandle {
      fn subscribe(&self) -> Option<RevocationReceiver> {
          self.notifications
              .upgrade()
              .map(|notifications| notifications.subscribe())
      }
  }
```

pubky-homeserver/src/client_server/auth/grant/service.rs
```diff
@@ -215,6 +211,25 @@ impl GrantAuthService {
...
+        token_expires_at: u64,
+        grant_id: &GrantId,
+    ) -> Result<GrantEntity, AuthServiceError> {
+        let now = Utc::now().timestamp();
+        if token_expires_at <= now as u64 {
+            return Err(AuthServiceError::SessionExpired);
+        }
+
+        let grant = self.get_grant(grant_id).await?;
+        grant.require_active(now)?;
```
andrei-21: Is it relevant what AI complains about?

> The revocation mechanism has ordering and validation races that can allow a revoked or rotated credential to receive private stream access. 
>
> [P2] Revalidate the bearer row during stream setup.
> If the bearer is rotated after middleware resolves it but before this final validation, `replace_for_grant` has deleted that exact session row, yet this method still succeeds because it checks only the copied expiry and backing grant. The now-invalid bearer can therefore open a private stream despite the documented opening-time rotation check; retain a bearer session identifier/hash and confirm that row still exists here.

MCarlomagno: Yes I think it's a cheap fix

andrei-21 (approved):

PR #502, opened 2026-07-16
fix(pkarr)!: adapt to the new client API

andrei-21:
Use explicit resolve policies and operation-specific errors.
Consume publish node counts directly and update testnet builders.
Configure test clients and relays with the testnet report policy.

BREAKING CHANGE: replace PkarrError::Query with PkarrError::Resolve.


pubky-common/src/auth/auth_token.rs
```diff
@@ -13,10 +13,65 @@ const CURRENT_VERSION: u8 = 0;
...
+        deserializer.deserialize_tuple(Signature::BYTE_SIZE, SignatureVisitor)
+    }
+}
+
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 /// Implementation of the [Pubky Auth spec](https://pubky.github.io/pubky-core/spec/auth.html).
 pub struct AuthToken {
     /// Signature over the token.
+    #[serde(with = "signature_serde")]
     signature: Signature,
```
andrei-21, line 75: @SeverinAlexB @86667
The Ed25519 upgrade changed how signatures are serialized:

- Old format: exactly 64 signature bytes.
- New format: a length byte followed by 64 signature bytes.

That extra byte shifted the AuthToken version from byte 74 to 75, causing verification to read part of `PUBKY:AUTH` as the version and return `UnknownVersion`.

The latest commit adds custom Serde code that always encodes and decodes signatures using the old fixed 64-byte format. This preserves compatibility with existing AuthTokens while allowing the dependency upgrade.

SeverinAlexB: Hmm, will this lead to an incompatibility when the signer is v3 generating the 65 bytes and the client (PubkyApp) is still on v3-rc?

andrei-21: Yes, if we do not provide this custom implementation.

pubky-sdk/src/actors/pkdns.rs
```diff
@@ -148,7 +148,12 @@ impl Pkdns {
             "Resolving homeserver for public key {} via PKARR",
             user_public_key
         );
-        let packet = self.client.pkarr().resolve(user_public_key).await?;
+        let packet = self
+            .client
+            .pkarr()
+            .resolve(user_public_key, ResolvePolicy::CacheFirst)
+            .await
+            .ok()?;
```
SeverinAlexB, line 156: I am actually wondering if we need to pass the ResolvePolicy as an argument here. When one requests the current homeserver, it's not clear if one wants a cached one. Especially when one just changed the homeserver due to a homeserver switch or so.

Not sure if this is part of this PR. Probably not.

pubky-sdk/src/actors/pkdns.rs
```diff
@@ -232,14 +237,23 @@ impl Pkdns {
...
             info,
             "Preparing to publish homeserver record for {} with mode {:?}",
             pubky,
             mode
         );
-        let existing = self.client.pkarr().resolve_most_recent(&pubky).await;
+        let resolve_policy = match mode {
+            PublishMode::IfStale => ResolvePolicy::CacheFirst,
+            PublishMode::Force => ResolvePolicy::NetworkOnly,
+        };
```
SeverinAlexB: Shouldn't this always be NetworkOnly? So far, it was always `.resolve_most_recent()` so what is the logic behind changing this to `CacheFirst`?

Edit AI comment:

High: conditional publishing can overwrite a newer record from another client  
pubky-sdk/src/actors/pkdns.rs:247-269 uses CacheFirst for IfStale. PKARR considers cache freshness using last_seen, while the SDK checks staleness using the packet’s signed timestamp. An old packet recently placed in cache can avoid network resolution, then be re-signed with a current timestamp and overwrite a newer network packet, reverting its homeserver and unrelated records. Once a cached packet is considered stale, resolve with NetworkOnly before rebuilding it.

Edit:
How does this work now actually. Republishing in the homeserver is still broken (gonna be fixed soon). So far, the relay also returned expired old packets. This way, users records could still be resolved. With pkarr v7, this is not the case anymore. So the homeserver will not get the expired packets anymore from the relay. Is this function still able to republish it even though it's not on the network/relay?

pubky-homeserver/src/republishers/pkarr_republisher/republisher.rs
```diff
@@ -64,8 +64,16 @@ impl Republisher {
...
+        let packet = match self
+            .client
+            .resolve(public_key, ResolvePolicy::NetworkOnly)
+            .await
+        {
+            Ok(packet) => packet,
+            Err(pkarr::errors::ResolveError::InvalidSignedPacket { .. }) => {
+                return Ok(RepublishOutcome::InvalidSignedPacket);
+            }
+            Err(_) => return Ok(RepublishOutcome::Missing),
```
SeverinAlexB, line 76: AI review:

Medium: homeserver republishing treats DHT outages as missing records  
pubky-homeserver/src/republishers/pkarr_republisher/republisher.rs:67-76 maps every resolution error except InvalidSignedPacket to RepublishOutcome::Missing. Errors such as NoResponses and NoDhtNodesQueried therefore bypass the retry mechanism in retrying_republisher.rs:80-87. Only NotFound should be classified as missing; operational errors should propagate and be retried or counted as failures.

SeverinAlexB (changes requested):
Some comments.

Do you usually ask AI to review first before you ask somebody else? If not, I strongly suggest to do so. GPT5.6 Sol is getting really good at finding nuanced issues.

pubky-homeserver/src/republishers/pkarr_republisher/republisher.rs
andrei-21: That is fair, but in this PR I am just trying to make it work with the new pkarr version, I am not trying to fix republishing yet.

pubky-sdk/src/actors/pkdns.rs
andrei-21: We can discuss later, but "current homeserver" naturally means a record within TTL. Again, I am not trying to fix problems, only upgrade pkarr.

pubky-sdk/src/actors/pkdns.rs
andrei-21: Ok, I refactored it into to steps, one with `CacheOnly` and the other one with `NetworkOnly`.

andrei-21:
> Do you usually ask AI to review first before you ask somebody else?

Sure I do.

pubky-homeserver/src/republishers/pkarr_republisher/republisher.rs
SeverinAlexB: The reason I am mentioning this is because I am worried we are introducing a new bug because the relay doesn't return expired packets anymore compared to before. I am not trying to make you fix existing bugs but not returning expired packets that are not on the DHT anymore does change how the system works.

Basically, there relays acted as a longterm packet storage for everybody. Now, removing this behaviour might lead to packet loss.

And I am not sure if this needs to be addressed here or in the republisher. Just trying to mention my worries.

SeverinAlexB (approved):
Approved as discussed in Slack

PR #503, opened 2026-07-17
docs(homeserver): Deploy guide

86667:


docs/DEPLOY.md
```diff
@@ -0,0 +1,207 @@
...
+| 6289 | Metrics — internal monitoring only. |
+
+## Configure the Homeserver
+
+Edit `~/.pubky/config.toml` with the following settings:
+
+```toml
+[drive]
+# Listen on all interfaces so Pubky TLS is reachable from the internet
+pubky_listen_socket = "0.0.0.0:6287"
```
SeverinAlexB, line 72: To make the guide consistent, should the Pubky TLS socket also be proxied through Caddy? I think having it behind the proxy is better than exposing the homeserver directly. This allows to use Caddy functions for Pubky TLS too

docs/DEPLOY.md
```diff
@@ -0,0 +1,207 @@
...
+# Pubky TLS clients can connect directly to the homeserver.
+public_ip = "your_ip"
+
+# Your domain name for HTTPS browser access (via the reverse proxy).
+icann_domain = "your_domain.com"
+```
+
+Replace `your_ip` with the public IP of the machine running the homeserver. You can find it with `curl -4 ifconfig.me`.
+
+> **Important:** Ensure your server has a **static (reserved) public IP**. On most cloud providers the default external IP is ephemeral. If your ip changes then both your DNS `A` record and the PKARR record (which embeds `public_ip`) will silently point at a dead address.
```
SeverinAlexB: Not sure if most are emphemeral. I would drop this sentence. Also doesn't add anything to the guide IMO.

```suggestion
> **Important:** Ensure your server has a **static (reserved) public IP**. If your ip changes then both your DNS `A` record and the PKARR record (which embeds `public_ip`) will silently point at a dead address.
```

docs/DEPLOY.md
```diff
@@ -0,0 +1,207 @@
...
+Install Caddy (Debian/Ubuntu):
+
+```bash
+sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl
+curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg
+curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list
+sudo apt update && sudo apt install -y caddy
+```
+
+For other install methods, see the [Caddy install docs](https://caddyserver.com/docs/install).
```
SeverinAlexB: I am always hesitant to have such details install instructions in a guide. As soon as Caddy changes something, this is outdated.

Isn't it enough to link to https://caddyserver.com/docs/install#debian-ubuntu-raspbian?

docs/INSTALL.md
```diff
@@ -226,48 +229,95 @@ docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserv
...
+
+```ini
+[Unit]
+Description=Pubky Homeserver
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+# Path to the pubky-homeserver binary
+ExecStart=/usr/local/bin/pubky-homeserver
```
SeverinAlexB: Should we add an explicity "--data-dir /absolute/path"?

docs/INSTALL.md
```diff
@@ -226,48 +229,95 @@ docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserv
...
+# The homeserver's data directory (contains config.toml and the keypair)
+WorkingDirectory=/home/youruser/.pubky
```
SeverinAlexB: Is this needed? Will this do anything?

docs/DEPLOY.md
```diff
@@ -0,0 +1,207 @@
...
+# Should return 200
+curl -I https://your_domain.com
+```
+
+### Find Your Public Key
+
+The remaining checks need your homeserver's public key. Retrieve it via the admin API:
+
+```bash
+curl -s "http://127.0.0.1:6288/info" -H "X-Admin-Password: admin" | grep public_key  # use your configured password. Default is "admin".
```
SeverinAlexB: AI review: 

Low: docs/DEPLOY.md:156 (docs/DEPLOY.md#L156) uses grep public_key, but /info returns compact JSON, so this prints the entire response rather than the key. Use jq -r '.public_key' and clarify that the command runs on the homeserver because port 6288 binds to localhost.

docs/DEPLOY.md
```diff
@@ -0,0 +1,207 @@
...
+- [Open Ports](#open-ports)
+- [Configure the Homeserver](#configure-the-homeserver)
+- [Set Up DNS](#set-up-dns)
+- [Set Up Caddy as a Reverse Proxy](#set-up-caddy-as-a-reverse-proxy)
+- [Verify the Deployment](#verify-the-deployment)
+  - [Check ICANN HTTPS](#check-icann-https)
+  - [Find Your Public Key](#find-your-public-key)
+  - [Check PKARR Record](#check-pkarr-record)
+  - [Check Pubky TLS](#check-pubky-tls)
+  - [Test a Signup](#test-a-signup)
```
SeverinAlexB: AI review:

Low: docs/DEPLOY.md:20 (docs/DEPLOY.md#L20) links to #test-a-signup, but that section does not exist. Add the promised end-to-end signup test or remove the entry.

docs/INSTALL.md
```diff
@@ -226,48 +229,95 @@ docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserv
...
+sudo nano /etc/systemd/system/pubky-homeserver.service
+```
+
+Paste the following:
+
+```ini
+[Unit]
+Description=Pubky Homeserver
+After=network-online.target
+Wants=network-online.target
```
SeverinAlexB, line 254: AI review:

```
Medium: docs/INSTALL.md:246 (docs/INSTALL.md#L246) creates a systemd service whose normal shutdown is not graceful. Systemd sends SIGTERM, while the binary only waits for SIGINT through tokio::signal::ctrl_c() (main.rs:85-89). Add KillSignal=SIGINT or implement SIGTERM handling before recommending this unit for production.
```

SeverinAlexB (changes requested):
Early review. I will try to execute the guide step by step now

docs/INSTALL.md
```diff
@@ -226,48 +229,95 @@ docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserv
...
+
+> **Tip:** If PostgreSQL runs on the same machine, add it to the `After=` line so the homeserver waits for it:
+> `After=network-online.target postgresql.service` (or `docker.service` if using Docker for Postgres).
+
+Enable and start the service:
+
+```bash
+sudo systemctl daemon-reload
+sudo systemctl enable --now pubky-homeserver
+```
```
SeverinAlexB, line 277: Can we do a regular enable & start instead of `--now`? Not sure a lot of people know the --now param and what it does and I want to avoid confusion.


```suggestion
```bash
sudo systemctl daemon-reload
sudo systemctl start pubky-homeserver
sudo systemctl enable pubky-homeserver
```
```

File: docs/INSTALL.md [hunk 10]
```diff
@@ -226,48 +229,95 @@ docker run -it --network=host -v ~/.pubky:/root/.pubky pubky-homeserver homeserv
...
+Check that it is running:
+
+```bash
+systemctl status pubky-homeserver
+```
+
+View logs:
+
+```bash
+journalctl -u pubky-homeserver -f
```
SeverinAlexB, line 288:
```suggestion
journalctl -u pubky-homeserver -ef
```

`-e` will jump to the end of the logs. Helpful so one does not need to scroll down.

docs/DEPLOY.md
86667: Since Caddy cant understand Pubky TLS it would be a raw TCP proxy, so it couldnt apply any of its http-level features. All we'd really get is connection-level metrics and IP-based rate limiting, which we control at the HS level anyway. 

We would require `layer4` module to support raw tcp proxying which requires a custom build of Caddy, eg:

```
# Install xcaddy
go install github.com/caddyserver/xcaddy/cmd/xcaddy@latest

# Build Caddy with the layer4 module
xcaddy build --with github.com/mholt/caddy-l4
```



I dont think its worth it.

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Check Pubky TLS
+
+From a separate machine with the [pkarr](https://github.com/pubky/pkarr) repository cloned:
+
+```bash
+cargo run --features=reqwest-builder --example http-get https://<homeserver-public-key>
+```
+
+A successful response prints `Pubky Homeserver`.
```
SeverinAlexB, line 255: Not part of this guide but we need to make these checks more convenient. For example:
The `homeservercli` should have a check command that does this automatically.

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+From your local machine (not the server), verify that internal ports are **not** reachable. Replace `YOUR_HOST` with your domain or IP address:
+
+```bash
+# These should all time out or refuse the connection
+curl --connect-timeout 5 http://YOUR_HOST:6286
+curl --connect-timeout 5 http://YOUR_HOST:6288/info
+curl --connect-timeout 5 http://YOUR_HOST:6289
+```
+
+If any of these respond, your firewall is misconfigured — go back to [Open Ports](#open-ports).
```
SeverinAlexB: Not needed. This is too much for a guide and should be general knowledge.

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+## Production Notes
+
+- Back up the homeserver's state regularly:
+  - The keypair at `~/.pubky/secret` — this is the homeserver's identity. If lost, the server cannot be recovered.
+  - User data — by default stored in `~/.pubky/data/files` (depends on `storage.type`).
+  - The PostgreSQL database.
+- Change the default admin password in `[admin].admin_password`.
+- **IP Address Only users:** Shortlived certificates expire in ~6 days. Caddy renews them automatically, but if renewal fails for more than a few days your HTTPS endpoint will go down. Monitor your Caddy logs and ensure port 80 stays reachable.
```
SeverinAlexB: No need to monitor the Caddy logs IMO

```suggestion
- **IP Address Only users:** Shortlived certificates expire in ~6 days. Caddy renews them automatically, but if renewal fails for more than a few days your HTTPS endpoint will go down. Ensure port 80 stays reachable.
```

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+
+| Port | Purpose |
+| --- | --- |
+| 80 | HTTP — [Caddy](#set-up-https) needs this for automatic TLS certificate provisioning. |
+| 443 | HTTPS — serves your homeserver via [Caddy](#set-up-https). Also used for TLS-ALPN-01 certificate challenges. |
+| 6287 | Pubky TLS — direct Pubky protocol connections (no certificate needed). |
+
+How you open these depends on your setup: a cloud provider's security group, a router's port-forwarding rules, or a host firewall.
+
+Do **not** expose these ports:
```
SeverinAlexB:
```suggestion
Do **not** expose these ports directly to the internet:
```

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+# With a Domain: your domain name for HTTPS browser access.
+# IP Address Only: set this to your IP address (same as public_ip).
+icann_domain = "YOUR_DOMAIN_OR_IP"
+```
+
+Replace `YOUR_IP` with the public IP of the machine running the homeserver. You can find it with `curl -4 ifconfig.me`.
+
+Replace `YOUR_DOMAIN_OR_IP` with your domain name or public IP, depending on [which setup you chose](#choose-your-setup).
+
+For the full list of settings (including `public_pubky_tls_port` and `public_icann_http_port` for non-standard port setups), see [`pubky-homeserver/config.sample.toml`](../pubky-homeserver/config.sample.toml).
```
SeverinAlexB: This is not important in this context IMO. Remove?

docs/DEPLOY.md
```diff
@@ -0,0 +1,293 @@
...
+
+> **Important:** Ensure your server has a **static (reserved) public IP**. If your IP changes then the PKARR record (which embeds `public_ip`), your Caddy configuration, and any DNS records will silently point at a dead address.
+
+Restart the homeserver after editing `config.toml` for the changes to take effect. If you have set it up as a background service, see [systemd Service](./INSTALL.md#systemd-service) in the install guide.
```
SeverinAlexB: Changing the config.toml made me realize that it would be very convenient to have a `homeservercli config-check` command to see if the config file is valid.

SeverinAlexB (changes requested):

docs/INSTALL.md
86667: `-f` starts from the end and follows? Did it act differently for you? Could be a macos difference

docs/DEPLOY.md
86667: I think it is - the thinking is that someone not using the standard default port setup should at least be pointed to where they can learn about custom configurations.

docs/INSTALL.md
SeverinAlexB: My bad, it looks like -f does already jump to the end.

docs/DEPLOY.md
86667: Yes! also some convenience functions for setting common vars, eg `homeservercli config set-database-connection-string`

docs/DEPLOY.md
SeverinAlexB: Should we use demphasize it then with a `> ℹ️`?

docs/DEPLOY.md
SeverinAlexB: We could. This would need to be made smart though so the user is aware what he overrides.
A flexible config set could look like this:

```
# homeservercli config set %title.key% %value%

$ homeservercli config set general.database_url "postgres://postgres:postgres@localhost:5432/pubky_homeserver"

Set "general.database_url "
from "postgres://postgres:postgres@localhost:5432/old_db"
to "postgres://postgres:postgres@localhost:5432/pubky_homeserver"
Confirm? [Yn]
```

docs/DEPLOY.md
```diff
@@ -0,0 +1,286 @@
...
+- **With a Domain** — requires a domain name and DNS setup. Standard 90-day Let's Encrypt certificates.
+- **IP Address Only** — no domain needed. Uses short-lived (~6-day) Let's Encrypt certificates, so the server must stay healthy for renewals.
+
+This guide covers both options.
+
+## Open Ports
+
+The homeserver speaks two protocols. Both endpoints serve the same data, the difference is how clients find and connect to your server:
+
+**Pubky TLS** is the native protocol: clients resolve your homeserver's public key on the DHT and connect directly on port 6287, bypassing centralized DNS.
```
SeverinAlexB: bypassing centralized DNS is not the differentiating quality here. It's bypassing any certificate authority.
```suggestion
**Pubky TLS** is the native protocol: clients resolve your homeserver's public key on the DHT and connect directly on port 6287, bypassing any certificate authority.
```

docs/DEPLOY.md
```diff
@@ -0,0 +1,286 @@
...
+
+**HTTPS** is the standard web protocol: browsers and the browser-based SDK use this on port 443, which requires a TLS certificate for your domain or IP address.
```
SeverinAlexB: Both endpoints do HTTPS. The difference is cert authority vs no-cert authority. So calling it "HTTPS" feels wrong here.

Here is a chat with alternative names: https://chatgpt.com/share/6a5f2fd4-8820-83eb-9fd9-b00fbb60ddf2

SeverinAlexB (changes requested):
We are almost there. One AI review that I am not sure about:

Medium: The documented Docker deployment does not fully survive a reboot. docs/INSTALL.md:226-229 gives the homeserver a restart policy, but the PostgreSQL container created at docs/INSTALL.md:151-158 has none. Likewise, ordering systemd after docker.service at docs/INSTALL.md:273-274 does not start that container. After reboot, the homeserver repeatedly fails because PostgreSQL remains stopped. Add a restart policy to the PostgreSQL command or explicitly start it through systemd.

docs/DEPLOY.md
86667: sure

docs/DEPLOY.md
86667: hmm i agree that theyre both https so this could be misleading, that name is useful though since everyone knows what it means. Instead, Ive attempted to communicate the idea of "this is the one youre used to" whilst also being technically correct but with minimal jargon. What do you think?

SeverinAlexB (approved):
LGTM

PR #505, opened 2026-07-17
docs(hs/sdk): add private storage documentation

### LINKED ISSUE #442: Spec + docs for `/priv/`
### Scope

Document the `/priv/` storage model and private-event behavior for app developers and maintainers.

- Document the two-root model: `/pub/` is public/world-readable, `/priv/` is private/authenticated, and visibility is a function of the path.
- Document `pubky://` semantics for private resources.
- Document that publishing means writing to `/pub/`, while private writes live under `/priv/`; nothing needs to move for existing apps.
- Document capability requirements: apps need a covering capability such as `/priv/app/:rw` for their private subtree, and reads require a covering read cap.
- Document event behavior:
  - public `/events/` remains public-only
  - private events are available through authenticated `/events-stream`
  - admin event streams may include private events
- Document SDK/event-stream auth behavior added around private subscriptions:
  - private path filters require an attached session
  - public-only streams do not require a session
  - private streams are limited to a single target tenant/user
  - bearer auth takes precedence over cookie auth
  - legacy homeserver-addressed cookie fallback only applies to private single-tenant event streams
- Document path filter normalization for event streams, including `.` / `..`, missing leading slash, duplicate slash, and invalid traversal cases.
- Add/adjust code comments or function docs where the implementation is intentionally subtle, especially around event-stream auth, tenant cookie fallback, and session request helper naming.
- Migration note for app devs: private storage is a new namespace; existing public data does not move.

## Dependencies

Part of epic #426.

Blocked by _(soft — finalize after these land, since docs describe their final contracts)_:
- #432, #433, #434 — storage behaviors
- #436 — public-feed filtering
- #429 — `/events-stream` authz
- #427 — admin events

MCarlomagno:
Closes #442.

### Summary

Documents `/priv/` storage


pubky-homeserver/openapi.yml
```diff
@@ -24,12 +24,12 @@ info:
     - **Grant-based (current):** `Authorization: Bearer <token>` obtained via `POST /auth/grant/session`.
     - **Cookie-based (deprecated):** Session cookie set by `POST /signup` or `POST /session`.
 
-    Grant middleware runs first; if it produces a session, cookie middleware is skipped.
+    A resolved bearer session takes precedence over cookie authentication.
```
MCarlomagno: this was introduced to prevent unexpected behaviour in the client, we don't allow an invalid bearer session with a valid cookie session

docs/PRIVATE_STORAGE.md
andrei-21, line 1: I think we should mention here only client API (not admin) otherwise it is a bit confusing.

docs/PRIVATE_STORAGE.md
```diff
@@ -0,0 +1,84 @@
+# Private Storage
+
+## Storage namespaces
+
```
andrei-21, line 8: Please mention somewhere here that all data might be read and written by a homeserver admin.

docs/PRIVATE_STORAGE.md
```diff
@@ -0,0 +1,88 @@
...
+form a union:
+
+```http
+GET /events-stream?user=<user>&path=/pub/my-app/&path=/priv/my-app/
+Authorization: Bearer <session-token>
+```
+
+If a Bearer header is present, cookie authentication is not considered. The
+legacy tenant-cookie fallback applies only to private, single-user streams sent
+directly to a homeserver. The admin event stream can include private events.
```
andrei-21, line 73: Hm, AI claims that it is no true:
> When `/events-stream` is user-addressed through `Host` or `pubky-host`, an invalid Bearer token still falls through to the normal cookie middleware, which can resolve a valid tenant cookie before `feed_stream` runs. Because `feed_stream` preserves that resolved session, the categorical claim that any Bearer header disables cookie authentication is false; restrict this statement to homeserver-addressed fallback or align the middleware behavior with the documented rule.

If it is correct, maybe we should fix the code.

MCarlomagno: yes, taht catch was somehow missed by mine, working on the fix

MCarlomagno: maybe this is not the right PR for the fix, but since it's minimal I will go ahead and do it here

andrei-21 (approved):

PR #510, opened 2026-07-22
fix(homeserver): Consolidate stateful OpenDAL layers into one

### LINKED ISSUE #449: Refactor: consolidate stateful OpenDAL layers into a single `WriteFinalizationLayer`
## Background

Our storage stack is composed of single-purpose OpenDAL layers, each wrapping the next. 

After #424 the runtime stack is:

```
WritePathLayer → PathCollisionLayer → EventsLayer → EntryLayer → UserQuotaLayer → backend
```

This single-responsibility layering (like Axum) is clean for stateless concerns, but it breaks down for operations that must be atomic with the blob write. OpenDAL layers have no mechanism for sharing request-scoped state between layers, unlike Axums request extensions, the builtin OpenDAL layers (retry, timeout) are all stateless. As a result, every layer that touches Postgres opens its **own** transaction in `close()`:

- `EntryLayer::close` — own `pool().begin()` to write the entry metadata
- `UserQuotaLayer::close` — own `pool().begin()` + `SELECT … FOR UPDATE` on the user row
- `EventsLayer::close` — own `pool().begin()` to create the event

We end up with three independent transactions and connections per upload, no shared lock, and no single place that defines what is atomic. As @86667  put it: using Axum like single-purpose layering here is fighting the design of the crate.

## Problems this causes

1. **No atomic "lock → check → write blob → unlock" unit.** This is what we actually want for collision + quota enforcement, and it can't be expressed cleanly across separate layers. → see the concurrency issue (https://github.com/pubky/pubky-core/pull/424#discussion_r3396739063).
2. **`EventsLayer` can report failure on a persisted write.** `EventsLayer::close` runs `self.inner.close().await?` first (blob + entry are already durably committed), then creates the event in a separate transaction. If event creation fails, the user gets an error, but the blob and entry persist.
3. **Coordinating the layers by nesting causes deadlocks.** An earlier attempt in #424 reordered the stack so the lock-holding layer wrapped `EntryLayer`, holding one pool connection + the user-row lock across a nested `pool().begin()` for a second connection. With the default sqlx pool (10 connections, 30s acquire timeout), enough concurrent closes exhaust the pool while each waits on a connection the lock-holder itself needs → the upload batch times out. This is a direct symptom of trying to share state OpenDAL won't let us share.

## Proposal

Combine all pre-checks and after-effects that depend on the blob write succeeding into a single layer — e.g. `WriteFinalizationLayer` — with a single Postgres transaction and a single lock, so one `close()` does:

```
begin tx
  lock user row (FOR NO KEY UPDATE)   // serialize same-user finalisations
  collision check
  quota check
  inner.close()                       // flush blob to backend
  write entry metadata                // same tx
  create event                        // same tx
  update user.used_bytes              // same tx
commit
```

Keep genuinely stateless pre-checks (e.g. `WritePathLayer`, path-format validation) as their own separate OpenDAL layers, that split is correct precisely because there's no shared state.

## Why this is simpler

The repository methods already accept a `&mut UnifiedExecutor` and **don't own their transactions** (`EntryService::write_entry`, `EventRepository::create`, `UserRepository::update`). Routing them all through one `UnifiedExecutor::from_tx(&mut tx)` is mostly **deleting** the duplicate `begin()/commit()` blocks in `EntryLayer`, `EventsLayer`, and `UserQuotaLayer`, the consolidated layer should be *smaller* than the sum of the three. One tx, one connection, one lock per close.

## Acceptance criteria

- [ ] A single `WriteFinalizationLayer` owns the collision check, quota check, entry write, event creation, and quota update within one transaction and one user-row lock.
- [ ] `WritePathLayer` (and any other purely stateless pre-check) remains a separate layer.
- [ ] No `close()` path holds more than one Postgres connection / acquires more than one at a time.
- [ ] N > pool-size concurrent uploads (same-user and cross-user) complete without pool-acquire timeouts.
- [ ] A forced event-creation failure rolls back the entry write (no user-visible failure on a persisted entry).
- [ ] Existing storage/e2e tests pass, the new layer reads clearly.

MCarlomagno:
Closes #449.

### Summary

Consolidates the four stateful OpenDAL layers (`PathCollisionLayer`, `UserQuotaLayer`, `EntryLayer`, `EventsLayer`) into a single `WriteFinalizationLayer`. Each of those layers opened its own Postgres transaction in `close()`, so there was no atomic "lock → check → write blob → record" unit: an event-insert failure surfaced as an error on a write that had already persisted, and coordinating the layers by nesting connections deadlocked the pool under concurrent uploads (#424).

One `close()` now runs a single transaction holding a single `FOR NO KEY UPDATE` lock on the user row:

```
begin tx
  lock user row (FOR NO KEY UPDATE)   // serialize same-user finalizations
  collision check
  quota check
  inner.close()                       // flush blob to backend
  write entry metadata
  create event
  update user.used_bytes
commit
```

Same-user operations serialize on the row lock, cross-user operations proceed in parallel, and no `close()` path ever holds more than one pool connection. The stack is now `WritePathLayer` (stateless pre-check) → `WriteFinalizationLayer` → backend.

### Changes

- Added `pubky-homeserver/src/persistence/files/write_finalization_layer.rs`, with tests covering write/overwrite/delete atomicity, event-failure rollback, concurrent collision races, duplicate deletes, mid-batch delete failures, and a pool-exhaustion regression test on a bounded 2-connection pool.
- Removed `path_collision_layer.rs`, `user_quota_layer.rs`, `entry/entry_layer.rs`, and `events/events_layer.rs`.
- `EntryService` is now stateless: `write_entry` takes the caller's `user_id` and pre-fetched `existing_entry`, and `delete_entry` returns the deleted entry for transactional quota accounting.
- Added `UserRepository::get_for_no_key_update` (+ `UserService` wrapper) to serialize storage accounting without blocking foreign-key checks.
- `build_storage_operators` no longer takes a `db` handle and builds two finalization instances: app-facing (collisions enforced) and admin (legacy-repair bypass).
- Added `SqlDb::test_with_pool_options` for tests that need a bounded connection pool.

### Acceptance criteria

- [x] A single `WriteFinalizationLayer` owns the collision check, quota check, entry write, event creation, and quota update within one transaction and one user-row lock.
- [x] `WritePathLayer` (and any other purely stateless pre-check) remains a separate layer.
- [x] No `close()` path holds more than one Postgres connection / acquires more than one at a time.
- [x] N > pool-size concurrent uploads (same-user and cross-user) complete without pool-acquire timeouts.
- [x] A forced event-creation failure rolls back the entry write (no user-visible failure on a persisted entry).
- [x] Existing storage/e2e tests pass, the new layer reads clearly.


pubky-homeserver/src/persistence/files/write_finalization_layer.rs
```diff
@@ -0,0 +1,971 @@
...
+impl<R: oio::Delete> oio::Delete for WriteFinalizationDeleter<R> {
+    fn delete(&mut self, path: &str, args: OpDelete) -> Result<()> {
+        let entry_path = EntryPath::parse_opendal(path)?;
+        self.inner.delete(entry_path.as_str(), args)?;
+        self.delete_queue.push(entry_path);
+        Ok(())
+    }
+
+    async fn flush(&mut self) -> Result<usize> {
+        let deleted_count = self.inner.flush().await?;
```
86667: Does this mean that we delete the data blobs before the checks and metadata operations? I prefer the way write works, where the blob write is only done after all checks and metadata operations have succeeded

pubky-homeserver/src/persistence/files/write_finalization_layer.rs
```diff
@@ -0,0 +1,971 @@
...
+    async fn write(&mut self, bs: opendal::Buffer) -> Result<()> {
+        self.metadata_builder.update(&bs.to_vec());
+        self.inner.write(bs).await
+    }
+
+    async fn abort(&mut self) -> Result<()> {
+        self.inner.abort().await
+    }
+
+    async fn close(&mut self) -> Result<opendal::Metadata> {
```
86667: Lots going on in this fn. Maybe more logic could be extracted into the `Finalizer`

pubky-homeserver/src/persistence/files/write_finalization_layer.rs
```diff
@@ -0,0 +1,971 @@
...
+        self.metadata_builder
+            .guess_mime_type_from_path(self.entry_path.path().as_str());
+        let file_metadata = self.metadata_builder.clone().finalize();
+        let entry_service = EntryService::new();
```
86667: nit: Create on the fly as needed? Or init once on Finalizer may be cleaner

86667 (commented):
Good to see these correctness issues being resolved 💪

pubky-homeserver/src/persistence/files/write_finalization_layer.rs
86667: Could we call `EntryRepository` directly instead, actually? What is the purpose of the EventService abstraction

pubky-homeserver/src/persistence/files/write_finalization_layer.rs
MCarlomagno: I decided to actually do a bigger refactor as it was getting messy, I split write_finalization_layer into a module with multiple files

pubky-homeserver/src/persistence/files/write_finalization_layer/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+mod quota;
+mod write;
+
+pub use delete::WriteFinalizationDeleter;
+pub(crate) use quota::{resolve_storage_max_bytes, would_exceed_limit};
+pub use write::WriteFinalizationWriter;
+
+/// Finalizes storage mutations whose database effects must share one transaction.
+///
+/// App-facing operators enable collision checks; admin operators do not.
```
86667: Could you also include a higher-level overview - eg This is an OpenDal layer of atomic actions performed during a write...

86667: And we have a general pattern of including only exports and docs in `mode.rs`, no code.

pubky-homeserver/src/persistence/files/write_finalization_layer/delete.rs
```diff
@@ -0,0 +1,443 @@
...
+
+        operator
+            .write(entry_path.as_str(), content.clone())
+            .await
+            .unwrap();
+        let usage_before_delete = user_usage(&db, &pubkey).await;
+
+        sqlx::query(
+            r#"
+            CREATE FUNCTION fail_delete_event_insert() RETURNS trigger AS $$
```
86667: Pull out into a helper just as a way to make it clear what its doing?

pubky-homeserver/src/persistence/files/write_finalization_layer/write.rs
```diff
@@ -188,6 +187,32 @@ impl Finalizer {
         PreparedWrite::new(user, existing_entry, file_metadata, self.default_storage_mb)
     }
 
+    async fn persist_entry(
```
86667: this feels like an entry repository concern

MCarlomagno: yup, it internally uses the repository, anyways this is called just once so I can inline this method in `apply_write_effects`

86667 (approved):

PR #512, opened 2026-07-23
ci: use rust-cache in `build-artifacts` job

ok300:
This PR partially addresses #486 by adding `Swatinem/rust-cache@v2` to the `build-artifacts` workflow, which runs on pushes to `main`.

### Cache scope

`rust-cache` caches Cargo registry data and installed tools, including the `cross` executable and Cargo installation metadata. On a warm cache hit, Cargo still resolves the pinned Git source but avoids recompiling and reinstalling `cross`.

The pinned `cross` revision is defined once as `CROSS_REV` and included in the cache key. Because GitHub cache entries are immutable, changing the revision creates a fresh cache instead of repeatedly restoring an outdated installation.

`ImageOS` is also included in the cache environment hash so the host-side `cross` executable is not shared across different Ubuntu runner generations and potentially incompatible glibc versions.

Workspace target-directory caching remains disabled. The Cross Docker images use mutable tags, so restored build scripts or artifacts may have been compiled against a different container image and glibc version. All five cross targets will therefore rebuild on each run, while Cargo downloads and the installed `cross` tool can still be reused.

The existing per-target build-directory layout is unchanged.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -10,6 +10,9 @@ on:
   push:
     branches: [main]
 
+env:
+  CROSS_REV: 51f46f296253d8122c927c5bb933e3c4f27cc317
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -26,14 +29,22 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+
       - name: Cache Rust
         uses: Swatinem/rust-cache@v2
         with:
           # Cache Cargo downloads and installed tools, but exclude target artifacts.
           # Cross images tagged :latest may change and make restored artifacts incompatible.
           cache-targets: false
+
+          # Cache the installed cross executable by its pinned source revision.
+          # Changing CROSS_REV therefore creates a fresh cache entry.
+          # Include ImageOS to avoid runner glibc incompatibilities.
+          key: cross-${{ env.CROSS_REV }}
+          env-vars: ImageOS
+
       - name: Install cross # Tool to cross-compile the binaries
-        run: cargo install cross --locked --git https://github.com/cross-rs/cross --rev 51f46f296253d8122c927c5bb933e3c4f27cc317
+        run: cargo install cross --locked --git https://github.com/cross-rs/cross --rev "$CROSS_REV"
 
 
       # Build artifacts
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -28,6 +28,10 @@ jobs:
           toolchain: stable
       - name: Cache Rust
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/build-artifacts.yml
```diff
@@ -26,8 +29,22 @@ jobs:
...
+
+          # Cache the installed cross executable by its pinned source revision.
+          # Changing CROSS_REV therefore creates a fresh cache entry.
+          # Include ImageOS to avoid runner glibc incompatibilities.
+          key: cross-${{ env.CROSS_REV }}
+          env-vars: ImageOS
+
       - name: Install cross # Tool to cross-compile the binaries
-        run: cargo install cross --locked --git https://github.com/cross-rs/cross --rev 51f46f296253d8122c927c5bb933e3c4f27cc317
+        run: cargo install cross --locked --git https://github.com/cross-rs/cross --rev "$CROSS_REV"
```
andrei-21, line 47: From AI (sounds legit):
> Once the new cache has been populated, subsequent main-branch runs can fail during the unconditional cross installation before producing artifacts.
>
> Skip reinstalling cross after a cache hit.
>  On any run that restores this cache, `rust-cache` restores `~/.cargo/bin/cross` but not Cargo's `.crates.toml`/`.crates2.json` installation metadata. The subsequent unconditional `cargo install cross` therefore treats the restored executable as untracked and exits with `binary 'cross' already exists in destination`, preventing artifact builds after the cache is populated. Make the installation conditional on `cross` being absent or explicitly cache the installation metadata.

ok300: Apparently not true for the more recent `rust-cache` versions (we're using the latest `v2.9.1`):

> - Since `v2.7.7` (released Dec 2024), `rust-cache` explicitly caches `~/.cargo/bin`, `~/.cargo/.crates.toml`, and `~/.cargo/.crates2.json`. See [src/config.ts:272-279](https://github.com/Swatinem/rust-cache/blob/e18b497796c12c097a38f9edb9d0641fb99eee32/src/config.ts#L272-L279) and the commit adding Cargo install metadata (https://github.com/Swatinem/rust-cache/commit/008623fb834cadde1d7ccee1a26dc84acb660ec3).
> - `cache-targets: false` only excludes workspace target directories; it does not disable binary or metadata caching.
> - Cargo treats an installation with the same source, revision, binaries, features, profile, and target as up to date. It exits successfully rather than reporting a binary collision. This is documented under `cargo install` (https://doc.rust-lang.org/cargo/commands/cargo-install.html#description).
> - The pinned `CROSS_REV` is part of both the exact and fallback cache key, so a cache from another Cross revision cannot be restored.
>
> The comment’s failure mode would be valid for a custom or older cache containing cross without Cargo’s metadata, but that is not what this PR produces.

andrei-21 (approved):

PR #514, opened 2026-07-23
refactor(common)!: Strict capabilities and shared storage paths

### LINKED ISSUE #478: Allow Capabilities to parse from &str and yield proper error messages
`Capabilities` is a bit of a mess currently. Converting a capabilities string like `/pub/app/:w,missing-leading-slash:r,/priv/file.txt:x` to a Capabilities struct has multiple issues:

1. Instead of a `parse()` method (FromStr trait), it uses the TryFrom<&str> which is [technically incorrect](https://chatgpt.com/share/6a47679a-ab58-83eb-b94d-b34a094d8ade).
2. The current TryFrom implementation swallows any error. Parsing `/pub/app/:w,missing-leading-slash:r,/priv/file.txt:x` just returns a `Capabilities([Capability { scope: "/pub/app/", actions: [Write] }])` which is incorrect.
3. Even if you would return the Capability parse error, the error would only say something like "InvalidScope" or "InvalidAction". It would not point out, which one of the three `/pub/app/:w,missing-leading-slash:r,/priv/file.txt:x` capabilities is wrong.
4. The JS SDK gets around this by writing a custom `validate_capabilities(input: &str)` method which violates the DRY and Separation of Concerns principles.

This needs attention.

SeverinAlexB:
Makes capability parsing strict and consolidates capability scopes and homeserver storage paths around a shared canonical `StoragePath` type. Malformed capability entries are now rejected instead of silently discarded. Fixes #478.

## Capability changes

- `Capabilities` now implements `FromStr`:
  ```rust
  let caps = "/pub/app/:rw,/priv/file:r".parse::<Capabilities>()?;
  ```
- Added contextual parsing errors:
  - `CapabilityParseError`
  - `CapabilitiesParseError`, containing the invalid entry, its one-based position, and the source error.
- Capability constructors and builder helpers are now fallible.
- Empty actions such as `/pub/app/:` are rejected.
- Leading, trailing, and repeated commas are rejected. An entirely empty capability list remains valid.
- Capability scopes must be canonical decoded storage paths. Relative paths, repeated separators, . and .. segments, control characters, oversized paths, and reserved : or , delimiters are rejected.
- `from_caps_url`, which silently ignored malformed capabilities, has been replaced by fallible `try_from_caps_url`.
- Serde, deep links, auth flows, and persisted capability strings now use the strict parser.
- Capability normalization still merges actions and removes scopes covered by broader capabilities.

## Shared storage paths

Introduces the public `StoragePath` and `StoragePathError `types in pubky-common, also re-exported by the Rust SDK.

`StoragePath` provides:
- Strict parsing for already-canonical decoded paths.
- Explicit normalization for HTTP/WebDAV input.
- URL encoding.
- File, directory, and root classification.
- Segment and total-path length validation.
- Strict serde support.

Capability scopes now use `StoragePath`. The homeserver's previous `WebDavPath` implementation has been removed, and routes, authorization, events, persistence, quotas, and storage operations now use the shared type.
HTTP/WebDAV paths are normalized at ingress, while capability scopes and persisted paths must already be canonical.

## Persistence migration

Adds a database migration for existing cookie sessions:

- Valid capability strings are normalized.
- Sessions containing any invalid capability entry are deleted.

This prevents legacy values accepted by the previous permissive parser from failing strict deserialization at runtime.

## Javascript

`validateCapabilities` and the JS auth-flow constructors now delegate to the Rust parser and normalizer.
Validation stops at the first malformed entry. The error message includes its position and `data.invalidEntries` contains that exact entry.
```
validateCapabilities("/pub/pubky.app/:rw,missing-leading-slash:r");
```
invalid capability at position 2 (`missing-leading-slash:r`): invalid capability scope: path must be absolute

## Breaking changes

This PR includes breaking changes mostly around the Capability/Capabilities struct.


pubky-common/src/capabilities.rs
```diff
@@ -250,57 +250,62 @@ impl Display for Capability {
...
-        if !value.starts_with('/') {
-            return Err(Error::InvalidScope);
+        if !scope.starts_with('/') {
+            return Err(CapabilityParseError::InvalidScope);
         }
 
-        let actions_str = value.rsplit(':').next().unwrap_or("");
+        if actions_str.is_empty() {
+            return Err(CapabilityParseError::MissingActions);
+        }
```
MCarlomagno, line 239: minor but worth noting: 
The public builder allows `Capability::builder("/foo").finish()`, which serializes as `/foo:`, but now the parser rejects that representation as `MissingActions`

HS can write `capabilities.to_string()` without validation, then use the parser when reading. This means that a session or grant can be inserted successfully but fail in every subsequent read

SeverinAlexB: I removed the builder. It's not necessary anymore.

pubky-homeserver/src/persistence/sql/migrations/m20260723_sanitize_capabilities.rs
```diff
@@ -106,6 +106,8 @@ mod tests {
         ("/pub/a/../b/:w", None),
         ("/foo,/bar:w", None),
         ("/foo:r,,/bar:w", None),
+        ("/priv/report :w", None),
```
MCarlomagno, line 109: maybe we should reject trailing whitespace in every path segment,

Consider `/priv/app/secret` and `/priv/app /secret` (whitespace before /secret). Our authorization treats them as different, but some backend filesystems might trim trailing spaces from folder names, making both point to the same file, this could let one app access or overwrite another apps private data - verified with AI

pubky-homeserver/src/shared/webdav/entry_path.rs
```diff
@@ -29,7 +29,7 @@ pub struct EntryPath {
 }
 
 impl EntryPath {
-    pub fn new(pubkey: PublicKey, path: WebDavPath) -> Self {
+    pub fn new(pubkey: PublicKey, path: StoragePath) -> Self {
```
MCarlomagno, line 32: nit: maybe for the sake of consistency we can rename `EntryPath` to something like `TenantStoragePath` (which includes a logic related to a storage path and a tenant) - but we can leave that for a different PR if is too much impact

pubky-homeserver/src/persistence/sql/migrations/m20260723_sanitize_capabilities.rs
SeverinAlexB: I would like to not blow up this PR with these details tbh. I created an issue yesterday exactly about this https://github.com/pubky/pubky-core/issues/519.

My best guess is that we probably need to encode our paths when writing them to a backend. Maybe URL encode even. We should not restrict our API too much because otherwise existing webdav applications, that might be currently compatible with our API, might not be anymore. I am unsure even with the trailing slash. Maybe this should just be encoded in the backend.

pubky-homeserver/src/shared/webdav/entry_path.rs
SeverinAlexB: I can do this in another. This is too much change otherwise.

MCarlomagno (approved):
LGTM

PR #515, opened 2026-07-24
feat(homeserver): skip PKARR republishing for migrated users

### LINKED ISSUE #444: [Homeserver]: Republisher does not check that PKARR records point to itself
A user may change Homeserver, in which case the old one shouldnt continue to republish its pkarr record.

What exactly the Homeserver _should_ do in this case should be thought through.

andrei-21:
Republish user packets only when their _pubky target matches the
current homeserver. Keep migrated users' stored data intact.

Closes #444

### DIFF
--- a/pubky-homeserver/src/republishers/pkarr_republisher/batch_republisher.rs
+++ b/pubky-homeserver/src/republishers/pkarr_republisher/batch_republisher.rs
@@ -2,8 +2,9 @@ use super::republish_summary::{RepublishResult, RepublishSummary};
 use super::republisher::{RepublishOutcome, Republisher, RepublisherSettings};
 use super::retrying_republisher::{RetrySettings, RetryingRepublisher};
 use futures_util::{stream::FuturesUnordered, TryStreamExt};
-use pkarr::{errors::BuildError, ClientBuilder, PublicKey};
-use std::{num::NonZeroUsize, sync::Mutex};
+use pkarr::{errors::BuildError, ClientBuilder, PublicKey, SignedPacket};
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Mutex};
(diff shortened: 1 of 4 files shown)


pubky-homeserver/src/republishers/pkarr_republisher/batch_republisher.rs
```diff
@@ -26,6 +27,18 @@ impl Default for BatchRepublisherSettings {
     }
 }
 
+impl BatchRepublisherSettings {
+    /// Set a condition that packets must satisfy before they are republished.
+    #[must_use]
+    pub(crate) fn with_republish_condition<F>(mut self, condition: F) -> Self
+    where
+        F: Fn(&SignedPacket) -> bool + Send + Sync + 'static,
```
MCarlomagno, line 35: nit: maybe a refactor renaming it to "Policy" instead of "Condition" and wrapping them as enums in order to prevent indirection like:

```rust
#[derive(Debug, Clone, Default)]
enum RepublishPolicy {
    #[default]
    Always,
    PointsToHomeserver(PublicKey),
}

impl RepublishPolicy {
    fn allows(&self, packet: &SignedPacket) -> bool {
        match self {
            Self::Always => true,
            Self::PointsToHomeserver(key) => {  // the one you actually implemented.
                packet_points_to_homeserver(packet, key)
            }
        }
    }
}
```

MCarlomagno (approved):

pubky-homeserver/src/republishers/pkarr_republisher/batch_republisher.rs
andrei-21: I think we are always going to filter them out, so it will always be `RepublishPolicy::PointsToHomeserver(PublicKey)`.

PR #521, opened 2026-07-28
docs(testnet): README to prioritise instructions for persisted static testnet setup

86667:
The readme for testnet felt a bit scattered and not friendly for new users. 

I decided that the most common use case for reading this file is likely a dev wanting to spin up a local static persisted testnet. Therefore i created a quick start section at the top showing how to do this, with notes on modifying configs and persistence. 

Notes on `EphemeralTestnet` are still present but come later.

### DIFF
--- a/pubky-testnet/README.md
+++ b/pubky-testnet/README.md
@@ -27,8 +27,6 @@ The data directory is auto-initialized on first run with a `config.toml` and ser
 
 The `TEST_PUBKY_CONNECTION_STRING` environment variable is read on every startup and overrides the `database_url` in the on-disk config.
 
-> **Warning**: Do not use `?pubky-test=true` in the connection string with `persist` — it causes the database to be deleted on shutdown.
-
 To seed a custom homeserver config on first run (errors if `config.toml` already exists):
 
 ```bash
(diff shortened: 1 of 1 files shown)


pubky-testnet/README.md
```diff
@@ -2,36 +2,65 @@
...
+```bash
+TEST_PUBKY_CONNECTION_STRING='postgres://postgres:postgres@localhost:5432/postgres' \
+  cargo run -p pubky-testnet -- persist ./my-testnet-data
+```
+
+The data directory is auto-initialized on first run with a `config.toml` and server keypair. On subsequent runs, the existing state is picked up and the homeserver keeps the same identity.
+
+The `TEST_PUBKY_CONNECTION_STRING` environment variable is read on every startup and overrides the `database_url` in the on-disk config.
+
+> **Warning**: Do not use `?pubky-test=true` in the connection string with `persist` — it causes the database to be deleted on shutdown.
```
MCarlomagno: nit: this seems inaccurate, AI suggests:

"persist rejects connection strings containing ?pubky-test=true at startup; remove it so state is kept in a real database."

86667: removed warning!

MCarlomagno (approved):

PR #523, opened 2026-07-30
feat(ci,sdk): Setup TypeDoc and deploy to Github Pages

86667:
Generates a [TypeDoc](https://typedoc.org/) reference from the Rust `///` doc comments in the SDK. A workflow publishes the generated site to `https://pubky.github.io/pubky-core/js-sdk-typedoc`.

The idea here is that all reference documentation should live alongside the SDK's Rust code. The Rust docs are already published automatically to docs.rs. This workflow publishes the JS SDK side to Github Pages for convenient reference.

I will review the contents of the doc comments in a later PR.

### DIFF
--- a/.github/workflows/js-sdk-docs.yml
+++ b/.github/workflows/js-sdk-docs.yml
@@ -1,0 +1,70 @@
+name: JS SDK Docs
+
+on:
+  push:
+    tags: ['v*']
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: js-sdk-docs-pages
+  cancel-in-progress: true
+
+# NOTE: deploy-pages replaces the *entire* GitHub Pages site.
+# If another workflow needs to publish to Pages, consolidate both
+# into a single deployment so content is not overwritten.
+
+jobs:
+  build:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: pubky-sdk/bindings/js/pkg/package-lock.json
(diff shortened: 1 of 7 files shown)


.github/workflows/js-sdk-docs.yml
```diff
@@ -0,0 +1,70 @@
+name: JS SDK Docs
+
+on:
+  push:
+    tags: ['v*']
```
MCarlomagno, line 5: `github-pages` environment currently allows deployments only from the `main` and `gh-pages ` branches, so the deploy job would be rejected by the environment protection rules. I think we need to add a `v*` tag policy to the env

MCarlomagno (commented):
we could add `npm run docs` to the CI to make sure they work on every PR, wdyt?

.github/workflows/js-sdk-docs.yml
86667: Where do you see this environment config? I cant find it

86667:
> we could add `npm run docs` to the CI to make sure they work on every PR, wdyt?

I dont think this is necessary. Extra work on every PR and PR update just to check something which is unlikely to break and easy to fix if it does break. Also, if docs dont deploy its not the end of the world.

.github/workflows/js-sdk-docs.yml
MCarlomagno: I think I don't have access either, it should be:

Settings → Environments → github-pages
- Deployment branches and tags
- Clicking Add deployment branch or tag rule
- Selecting Tag
- Entering v*

MCarlomagno: Anyways we can add that afterwards, approving the PR

MCarlomagno (dismissed):

.github/workflows/js-sdk-docs.yml
86667: I dont think it exists yet because the action hasnt been run before. Lets see what happens on release

86667:
@MCarlomagno Thanks. I did a test deploy and configured the necessary vars. See here https://pubky.github.io/pubky-core/js-sdk-typedoc.

Please approve again as i had to force-push my test changes outl

MCarlomagno (approved):

PR #533, opened 2026-07-31
feat(sdk): add support to callback params

### LINKED ISSUE #336: [feature] Signin/up Callback URLs
When PubkyApp opens Ring for a sign-in/up, there is no way for Ring to go back to the application that opened it after the authorization is completed. This is a common problem on mobile devices. See https://x-callback-url.com/ .

We need to research if it's smart to add this x-callback to the signin/up flow?

Some questions that come to my mind:
- Would this work on all mobile devices?
- Will the QR code be too big? Do we need to optimize the QR code to prevent this?
- Do we need other actions like `x-cancel`?

So this is not just a programming task but also a feasibility/UX task.

FYI @coreyphillips @86667

MCarlomagno:
Closes #336.

### Summary

Adds first-class support for `x-callback` metadata in Pubky deep links:

- `x-source`
- `x-success`
- `x-error`
- `x-cancel`

The SDK now parses, preserves, and serializes these parameters through the typed deep-link model. Cookie and grant auth builders can attach them directly, and JS/WASM consumers receive them through the typed `xCallback` API instead of manually parsing the authorization URL.

### Behavior

- Callback values are decoded exactly once, matching Pubky Ring behavior.
- Literal `+` characters and nested percent-encoded values round-trip correctly.
- The legacy `callback` parameter is accepted as an `x-success` fallback; serialization emits the canonical `x-success` form.
- Callback metadata survives cookie and grant flow persistence/resume, including delegated grant flows.
- Signin, signup, grant, direct-signup, and seed-export deep-link types expose the parsed metadata.
- Callback-free authorization URLs remain unchanged.

Callbacks remain untrusted navigation hints. Authentication completion is still determined through the encrypted relay, and the client application remains responsible for deciding whether to open a callback destination. No homeserver behavior changes.

### API

Rust auth builders accept `XCallbackParams` through `.x_callback(...)`.

JavaScript supports callbacks through:

- `startCookieAuthFlow(caps, kind, relay, xCallback)`
- `startGrantAuthFlow(caps, kind, { clientId, relay, xCallback })`
- The `xCallback` getter on parsed deep-link objects.

### Acceptance criteria

- [x] Parse and serialize `x-source`, `x-success`, `x-error`, and `x-cancel`.
- [x] Match Ring’s exact-once decoding and literal-plus behavior.
- [x] Accept legacy `callback` as an `x-success` fallback.
- [x] Support both cookie and grant signin/signup flows.
- [x] Preserve callbacks when auth flows are saved and resumed.
- [x] Expose typed callback metadata through the JS/WASM bindings.
- [x] Keep existing callback-free auth flows unchanged.
- [x] Document callback security and relay-authoritative completion behavior.

### DIFF
--- a/pubky-sdk/src/actors/auth/deep_links/deep_link.rs
+++ b/pubky-sdk/src/actors/auth/deep_links/deep_link.rs
@@ -5,7 +5,7 @@ use url::Url;
 use crate::actors::auth::deep_links::{
     DEEP_LINK_SCHEMES, direct_signup::DirectSignupDeepLink, error::DeepLinkParseError,
     seed_export::SeedExportDeepLink, signin::SigninDeepLink, signin_grant::SigninGrantDeepLink,
-    signup::SignupDeepLink, signup_grant::SignupGrantDeepLink,
+    signup::SignupDeepLink, signup_grant::SignupGrantDeepLink, x_callback::XCallbackParams,
 };
 
 /// A parsed Pubky deep link.
@@ -30,6 +30,32 @@ pub enum DeepLink {
 }
 
 impl DeepLink {
+    /// Return the optional x-callback-url metadata carried by this deep link.
+    #[must_use]
+    pub const fn x_callback(&self) -> &XCallbackParams {
+        match self {
+            DeepLink::Signin(link) => link.x_callback(),
+            DeepLink::Signup(link) => link.x_callback(),
+            DeepLink::DirectSignup(link) => link.x_callback(),
+            DeepLink::SigninGrant(link) => link.x_callback(),
+            DeepLink::SignupGrant(link) => link.x_callback(),
+            DeepLink::SeedExport(link) => link.x_callback(),
+        }
+    }
+
+    /// Attach x-callback-url metadata to this deep link.
+    #[must_use]
+    pub(crate) fn with_x_callback(self, x_callback: XCallbackParams) -> Self {
+        match self {
+            Self::Signin(link) => Self::Signin(link.with_x_callback(x_callback)),
+            Self::Signup(link) => Self::Signup(link.with_x_callback(x_callback)),
+            Self::DirectSignup(link) => Self::DirectSignup(link.with_x_callback(x_callback)),
+            Self::SigninGrant(link) => Self::SigninGrant(link.with_x_callback(x_callback)),
+            Self::SignupGrant(link) => Self::SignupGrant(link.with_x_callback(x_callback)),
+            Self::SeedExport(link) => Self::SeedExport(link.with_x_callback(x_callback)),
+        }
+    }
+
     /// Convert the deep link to a simple URL.
     #[must_use]
     pub fn to_url(self) -> Url {
--- a/pubky-sdk/src/actors/auth/deep_links/typed_deep_link.rs
+++ b/pubky-sdk/src/actors/auth/deep_links/typed_deep_link.rs
@@ -88,7 +88,6 @@ where
     }
 
     /// Return the optional x-callback-url metadata carried by this deep link.
-    #[must_use]
     pub const fn x_callback(&self) -> &XCallbackParams {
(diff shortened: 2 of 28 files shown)


pwltr (dismissed):
LGTM

I checked this against Pubky Ring’s current x-callback parser behavior, and the implementation covers the same semantics: raw query extraction, exact-once decoding, literal `+` preservation, `x-source` / `x-success` / `x-error` / `x-cancel`, legacy `callback` fallback, and canonical `x-success` serialization.

There is remaining work downstream, wiring this through `pubky-core-ffi` and `react-native-pubky`, then Ring can consume the typed `xCallback` data directly. So utACK from me, also note I am not a Rust dev so would prefer a second approval regarding structure/quality.

coreyphillips (dismissed):

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
```diff
@@ -30,6 +30,32 @@ pub enum DeepLink {
...
+    pub(crate) fn with_x_callback(self, x_callback: XCallbackParams) -> Self {
+        match self {
+            Self::Signin(link) => Self::Signin(link.with_x_callback(x_callback)),
+            Self::Signup(link) => Self::Signup(link.with_x_callback(x_callback)),
+            Self::DirectSignup(link) => Self::DirectSignup(link.with_x_callback(x_callback)),
+            Self::SigninGrant(link) => Self::SigninGrant(link.with_x_callback(x_callback)),
+            Self::SignupGrant(link) => Self::SignupGrant(link.with_x_callback(x_callback)),
+            Self::SeedExport(link) => Self::SeedExport(link.with_x_callback(x_callback)),
+        }
+    }
```
andrei-21, line 57: We can simplify these two methods if we have one method which returns internal `link`. Then you have the big switch only once.

pubky-sdk/src/actors/auth/deep_links/typed_deep_link.rs
```diff
@@ -84,6 +87,19 @@ where
         &self.params
     }
 
+    /// Return the optional x-callback-url metadata carried by this deep link.
+    #[must_use]
```
andrei-21: I think we should not overuse `#[must_use]`. From the Rust [guideline](https://std-dev-guide.rust-lang.org/policy/must-use.html):
> "failing to explicitly consider [the value] or its output is almost certainly a bug"

pubky-sdk/src/actors/auth/deep_links/x_callback.rs
```diff
@@ -0,0 +1,242 @@
...
+        if !Self::has_valid_percent_triplets(value) {
+            return value.to_owned();
+        }
+
+        percent_decode_str(value)
+            .decode_utf8()
+            .map_or_else(|_| value.to_owned(), std::borrow::Cow::into_owned)
+    }
+
+    fn has_valid_percent_triplets(value: &str) -> bool {
```
andrei-21, line 117: Will the following implementation be simpler?
```rust
fn has_valid_percent_triplets(value: &str) -> bool {
    value.split('%').skip(1).all(|escape| {
        matches!(
            escape.as_bytes(),
            [first, second, ..] if first.is_ascii_hexdigit() && second.is_ascii_hexdigit()
        )
    })
}
```

pubky-sdk/src/actors/auth/deep_links/deep_link.rs
MCarlomagno: Mm I think that is not possible, because `link` can adopt different types depending on the variant of `DeepLink` (see `DeepLink` enum above)

andrei-21: Ah, sorry, I haven't looked that deep. Ok.

andrei-21 (approved):

PR #534, opened 2026-07-31
docs(homeserver): Cloudflare tunnel guide

86667:
- Adds hs deploy guide for getting setup with cloudflare tunnel
- Split deploy guide into 3 separate guides for each method: domain, ip-only and cloudflare tunnel
- Centralised prod notes, verification steps and troubleshooting

### DIFF
--- a/docs/deploy/cloudflare-tunnel.md
+++ b/docs/deploy/cloudflare-tunnel.md
@@ -1,3 +1,170 @@
 # Deploy with a Cloudflare Tunnel
 
-<!-- TODO: Write this guide -->
+This guide is for users who **cannot open ports** on their network or **don't have a static IP** — for example behind NAT, CGNAT, or a home router without port-forwarding access. Cloudflare proxies incoming HTTPS traffic through an outbound tunnel from your machine, so no inbound ports are needed.
+
+If you can open ports and have a static IP, consider [deploying with a domain](domain.md) or [deploying with an IP address](ip-only.md) instead — both support full Pubky TLS connectivity.
+
+This guide assumes you have already [installed the homeserver](../INSTALL.md).
+
+Commands and package names assume a Debian-based system (Ubuntu, Debian, etc.), adapt as needed for other distributions.
+
+## Requirements
+
+- A free [Cloudflare account](https://dash.cloudflare.com/sign-up)
+- A **domain name** with its DNS managed by Cloudflare (free plan is fine)
+- `cloudflared` installed on the server
+
+## Limitations
+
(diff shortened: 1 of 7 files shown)


docs/deploy/cloudflare-tunnel.md
```diff
@@ -0,0 +1,157 @@
...
+icann_domain = "YOUR_DOMAIN"
+
+# public_ip is not needed — Cloudflare proxies all traffic, so there is
+# no direct IP for Pubky TLS clients to connect to.
+
+# pubky_listen_socket can stay at the default (127.0.0.1:6287) since
+# Pubky TLS is not used with this setup.
+```
+
+Restart the homeserver after editing `config.toml`. See [Run](../INSTALL.md#run) in the install guide.
```
gcomte, line 83: This is not very straight-forward imo. For me the right command would be 'sudo systemctl restart pubky-homeserver' I guess. But that command isn't even mentioned in the INSTALL.md guide. So one is left with ambiguity and needs to search a bit for the right command.

docs/deploy/cloudflare-tunnel.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Run the Tunnel
+
+### Test manually first
+
+```bash
+cloudflared tunnel run pubky-homeserver
+```
+
+Verify that `https://YOUR_DOMAIN` responds (see [Verify](#verify-the-deployment) below). Press Ctrl+C to stop.
```
gcomte, line 95: 1. I did not have to press `Ctrl+C` .
2. Maybe you want to style  `Ctrl+C` as code (?)
3. Why do you mention the  `Ctrl+C` part here, while the rest is in a different section where the  `Ctrl+C` part is not mentioned?

Aside from that, I got a `HTTP 530`, which apparently means that Cloudflare is unable to reach my HS. Maybe we could document that in the verify section and give hints to what might have gone wrong. (BTW, curling without the -I flag returned `error code: 1033`, which is an error code from Cloudflare).

Edit: I was jumping around in the docs, so the reason it did not work is because I had killed the claudeflared tunnel (command just right above). After running that command again, it worked ✔️

docs/deploy/cloudflare-tunnel.md
86667: The thing is, it depends on which setup youve gone for. The options are either to link to the `Run` instructions which list all options, or list all options at every command. The command is in INSTALL with the systemd setup instructions:
```
Enable and start the service:

sudo systemctl daemon-reload
sudo systemctl enable pubky-homeserver
sudo systemctl start pubky-homeserver
```

Do you have an idea of how this could be better structured? We could have a separate guide for each setup option, but then we end up with repetition of many other parts. what do you think?

86667: Wait i see that `restart` isnt there lol

86667: I improvde this `Run` section with some service management commands, primarily adding `restart`

gcomte (approved):

PR #536, opened 2026-08-03
feat(homeserver): add support to path-addressed storage reads

### LINKED ISSUE #524: [homeserver] Add path-addressed storage read routes
## Parent

#172

## What to build

Add the client storage route:

`/storage/{user-z32}/{storage-path}`

Support GET, HEAD, and paginated directory listings while preserving the legacy owner-relative routes using `pubky-host`.

The route accepts only raw 52-character z32 keys. The path owner is authoritative:

- No `pubky-host`: process normally.
- `pubky-host`: accept as deprecated behavior, but ignored.

Both route forms must create the same internal `EntryPath` and use the existing storage implementation.

When `/storage` becomes canonical, mark the legacy storage routes and `pubky-host` parameter as deprecated in the API documentation. Keep legacy routing isolated from shared storage handlers so it can be removed cleanly later.

## Acceptance criteria

- [ ] GET, HEAD, and directory listing work through `/storage/{user-z32}/...` without `pubky-host`.
- [ ] Raw z32 keys are accepted.
- [ ] Invalid and `pubky`-prefixed keys return `400 Bad Request`.
- [ ] Existing pagination, authorization, caching, tracing, and rate-limit behavior is preserved.
- [ ] A `pubky-host` is accepted, but does not override the path owner.
- [ ] Legacy read routes continue working.
- [ ] Legacy and new routes resolve to the same internal `EntryPath`.
- [ ] A resource written through a legacy route can be read through the new route.
- [ ] No database, OpenDAL, or physical-storage migration is introduced.
- [ ] Unknown or malformed routes return `4xx`, never an internal missing-`PubkyHost` error.
- [ ] Legacy storage routes and `pubky-host` are marked deprecated in the API specification.
- [ ] Legacy route wiring is clearly isolated from the canonical `/storage` route.
- [ ] Shared handlers, `EntryPath`, and persistence code are not marked deprecated.

MCarlomagno:
Closes #524.

### Summary

Adds new APIs for storage reads:

- `GET /storage/{user_z32}/{path}`
- `HEAD /storage/{user_z32}/{path}`

The path owner is authoritative. `Host` remains the HTTP authority, while legacy `pubky-host` headers and query parameters are ignored by canonical routes. Legacy owner-relative routes remain supported for compatibility.

| | New path-addressed route | Legacy owner-relative route |
|---|---|---|
| Example | `GET /storage/{user_z32}/pub/file.txt` | `GET /pub/file.txt` |
| Storage owner | `{user_z32}` path segment | `pubky-host`, `Host`, or query fallback |
| `pubky-host` | Accepted but ignored | Used for owner resolution |
| Read methods | GET and HEAD | GET and HEAD |

Both route forms create the same internal `EntryPath` and use the existing persistence and OpenDAL implementation. No database or physical-storage migration is introduced.

### Acceptance criteria

- [x] GET, HEAD, and paginated directory listings work through `/storage/{user_z32}/...`.
- [x] The path owner cannot be overridden by legacy tenant inputs.
- [x] Legacy read and write routes continue working.
- [x] Legacy and canonical reads share the same storage handlers and `EntryPath`.
- [x] Data written through a legacy route can be read through the canonical route.
- [x] Authorization, caching, tracing, and rate-limit behavior is preserved.
- [x] Malformed canonical routes return `4xx`.
- [x] No database, OpenDAL, or physical-storage migration is required.
- [x] Legacy routes and parameters are documented as deprecated.

### DIFF
--- a/pubky-homeserver/src/client_server/middleware/request_tenant.rs
+++ b/pubky-homeserver/src/client_server/middleware/request_tenant.rs
@@ -17,7 +17,6 @@ use crate::shared::webdav::StoragePath;
 use super::pubky_host::extract_legacy_pubky;
 
 const STORAGE_ROUTE_PREFIX: &str = "/storage/";
-const RAW_PUBLIC_KEY_LENGTH: usize = 52;
 
 /// Tenant and optional owner-relative storage path resolved before auth runs.
 #[derive(Debug, Clone)]
@@ -35,25 +34,17 @@ impl RequestTenant {
         self.storage_path.as_ref()
     }
 
-    /// The path used for storage policy, independent of its HTTP route shape.
-    pub fn logical_path<'a>(&'a self, request_path: &'a str) -> &'a str {
-        self.storage_path
+    pub fn pubky_url(&self, request_uri: &Uri) -> String {
+        let storage_path = self
+            .storage_path
             .as_ref()
-            .map(StoragePath::as_str)
-            .unwrap_or(request_path)
-    }
-
-    pub fn logical_uri(&self, uri: &Uri) -> String {
-        let mut logical = format!(
-            "pubky://{}{}",
-            self.public_key.z32(),
-            self.logical_path(uri.path())
-        );
-        if let Some(query) = uri.query() {
-            logical.push('?');
-            logical.push_str(query);
+            .map_or(request_uri.path(), StoragePath::as_str);
+        let mut pubky_url = format!("pubky://{}{}", self.public_key.z32(), storage_path);
+        if let Some(query) = request_uri.query() {
+            pubky_url.push('?');
+            pubky_url.push_str(query);
         }
-        logical
+        pubky_url
     }
 
     fn from_request(req: &Request) -> Result<Option<Self>, String> {
@@ -61,10 +52,7 @@ impl RequestTenant {
             return Ok(Some(tenant));
         }
 
-        Ok(extract_legacy_pubky(req).map(|public_key| Self {
-            public_key,
-            storage_path: None,
-        }))
+        Ok(extract_legacy_pubky(req).map(Self::legacy))
     }
 
     /// Returns `Ok(None)` when the URL is not in the `/storage` namespace.
@@ -81,13 +69,6 @@ impl RequestTenant {
             .ok_or_else(|| "Missing storage path".to_string())?;
         if raw_storage_path.is_empty() {
             return Err("Missing storage path".to_string());
-        }
-
-        if PublicKey::is_pubky_prefixed(raw_public_key) {
-            return Err("Storage owner must be a raw z32 public key".to_string());
-        }
-        if raw_public_key.len() != RAW_PUBLIC_KEY_LENGTH {
-            return Err("Storage owner must be a 52-character z32 public key".to_string());
         }
 
         let public_key = PublicKey::try_from_z32(raw_public_key)
@@ -112,6 +93,8 @@ impl RequestTenant {
             }
             Ok(None) => next.run(request).await,
             Err(message) => {
+                // Tenant-aware limits require a resolved owner; malformed requests are
+                // rejected before authentication and storage access.
                 tracing::warn!(
                     method = %request.method(),
                     path = request.uri().path(),
@@ -123,7 +106,6 @@ impl RequestTenant {
         }
     }
 
-    #[cfg(test)]
     pub(crate) fn legacy(public_key: PublicKey) -> Self {
         Self {
             public_key,
@@ -164,7 +146,7 @@ mod tests {
         assert_eq!(tenant.public_key(), &owner);
         assert_eq!(tenant.storage_path().unwrap().as_str(), "/pub/example.txt");
         assert_eq!(
-            tenant.logical_uri(&format!("{path}?limit=10").parse().unwrap()),
+            tenant.pubky_url(&format!("{path}?limit=10").parse().unwrap()),
             format!("pubky://{}/pub/example.txt?limit=10", owner.z32())
         );
     }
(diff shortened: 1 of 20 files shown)


pubky-homeserver/src/client_server/middleware/request_tenant.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    pub(crate) async fn resolve(mut request: Request, next: Next) -> Response {
+        match Self::from_request(&request) {
+            Ok(Some(tenant)) => {
+                request.extensions_mut().insert(tenant);
+                next.run(request).await
+            }
+            Ok(None) => next.run(request).await,
+            Err(message) => {
```
andrei-21, line 95: From AI, sounds reasonable:
> Malformed canonical storage requests short-circuit outside the configured rate-limiting middleware, creating a discrete rate-limit bypass. The valid path-addressed read behavior otherwise appears consistent with the intended change.
Keep tenant-resolution errors behind rate limiting — `pubky-homeserver/src/client_server/middleware/request_tenant.rs:114-121`
>  When a `/storage/...` request has a malformed owner or path, this branch returns before the inner request-count and bandwidth limiters run because `RequestTenant::resolve` wraps the entire middleware stack. Consequently, deployments with IP/global limits can be flooded with malformed storage requests that bypass those limits and generate warning logs; defer the 400 until after applicable rate limiting or otherwise rate-limit this error path.

pubky-homeserver/src/client_server/middleware/request_tenant.rs
```diff
@@ -0,0 +1,222 @@
...
+            return Err("Missing storage path".to_string());
+        }
+
+        if PublicKey::is_pubky_prefixed(raw_public_key) {
+            return Err("Storage owner must be a raw z32 public key".to_string());
+        }
+        if raw_public_key.len() != RAW_PUBLIC_KEY_LENGTH {
+            return Err("Storage owner must be a 52-character z32 public key".to_string());
+        }
+
```
andrei-21: There is not need for the explicit checks, `try_from_z32()` does the job.

pubky-homeserver/src/client_server/middleware/request_tenant.rs
MCarlomagno: I think the observation is valid, but the impact is low. Malformed requests are rejected before authentication or storage access, and no global rate limit is enabled by default. 

The proposed fix (moving tenant resolution behind the existing limiters) would break the canonical requests because logical `/pub/` and user limits require the resolved tenant. A correct fix would add a separate IP limiter before tenant resolution - protecting from malformed requests without changing rate limit logic

IMO, given the limited impact, that added complexity doesnt seem to be justified for now, wdyt?

pubky-homeserver/src/client_server/middleware/request_tenant.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn from_request(req: &Request) -> Result<Option<Self>, String> {
+        if let Some(tenant) = Self::from_storage_route(req.uri().path())? {
+            return Ok(Some(tenant));
+        }
+
+        Ok(extract_legacy_pubky(req).map(|public_key| Self {
+            public_key,
+            storage_path: None,
+        }))
```
andrei-21: You can reuse `legacy()` constructor here.

pubky-homeserver/src/client_server/middleware/request_tenant.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// The path used for storage policy, independent of its HTTP route shape.
+    pub fn logical_path<'a>(&'a self, request_path: &'a str) -> &'a str {
+        self.storage_path
+            .as_ref()
+            .map(StoragePath::as_str)
+            .unwrap_or(request_path)
+    }
+
+    pub fn logical_uri(&self, uri: &Uri) -> String {
```
andrei-21: What does `logical` mean here? Maybe `pubky_url()`? Also confusing `uri` argument.

pubky-homeserver/src/client_server/middleware/request_tenant.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn public_key(&self) -> &PublicKey {
+        &self.public_key
+    }
+
+    pub fn storage_path(&self) -> Option<&StoragePath> {
+        self.storage_path.as_ref()
+    }
+
+    /// The path used for storage policy, independent of its HTTP route shape.
+    pub fn logical_path<'a>(&'a self, request_path: &'a str) -> &'a str {
```
andrei-21: This method is really confusing to me. What does `logical` mean in this context? Why are we passing `request_path` here? I don't think the `request_path` argument is necessary; in the actual use cases, you have to provide the `request_path` anyway if `RequestTenant` is `None`.

pubky-homeserver/src/client_server/middleware/request_tenant.rs
andrei-21: Sounds ok, maybe you can add a short comment to the code why we are fine with that?

pubky-homeserver/src/client_server/middleware/request_tenant.rs
MCarlomagno: so before this change, the endpoint path was the same as the logical path, now the endpoint path is `/storage/{user_public_key}/{storage_path}`. But maybe the confusing part is "logical", we can rename it to "storage_path" maybe?

Anyways, we can inline this function and make it explicit, is just one line

pubky-homeserver/src/client_server/middleware/request_tenant.rs
MCarlomagno: Agree on simplifying to URL, I pass the Uri providing the path + the query params to reconstruct the full URL, this is for traceability mostly

andrei-21 (approved):

PR #537, opened 2026-08-05
docs(sdk): Make in-code documentation more user friendly

86667:
The in-code docs for the SDK are used directly in docs.rs and now also our JS [typedoc page](https://pubky.github.io/pubky-core/js-sdk-typedoc). I aim here to fit them better for this purpose by:
- Provide user-friendly overviews of high level actors
- Add basic usage examples
- Reduce technical details which are not important for most users

### DIFF
--- a/pubky-sdk/src/actors/auth/kind.rs
+++ b/pubky-sdk/src/actors/auth/kind.rs
@@ -1,31 +1,43 @@
 use crate::PublicKey;
 
 /// The kind of authentication flow to perform.
+///
+/// Used when starting a [`PubkyGrantAuthFlow`](crate::PubkyGrantAuthFlow) or
+/// legacy cookie auth flow to tell the SDK whether the user already has an
+/// account (`SignIn`) or needs to create one (`SignUp`).
+///
+/// # When to use which
+///
+/// - **`SignIn`** — the user already signed up on a homeserver. The flow
+///   authenticates against the existing account.
+/// - **`SignUp`** — the user does not yet have an account. The flow creates
+///   the account on the specified homeserver as part of the auth handshake.
+///   Some homeservers require a `signup_token` (invite code).
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum AuthFlowKind {
     /// Sign in to an existing account.
     SignIn,
-    /// Sign up for a new account.
+    /// Sign up for a new account on a specific homeserver.
     SignUp {
         /// The public key of the homeserver to sign up on.
         homeserver_public_key: Box<PublicKey>,
-        /// The signup token to use for the signup flow.
-        /// This is optional.
+        /// Optional invite token required by some homeservers.
         signup_token: Option<String>,
     },
 }
 
 impl AuthFlowKind {
-    /// Create a sign in flow.
+    /// Create a sign-in flow kind.
     #[must_use]
     pub fn signin() -> Self {
         Self::SignIn
     }
 
-    /// Create a sign up flow.
+    /// Create a sign-up flow kind.
+    ///
     /// # Arguments
-    /// * `homeserver_public_key` - The public key of the homeserver to sign up on.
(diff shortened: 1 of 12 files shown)


pubky-sdk/src/actors/auth/kind.rs
```diff
@@ -1,31 +1,43 @@
...
+/// legacy cookie auth flow to tell the SDK whether the user already has an
+/// account (`SignIn`) or needs to create one (`SignUp`).
+///
+/// # When to use which
+///
+/// - **`SignIn`** — the user already signed up on a homeserver. The flow
+///   authenticates against the existing account.
+/// - **`SignUp`** — the user does not yet have an account. The flow creates
+///   the account on the specified homeserver as part of the auth handshake.
+///   Some homeservers require a `signup_token` (invite code).
```
andrei-21: Minor. I would swap items to have the logical order.

pubky-sdk/src/actors/session/core.rs
```diff
@@ -8,36 +8,52 @@ use super::credential::SessionCredential;
...
+/// - Provides [`storage()`](Self::storage) for path-based CRUD operations as this user.
+/// - Supports session lifecycle: [`revalidate()`](Self::revalidate) and
+///   [`signout()`](Self::signout).
+/// - Offers credential-specific views via [`as_grant()`](Self::as_grant) and
+///   [`as_cookie()`](Self::as_cookie) for grant management or secret export.
+///
+/// # Concurrency
+///
+/// `PubkySession` is **cheap to clone** and **thread-safe**. It shares the
+/// underlying [`PubkyHttpClient`] and credential state via `Arc`. Pass clones
```
andrei-21: We do not need to mention `Arc` here, it is an implementation detail.

pubky-sdk/src/pubky.rs
```diff
@@ -73,10 +73,54 @@ use crate::errors::RequestError;
...
+///
+/// # Actor overview
+///
+/// ```text
+///   Pubky::new()
+///     ├── .signer(keypair)          → PubkySigner        (signup / signin / approve auth)
+///     │     └── .signin(cid)        → PubkySession        (authenticated handle)
+///     │           └── .storage()    → SessionStorage       (put / get / delete / list)
+///     ├── .public_storage()         → PublicStorage        (read anyone's data, no keys)
+///     ├── .pkdns()                  → Pkdns                (resolve / publish homeserver records)
```
andrei-21: From AI, sounds legit:
> Do not advertise publishing through the read-only PKDNS actor —pubky-sdk/src/pubky.rs:94-94
  `Pubky::pkdns()` constructs `Pkdns::with_client` without a keypair and is explicitly read-only, so users following this overview cannot publish homeserver records through that actor; publishing requires `PubkySigner::pkdns()` (or another keypair-backed constructor). Describe this branch as resolution-only or point publishing to the signer branch.

pubky-sdk/src/actors/storage/list.rs
```diff
@@ -136,7 +136,10 @@ impl<'a> ListBuilder<'a> {
         self
     }
 
-    /// Resume listing from a previous `cursor` token.
+    /// Resume listing from a previous cursor.
+    ///
+    /// The cursor value comes from the last entry's path in a previous
+    /// listing response. Use it to paginate through large directories.
```
andrei-21: Again from AI, seems legit, in the tests we do like the AI says:
> Pass the full addressed resource as the list cursor — pubky-sdk/src/actors/storage/list.rs:139-142
    When callers follow this wording and pass `entries.last().unwrap().path.as_str()`, the homeserver rejects the cursor with `400 Invalid cursor` because it parses an `EntryPath` containing both the raw public key and path. Document that callers should pass the full addressed entry, such as its `to_pubky_url()` value, rather than only the `PubkyResource::path` field.

andrei-21 (approved):

PR #540, opened 2026-08-06
chore: increase docker build pipeline timeout

SpontaneousOverthrow:


86667 (commented):
Do we really expect this workflow to take more than 90 mins sometimes?

Edit: Answer: Yes, arm build before any cache have been populated.

86667 (approved):

PR #545, opened 2026-08-10
refactor(homeserver): All user-based operations to flow through `UserService`

86667:
User operations are currently done in a mix of `UserService` operations, direct `UserRepository` db calls and logic in caller functions. Here we route all user-based operations through `UserService`.

- `UserService` is now the single entry point for user CRUD, quota management, and admin operations
- Admin enable/disable logic centralised into `UserService::admin_disable`/`admin_enable`, replacing duplicated get-update-commit sequences in route handlers.
- `Finalizer` holds `SqlDb` separately from `UserService`, keeping non-user database work distinct from the user service's pool. Currently it uses `UserService`'s `pool()`, which is weird.

### DIFF
--- a/pubky-homeserver/src/admin_server/routes/disable_users.rs
+++ b/pubky-homeserver/src/admin_server/routes/disable_users.rs
@@ -1,69 +1,38 @@
 use super::super::app_state::AppState;
-use crate::{
-    persistence::sql::{uexecutor, user::UserRepository},
-    shared::{HttpError, HttpResult, Z32Pubkey},
-};
+use crate::shared::{HttpResult, Z32Pubkey};
 use axum::{
     extract::{Path, State},
     http::StatusCode,
     response::IntoResponse,
 };
 
-/// Delete a single entry from the database.
+/// Disable a user account.
 ///
 /// # Errors
 ///
 /// - `400` if the pubkey is invalid.
-/// - `404` if the entry does not exist.
+/// - `404` if the user does not exist.
 ///
 pub async fn disable_user(
     State(state): State<AppState>,
     Path(pubkey): Path<Z32Pubkey>,
 ) -> HttpResult<impl IntoResponse> {
-    let mut tx = state.sql_db.pool().begin().await?;
-    let mut user = match UserRepository::get(&pubkey.0, uexecutor!(tx)).await {
-        Ok(user) => user,
-        Err(sqlx::Error::RowNotFound) => {
-            return Err(HttpError::new_with_message(
-                StatusCode::NOT_FOUND,
-                "User not found",
-            ))
-        }
-        Err(e) => return Err(e.into()),
-    };
-    user.disabled = true;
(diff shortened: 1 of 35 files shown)


pubky-homeserver/src/client_server/auth/signup_service.rs
MCarlomagno, line 1: `ensure_user_not_exists` uses a repo method instead of a service one, I'm ok if that's intended but flagging just in case

pubky-homeserver/src/persistence/files/file/file_service.rs
```diff
@@ -154,11 +152,10 @@ mod tests {
         let context = AppContext::test().await;
         let file_service = FileService::new_from_context(&context).unwrap();
         let db = context.sql_db.clone();
```
ok300: `cargo clippy --tests` finds a few `unused_variable` warnings.

This is because `db` is initialized, but now the `user_service` is used with calls that don't reference `db` anymore.

IMO in those cases, the `db` assignment can go:

```rust
let db = context.sql_db.clone();
```

pubky-homeserver/src/client_server/auth/signup_service.rs
86667: Good spot - fixed. Also made some extra small improvements to docs and tests.

pubky-homeserver/src/admin_server/routes/disable_users.rs
```diff
@@ -1,69 +1,38 @@
...
 /// # Errors
 ///
 /// - `400` if the pubkey is invalid.
-/// - `404` if the entry does not exist.
+/// - `404` if the user does not exist.
 ///
 pub async fn disable_user(
     State(state): State<AppState>,
     Path(pubkey): Path<Z32Pubkey>,
 ) -> HttpResult<impl IntoResponse> {
```
ok300, line 19: NIT: Admin 404 body text changed

Old code returned `HttpError::new_with_message(StatusCode::NOT_FOUND, "User not found")`. `UserService::set_disabled` returns bare `HttpError::not_found()`, whose detail is "Not Found" (`shared/http_error.rs:33`). Status code is unchanged, only the response body detail.

Not sure if intentional. IMO not a big issue, even if previous msg was clearer.

pubky-homeserver/src/persistence/files/file/file_service.rs
ok300: You addressed this just few minutes ago :)

pubky-homeserver/src/admin_server/routes/disable_users.rs
86667: This is to make "user not found" error consistent across the service. Otherwise its different error messages for different fns for the same error. 

I figured this is fine to change since its only the message changed slightly and an admin route which is rarely used.

ok300 (approved):
LGTM

PR #543, opened 2026-08-07
feat(homeserver): add path addressed storage writes

MCarlomagno:
Closes #525.

### Summary

Adds authenticated storage writes to the path-addressed API:

- `PUT /storage/{user_z32}/{path}`
- `DELETE /storage/{user_z32}/{path}`

| | Path-addressed route | Legacy owner-relative route |
|---|---|---|
| Example | `PUT /storage/{user_z32}/pub/file.txt` | `PUT /pub/file.txt` |
| Storage owner | `{user_z32}` path segment | `pubky-host`, `Host`, or query fallback |
| `pubky-host` | Accepted but ignored | Used for owner resolution |
| Write methods | PUT and DELETE | PUT and DELETE |

Both route forms create the same internal `EntryPath` and use the existing write, quota, metadata, event, and persistence implementation. Legacy handlers are explicitly isolated behind `legacy_*` names so the compatibility route and its tests can be removed together later.

The owner-relative GET, HEAD, PUT, and DELETE operations and their `pubky-host` parameters are documented as deprecated. New integrations should use `/storage/{user_z32}/{path}`.

No database, OpenDAL, or physical-storage migration is introduced.

### Acceptance criteria

- [x] PUT and DELETE work through `/storage/{user_z32}/{path}` with grant/bearer authentication.
- [x] PUT and DELETE work with cookie authentication during the transition.
- [x] Cookie selection and authorization use the owner from the path.
- [x] The authenticated identity and capabilities are checked against the path owner and storage path.
- [x] Anonymous, under-scoped, and cross-tenant writes are rejected.
- [x] Directory-shaped paths are rejected for PUT and DELETE.
- [x] Quotas, metadata, and PUT/DEL event behavior use the existing storage pipeline.
- [x] Resources written through either route can be read and deleted through the other.
- [x] Legacy storage writes continue working with `pubky-host`.
- [x] Cookie `/signup` and `/session` compatibility remains unchanged.
- [x] Legacy endpoints and parameters are documented as deprecated.
- [x] No database, OpenDAL, or physical-storage migration is required.

PR #546, opened 2026-08-10
feat(sdk): add support to path addressed requests in sdk

### LINKED ISSUE #527: [sdk] Use path-addressed URLs for storage requests
## Parent

#172

## What to build

Update native and WASM SDK transports to convert logical Pubky addresses and session-relative storage paths into `/storage/{user-z32}/{storage-path}` requests.

For example:

`pubky://<user-z32>/pub/example.txt`

becomes:

`https://<resolved-authority>/storage/<user-z32>/pub/example.txt`

Only storage transport changes. Public SDK addresses and session-relative APIs remain unchanged. Deprecated cookie-session operations continue using `pubky-host`.

## Deprecation notice

Complete Rust `#[deprecated]` and JavaScript `@deprecated` annotations for public cookie-auth and legacy-addressing APIs. Generic session and storage APIs that also support grant authentication remain non-deprecated.

## Acceptance criteria

- [ ] Public-storage reads generate `/storage/{user-z32}/...` URLs.
- [ ] Session storage reads, listings, writes, and deletes generate the new URLs.
- [ ] Explicit-homeserver requests retain the resource owner in the path.
- [ ] Native and WASM storage requests no longer send `pubky-host`.
- [ ] Cookie `/signup` and `/session` compatibility remains operational.
- [ ] Logical `pubky://` and session-relative SDK APIs remain unchanged.
- [ ] The SDK documents its minimum compatible homeserver version.
- [ ] Tests cover old SDK/new homeserver compatibility.
- [ ] Tests cover new SDK/new homeserver behavior.
- [ ] Migration documentation states that new SDKs require a homeserver with `/storage` support.
- [ ] Public legacy APIs have consistent Rust and JavaScript deprecation annotations.
- [ ] Deprecation messages identify the replacement.
- [ ] Shared grant-compatible APIs are not marked deprecated.

## Blocked by

- https://github.com/pubky/pubky-core/issues/524
- https://github.com/pubky/pubky-core/issues/525

MCarlomagno:
Closes #527.

### Summary

Updates the Rust and JavaScript SDKs to use the path-addressed storage routes introduced in #536 and #543:

```text
https://_pubky.{owner}/storage/{owner}/{path}
```

| | Path-addressed transport | Legacy transport |
|---|---|---|
| Example | `/storage/{owner}/pub/file.txt` | `/pub/file.txt` |
| Storage owner | `{owner}` path segment | `pubky-host` |
| `pubky-host` | Omitted | Sent |
| SDK support | Generated by new SDKs | Accepted when parsing transport URLs |

`resolve_pubky`, `resolvePubky`, and `PubkyResource::to_transport_url` now produce the path-addressed form. `PubkyResource::from_transport_url` accepts both forms, taking the owner from the path for canonical URLs even when the request uses an explicit homeserver authority.

### Acceptance criteria

- [x] Public and session storage operations use `/storage/{owner}/...`.
- [x] `resolve_pubky`, `resolvePubky`, and `to_transport_url` return path-addressed URLs.
- [x] Canonical and legacy transport URLs remain parseable.
- [x] The path owner is authoritative when an explicit homeserver authority is used.
- [x] Native, WASM, and JavaScript storage requests omit `pubky-host`.
- [x] Cookie signup and session requests retain `pubky-host`.
- [x] Logical Pubky addresses and session-relative APIs remain unchanged.
- [x] Old-SDK/new-homeserver interoperability uses the legacy `pubky-host` wire shape.
- [x] Cookie-only compatibility APIs identify their grant-backed replacements.
- [x] Shared grant-compatible APIs remain non-deprecated.
- [x] SDK documentation and transport URL examples use the path-addressed form.


pubky-sdk/src/actors/storage/resource.rs
```diff
@@ -220,39 +221,35 @@ impl PubkyResource {
     /// - Returns [`Error::Request`] if the constructed transport URL is invalid.
     pub fn to_transport_url(&self) -> Result<Url, Error> {
         let rel = self.path.as_str().trim_start_matches('/');
```
andrei-21: This line appears in multiple methods. You didn't really touched this code, but would it make sense to encapsulate or make it somehow reusable?.

pubky-sdk/src/actors/storage/resource.rs
```diff
@@ -210,7 +211,7 @@ impl PubkyResource {
         format!("pubky://{}/{}", self.owner.z32(), rel)
     }
 
-    /// Render as `https://_pubky.<owner>/<abs-path>` for transport.
+    /// Render as `https://_pubky.<owner>/storage/<owner>/<abs-path>` for transport.
     ///
     /// This converts the addressed resource into the actual homeserver URL used
```
andrei-21: This comment and the implementation confuse me. By the actual homeserver URL I would expect homeserver's host or something. Maybe I miss something in terminology here?..

pubky-sdk/src/actors/storage/resource.rs
```diff
@@ -220,39 +221,35 @@ impl PubkyResource {
     /// - Returns [`Error::Request`] if the constructed transport URL is invalid.
     pub fn to_transport_url(&self) -> Result<Url, Error> {
         let rel = self.path.as_str().trim_start_matches('/');
-        let https = format!("https://_pubky.{}/{}", self.owner.z32(), rel);
+        let owner = self.owner.z32();
+        let https = format!("https://_pubky.{owner}/storage/{owner}/{rel}");
```
andrei-21: AI claims that `session.storage().list("/")`  will not work anymore, because the server allows to list `/pub` or `/priv`. Is it intentional?

pubky-sdk/src/actors/storage/resource.rs
MCarlomagno: yes right, I didn't notice the confusing wording, I will replace by something like

"/// Converts the resource into its canonical transport URL"

pubky-sdk/src/actors/storage/resource.rs
MCarlomagno: Yes intended, since we implemented `/priv` / `/pub` paths. 
Maybe it wasn't explicit in the sdk before, but the rule is established in the homeserver [here](https://github.com/pubky/pubky-homeserver/pull/448) and [here](https://github.com/pubky/pubky-homeserver/issues/434) is the reason

pubky-sdk/src/client/http_targets/storage.rs
```diff
@@ -0,0 +1,224 @@
...
+        let transport_host = classify_transport_host(url.host_str().unwrap_or_default())?;
+        if transport_host == TransportHost::Other {
+            return Ok(StorageAddressing::Standard);
+        }
+        let (owner, path) = path
+            .split_once('/')
+            .ok_or_else(|| RequestError::Validation {
+                message: "path-addressed storage URL is missing a resource path".to_string(),
+            })?;
+        let owner = PublicKey::try_from_z32(owner).map_err(|_error| RequestError::Validation {
```
andrei-21, line 41: We also should reject URLs like `https://_pubky.<A>/storage/<B>/...` where A != B.

pubky-sdk/src/actors/storage/resource.rs
```diff
@@ -206,60 +212,76 @@ impl PubkyResource {
...
+    /// It is the same mapping performed by [`resolve_pubky`].
     ///
     /// # Errors
     /// - Returns [`Error::Request`] if the constructed transport URL is invalid.
     pub fn to_transport_url(&self) -> Result<Url, Error> {
-        let rel = self.path.as_str().trim_start_matches('/');
-        let https = format!("https://_pubky.{}/{}", self.owner.z32(), rel);
+        let owner = self.owner.z32();
+        let path = self.path.as_root_relative_str();
+        let https = format!("https://_pubky.{owner}/storage/{owner}/{path}");
```
andrei-21, line 229: From AI:
The SDK now has two storage URL formats:

- Old homeservers expect:
  `https://_pubky.<user>/pub/file.txt`
- New homeservers support:
  `https://_pubky.<user>/storage/<user>/pub/file.txt`

The high-level SDK methods, such as `storage().get(...)`, first ask the homeserver which format it supports. If it is old, the SDK rewrites the URL to the legacy format. So those methods work correctly.

The problem is that public helpers like `to_transport_url()` always return the new format. If someone passes that URL to a normal HTTP client—or the public synchronous `PubkyHttpClient::request()`—the compatibility check never happens. An old homeserver receives `/storage/...`, does not recognize it, and the request fails.

So, in short:

> The SDK knows how to support old servers, but that support only exists in the high-level request path. Publicly generated URLs bypass it.

This only affects callers using raw URLs or the low-level request API. Users of `storage().get()`, `put()`, etc. are already covered. The likely fixes are either to keep public URLs in the old compatible format or make the public request API perform the same feature negotiation.

andrei-21 (approved):

PR #548, opened 2026-08-12
feat: add `Retry-After` header in HTTP 429 responses

### LINKED ISSUE #407: HS rate limit response (HTTP 429) should include `Retry-After` header
Currently, the caller has to guess and approximate how many times to retry, how often, for how long, etc.

A header in the 429 response from the HS

```
Retry-After: <delay-seconds>
```

would massively simplify this.

This affects Nexus and other clients that talk to the HS directly (probably event Pubky App).

ok300:
This PR adds a `Retry-After` header to HTTP 429 "Too Many Requests" errors, which are thrown when the API reaches a rate-limit.

Fixes #407

### DIFF
--- a/pubky-homeserver/src/client_server/middleware/rate_limiter/request_rate_limit.rs
+++ b/pubky-homeserver/src/client_server/middleware/rate_limiter/request_rate_limit.rs
@@ -3,10 +3,11 @@
 //! Enforces per-path request-count quotas (`[[drive.rate_limits]]` in config).
 //! Each path+method pattern has a governor rate limiter keyed by IP or user.
 
+use axum::http::header::RETRY_AFTER;
 use axum::response::{IntoResponse, Response};
 use axum::{
     body::Body,
-    http::{HeaderName, Request, StatusCode},
+    http::{Request, StatusCode},
 };
 use futures_util::future::BoxFuture;
 use governor::clock::Clock;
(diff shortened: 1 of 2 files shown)


pubky-homeserver/src/client_server/middleware/rate_limiter/request_rate_limit.rs
```diff
@@ -122,26 +124,46 @@ fn check_request_count_limits(limits: &[LimitTuple], req: &Request<Body>) -> Res
...
 
+/// Build a `429 Too Many Requests` response carrying a `Retry-After: <seconds>` header.
+///
+/// The value is a wait time in seconds, rounded up to the nearest second with a minimum of 1.
+///
+/// See also:
+/// - https://www.rfc-editor.org/info/rfc6585/#section-4
+/// - https://www.rfc-editor.org/info/rfc9110/#section-10.2.3
+fn rate_limited_response(retry_after: Duration) -> Response {
+    let temp_secs = retry_after.as_secs();
```
86667: Nit: this code looks like somehting which could be expressed in a single line of code

86667: I also wonder if this logic lives in this function - `rate_limited_response` implies "just build the response".
Is this function even necessary?

ok300: Good points, addressed both with d31368bcd7ef53c34629777a38ca16cdafa4bf3d

86667 (approved):

PR #550, opened 2026-08-12
feat(homeserver): feature discovery endpoint

### LINKED ISSUE #549: Expose feature discovery endpoint in client server
## Problem

SDKs need a reliable way to discover which protocol features a homeserver supports.

Inferring support from a failed feature request is unreliable. Older homeservers route unknown paths through the legacy `/{*path}` wildcard, so a missing endpoint may return authentication, extraction, or validation errors instead of a normal `404`.

Exposing the exact homeserver version would also couple SDK behavior to release versions and unnecessarily disclose deployment information.

## Proposal

Add an unauthenticated `GET /info` endpoint to the client server:

```json
{
  "features": []
}
```

The endpoint describes client-facing protocol features using stable string identifiers. It initially returns an empty list; supported features will be advertised separately as their SDK integrations are implemented.

This endpoint is separate from the existing password-protected admin `/info`, which contains operational statistics and the exact homeserver version.

## Behavior

- `GET /info` returns `200 OK` with an `application/json` response.
- `features` is always an array of strings.
- The initial response contains no features.
- The endpoint is homeserver-scoped and does not require a tenant, `pubky-host`, cookie, or bearer credential.
- It is available through both PubkyTLS and ICANN transports.
- It does not expose the homeserver version, build information, usage statistics, configuration, or other admin data.
- Future clients must ignore unknown feature identifiers.
- If `/info` is unavailable, returns a non-success status, or contains an invalid response, clients treat the homeserver as supporting no advertised features.

## Compatibility

Adding the endpoint does not change existing client-server or admin-server behavior.

Older homeservers do not implement `/info` and may return errors other than `404` because of the legacy wildcard route. Treating any discovery failure as an empty feature set provides a safe compatibility default.

## Acceptance Criteria

- An unauthenticated client-server `GET /info` returns exactly:

  ```json
  {
    "features": []
  }
  ```

- The response content type is JSON.
- The request succeeds without tenant or authentication headers.
- The existing protected admin `/info` response remains unchanged.
- Tests cover the response schema and confirm that authentication and tenant resolution are not required.

## Out of Scope

- Specific feature flags, we will add them as we need.
- SDK-side discovery, caching, or endpoint selection.
- Exposing the homeserver version publicly.
- Changing the existing admin `/info` endpoint.

MCarlomagno:
Closes #549.

### Summary

Adds a public `GET /info` endpoint to the client server:

```json
{
  "features": []
}
```
The endpoint exposes stable, opaque feature identifiers that SDKs can use to detect homeserver capabilities. It initially advertises no features.

### DIFF
--- a/pubky-homeserver/src/client_server/app.rs
+++ b/pubky-homeserver/src/client_server/app.rs
@@ -31,7 +31,7 @@ use super::middleware::{
     request_tenant::RequestTenant,
     trace::with_trace_layer,
 };
-use super::routes::{events, root, signup_tokens, tenants};
+use super::routes::{events, info, root, signup_tokens, tenants};
 
 /// Errors that can occur when building a `HomeserverCore`.
 #[derive(Debug, thiserror::Error)]
@@ -248,7 +248,9 @@ pub fn create_app(
         .with_state(state)
         .merge(auth::base_router(auth_state.clone()))
         .merge(auth::tenant_router(auth_state))
-        .layer(middleware);
+        .layer(middleware)
+        // Keep feature discovery independent of authentication and database-backed quotas.
+        .route("/info", get(info::get));
 
     // Resolve the target before tracing and authentication. Valid `/storage/...`
     // requests are therefore logged using their Pubky URL.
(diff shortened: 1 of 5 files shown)


pubky-homeserver/src/client_server/routes/info.rs
```diff
@@ -0,0 +1,24 @@
+//! Client-facing homeserver feature information.
+
+use axum::{
+    http::{header, StatusCode},
+    response::IntoResponse,
+    Json,
+};
+use serde::Serialize;
+
+const FEATURES: &[&str] = &[];
```
MCarlomagno, line 10: the idea is to add static strings describing features in this list when they might not be compatible with older versions of the sdk

pubky-homeserver/src/client_server/app.rs
```diff
@@ -248,7 +248,9 @@ pub fn create_app(
         .with_state(state)
         .merge(auth::base_router(auth_state.clone()))
         .merge(auth::tenant_router(auth_state))
-        .layer(middleware);
+        .layer(middleware)
+        // Keep feature discovery independent of authentication and database-backed quotas.
+        .route("/info", get(info::get));
```
MCarlomagno, line 253: request avoids the database, storage, and authentication and returns a tiny static response

andrei-21 (approved):

PR #554, opened 2026-08-14
refactor(homeserver): Services own their construction from `AppContext`

86667:
Services that depend on `AppContext` now construct themselves via `from_context(&AppContext)` instead of callers manually extracting and passing individual fields. I've found the whole `AppContext` and `AppState` pattern slightly confusing in the past, i hope that this will make things clearer:
- `AppContext` holds shared-state services which require a single shared instance
- Each `AppState` holds all of `AppContext` plus any extra services/functionality which only that particular server needs
- Services without state are instantiated by their parent service  

This means that a service can be added without having to wire it through each layer, config can be accessible directly from `state` and a `*Service::from_context()` pattern can be standardised throughout.

I'm ultimately attempting to simplify at the cost of making service dependencies less explicit. 

Also adds convenience `test_with_config()` for creating `AppContext` with config overrides.

### DIFF
--- a/pubky-homeserver/src/client_server/app_state.rs
+++ b/pubky-homeserver/src/client_server/app_state.rs
@@ -1,40 +1,20 @@
 use axum::extract::FromRef;
 
 use crate::client_server::auth::AuthState;
-use crate::observability::Metrics;
-use crate::persistence::files::events::EventsService;
-use crate::persistence::files::FileService;
-use crate::persistence::sql::SqlDb;
-use crate::services::user_service::UserService;
-use crate::SignupMode;
+use crate::AppContext;
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub(crate) struct AppState {
     /// Auth sub-state (extracted via `FromRef` by auth handlers).
     pub(crate) auth_state: AuthState,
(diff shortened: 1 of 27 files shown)


pubky-homeserver/src/client_server/app_state.rs
```diff
@@ -1,27 +1,22 @@
...
-    pub(crate) sql_db: SqlDb,
-    pub(crate) file_service: FileService,
-    pub(crate) signup_mode: SignupMode,
-    pub(crate) events_service: EventsService,
-    pub(crate) metrics: Metrics,
-    /// User service for user lookups, creation, and cache access.
-    pub(crate) user_service: UserService,
-    /// Default per-user storage quota in MB (from `[storage].default_quota_mb`).
-    pub(crate) default_storage_mb: Option<u64>,
+    pub(crate) context: AppContext,
```
MCarlomagno: I think this should be `Arc<AppContext>` in order to avoid cloning the entire context on every request

86667: Updated.

MCarlomagno (approved):
LGTM

PR #564, opened 2026-08-20
refactor(homeserver): Services to own `sql_db`

86667:


pubky-homeserver/src/persistence/sql/pg_event_listener.rs
```diff
@@ -295,6 +295,7 @@ mod tests {
 
     /// Helper: create a real event in the DB and send a NOTIFY to wake the listener.
     async fn create_event_and_notify(
+        events_service: &EventsService,
```
ok300, line 298: Since `EventsService` now has a DB handle, wouldn't it make sense to remove the `db: &SqlDb` arg just below?

ok300: In a similar vein, there are a few more methods asking for both "`EventsService` + DB handle", where IMO the DB handle is now superfluous:

- `PgEventListener::start`
- `PgEventListener::start_with_poll_interval`
- `WriteFinalizationLayer::new`
- `Finalizer::new`
- `opendal_service::build_storage_operators`
- `OpendalService::new_from_config`
- `FileService::new_from_config`
- `tests::create_event_and_notify`

They could all simplify their signature (get rid of the DB handle) and internally rely on a new `EventsService` getter:

```rust
 pub(crate) fn sql_db(&self) -> &SqlDb {
        &self.sql_db
    }
```

then depending on the needed type, reconstruct it from there:

```rust
let db : &SqlDb = events_service.sql_db();

let db : SqlDb = events_service.sql_db().clone();

let pool : PgPool = events_service.sql_db().pool().clone();
```

What do you think?

86667: Thanks for the suggestion. I feel i could have explained these changes better in the PR description!

What im going for here is the opposite of what you suggest: If a service needs an `sql_db` for its own purposes, then it should own a `sql_db` itself. what you suggest here is using `EventsService` for general database access, meaning that services rely on `EventsService` for non-events related operations. 

I do agree that we are now passing around sql_db a lot and the fn signatures are arguably less clean than before. Im not sure on a clean way to achieve the same de-coupling of services without cloning sql_db a lot at initialisation though. 

what are your thoughts?

ok300: I see, alright makes sense.

I thought the goal here was simplification / cleanup. The way you describe it is conceptually cleaner though.

ok300 (approved):

PR #573, opened 2026-08-26
fix: sort configured rate-limits before evaluating

ok300:
Due to a known behavior of `governor`, a successful limiter check consumes quota immediately, even if a later limiter rejects the same request.

When overlapping rate limits are evaluated with a longer window first, a request that is subsequently rejected by a shorter-window limit can therefore consume quota from the slower-to-recover limit.

For example, given `1000r/d` followed by `10r/s`, a large burst can consume the daily quota before the per-second limit rejects most of the requests.

## Summary

This PR sorts configured `[[drive.rate_limits]]` entries by quota time unit before creating the `governor` limiters.

Shorter-window limits are now evaluated before longer-window limits:

`seconds → minutes → hours → days`

This ensures that a shorter-window rejection does not first consume quota from a longer-window limiter, regardless of the order in which those limits appear in TOML.

Limits with the same time unit retain their original TOML order; no additional canonical tie-breaking is performed.

## Example protected by this change

Given these overlapping limits, configured in the harmful order:

```toml
# POST /upload, per IP
2r/m
1r/s
```

Request sequence:

1. The first request succeeds and consumes one token from both limits.
2. An immediate second request is rejected by `1r/s`.
   - **Before:** `2r/m` was checked first and consumed its second token before `1r/s` rejected the request.
   - **After:** `1r/s` is checked first, so the rejected request does not consume another `2r/m` token.
3. After waiting approximately 1.1 seconds:
   - **Before:** the third request remains blocked because `2r/m` was exhausted.
   - **After:** the third request succeeds because the second `2r/m` token was preserved.

### DIFF
--- a/pubky-homeserver/src/client_server/middleware/rate_limiter/request_rate_limit.rs
+++ b/pubky-homeserver/src/client_server/middleware/rate_limiter/request_rate_limit.rs
@@ -11,7 +11,7 @@ use axum::{
 };
 use futures_util::future::BoxFuture;
 use governor::clock::Clock;
-use std::{convert::Infallible, task::Poll};
+use std::{convert::Infallible, task::Poll, time::Duration};
 use tower::{Layer, Service};
 
 use crate::shared::quota::PathLimit;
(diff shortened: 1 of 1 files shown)


86667 (commented):
Okay, so now a longer window quota is protected from a shorter window rejection, but a shorter window quota is still used up for a longer window rejection. This is fine because we expect that by the time the longer window is refreshed, the shorter window would be too. Is this correct?

Edit: Just read your issue ticket and diagram, the answer to the above question is "yes".

I do wonder what the use case is for wanting to set two windows on the same endpoint, especially since we allow for bursts.

86667 (approved):

PR #574, opened 2026-08-28
refactor(homeserver)!: Removed `DataDir` trait 

86667:
Removes the `DataDir` trait and its mock implementation, replacing it with explicit parameter passing into `AppContext::new()`. The purpose is to make constructing a homeserver more flexible and easier to follow: no trait objects, and no configuration assembled across multiple files.

Along the way this fixes three bugs that the old indirection was hiding. Production behaviour is unchanged.

## `AppContext` construction

`AppContext::new()` now takes the four things a context actually needs, each resolved independently by the caller:

```rust
AppContext::new(data_path, config, keypair, db_mode, pkarr_builder)
```

with three convenience constructors.

## The `database_url` default moved into code

`config.default.toml` used to ship `database_url = "postgres://localhost:5432/pubky_homeserver"`, and that file is merged *underneath* every config read from disk — so `[general].database_url` came back `Some(...)` for every server, whether or not anyone had chosen a database.

That destroys the only thing the `Option` is good for: telling "the operator chose this" apart from "nobody chose". The database-resolution rule below reads exactly that distinction, so the default now lives in code as `DEFAULT_DATABASE_URL` and is applied by `DatabaseMode::direct_or_default`.

## Bugs fixed

There were a few bugs fixed along the way:

- **Testnet homeservers were joining the public DHT.**: Network isolation was gated on `database_url.is_none()`, but every testnet path set `database_url`, so the gate never fired and testnet homeservers built against the public DHT and mainnet relays.
- **`StaticTestnet` persistent mode never read `TEST_PUBKY_CONNECTION_STRING`.**: `TestnetDataDir` only did `postgres_connection_string.or(config.database_url)`, and `Testnet::new()` leaves that `None`, so the env var was ignored despite the README claiming otherwise.
- **A bad `dht_relay_nodes` entry panicked at startup.**:  `builder.relays(...).expect("parameters are already urls and therefore valid.")` is now `AppContextBuildError::RelayNodes`. A `url::Url` is not necessarily a valid relay, and it  comes straight from an operator-edited config file.

## Breaking changes

**`pubky-homeserver`**

| Removed / renamed | Replacement |
| --- | --- |
| `DataDir` trait | — (pass parameters explicitly) |
| `MockDataDir` | `AppContext::new_ephemeral` |
| `AppContext::read_from` | `from_persistent_dir` / `new_ephemeral` / `new` |
| `AppContextConversionError` | `AppContextBuildError` |
| `HomeserverApp::start_with_mock_data_dir` | `HomeserverApp::start(ctx)` |
| `ClientServer::start_with_mock_data_dir` | `ClientServer::start(ctx)` |
| `AdminServer::from_mock_dir` | `AdminServer::start(ctx)` |
| `DatabaseMode::resolve_test(explicit)` | `resolve_test(override, from_config)` |

**`pubky-testnet`**

`Testnet::create_homeserver_app_with_mock(mock_dir)` → `Testnet::create_homeserver_with(config, keypair)`.

## Deferred review threads from #500

Two threads on #500 were left open with "will be addressed in the follow-up refactor".
This is that refactor.

- [`r3844757914`](https://github.com/pubky/pubky-homeserver/pull/500#discussion_r3844757914)
  — pkarr isolation was gated on `database_url.is_none()`, so a context built with an
  explicit `database_url` could publish test records to the public DHT. The gate is gone;
  isolation is now an explicit builder the caller passes. The reported scenario was in fact
  wider than described: every testnet path set `database_url`, so the gate never fired for
  any of them.
- [`r3844908580`](https://github.com/pubky/pubky-homeserver/pull/500#discussion_r3844908580)
  — `persist` ignored `TEST_PUBKY_CONNECTION_STRING` and fell back to a config default with
  no credentials and the wrong database name. Both halves are fixed: the persistent path
  resolves through the shared rule, and that default no longer exists.

## Related issues

Deliberately **not** fixed here, to keep this PR to the refactor:

- **#602** — pkarr network isolation is a property of the (builder, config) *pair*, not an
  enforced invariant. This PR removes the broken gate and documents the invariant, but does
  not encode it in the type system. `Testnet::create_homeserver_with` now clears
  `dht_relay_nodes` unconditionally, which narrows the blast radius without closing the
  design issue.
- **#603** — `persist` treats `TEST_PUBKY_CONNECTION_STRING` as *the database*, while every
  other mode treats it as *a server* to create an ephemeral database on. This is a
  consequence of unifying resolution in this PR rather than a pre-existing bug: following
  the documented `.../postgres` convention now puts the persistent testnet's tables in the
  maintenance database. Worth reading before merging, as the README here tells people to do
  exactly that.
- **#601** — ephemeral `pubky_test_{uuid}` databases can be orphaned when the cleanup pass
  runs before they are queued for deletion. A fix for this was in this branch and was
  stripped in `0640de56` to keep the diff focused.


pubky-homeserver/src/app_context.rs
```diff
@@ -97,182 +110,261 @@ pub struct AppContext {
...
-        let builder_debug = format!("{builder:?}");
+    fn public_mode_keeps_defaults_and_applies_config() {
+        use crate::DomainPort;
+        use std::str::FromStr;
+
+        let mut config = ConfigToml::default_test_config();
+        config.pkdns.dht_bootstrap_nodes =
+            Some(vec![DomainPort::from_str("127.0.0.1:6881").unwrap()]);
+
+        let builder = AppContext::resolve_dht_mode(DhtMode::Public, &config);
```
ok300: Hmm is this really still `DhtMode::Public` if the config sets a local DHT bootstrap node?

pubky-homeserver/src/app_context.rs
```diff
@@ -97,182 +110,261 @@ pub struct AppContext {
...
+            }
+            DhtMode::Custom(builder) => builder,
         }
-        if let Some(bootstrap_nodes) = &config_toml.pkdns.dht_bootstrap_nodes {
+    }
+
+    /// Apply DHT configuration (bootstrap nodes, relays, timeouts) from config
+    /// to a pkarr client builder.
+    fn apply_config_to_pkarr(builder: &mut pkarr::ClientBuilder, config: &ConfigToml) {
+        if let Some(bootstrap_nodes) = &config.pkdns.dht_bootstrap_nodes {
```
ok300, line 266: This changes the logic. Now setting `dht_bootstrap_nodes` doesn't disable the default pkarr relays.

Before, setting `dht_bootstrap_nodes` would have meant DHT-only, now it leads to a custom-DHT + public-relays combo.

The comment in the previous implementation seemed reasonable:

```rust
// If we set custom bootstrap nodes, we don't want to use the default pkarr relay nodes.
// Otherwise, we could end up with a DHT with testnet boostrap nodes and mainnet relays
// which would give very weird results.
builder.no_relays();
```

This also contradicts how `config.sample.toml` describes `dht_relay_nodes`:

```toml
# If not set and no bootstrap nodes are set, the default pkarr relay nodes will be used.
dht_relay_nodes = ["https://pkarr.pubky.app", "https://pkarr.pubky.org"]
```

Judging by the new `public_mode_keeps_defaults_and_applies_config` this seems intentional, but doesn't the older behavior make more sense?

pubky-homeserver/src/data_directory/persistent_data_dir.rs
```diff
@@ -81,44 +86,15 @@ impl PersistentDataDir {
     /// Creates the directory, writes a sample config file (if absent),
     /// and generates a server keypair (if absent).
     pub fn init(&self) -> anyhow::Result<()> {
-        self.ensure_data_dir_exists_and_is_writable()?;
+        self.ensure_exists_and_is_writable()?;
         self.read_or_create_config_file()?;
         self.read_or_create_keypair()?;
         Ok(())
```
ok300:
```suggestion
        self.bootstrap().map(|_| ())
```

This could use the new `bootstrap` method.

pubky-homeserver/src/persistence/sql/database_mode.rs
```diff
@@ -47,27 +59,19 @@ impl DatabaseMode {
...
-    /// Tests always get ephemeral databases — the decision is encoded here,
-    /// not in a URL query parameter.
     pub fn resolve_test(explicit: Option<ConnectionString>) -> anyhow::Result<Self> {
-        let env_val = std::env::var("TEST_PUBKY_CONNECTION_STRING").ok();
-        Self::resolve_test_inner(explicit, env_val)
+        Self::resolve_test_inner(explicit, ConnectionString::from_test_env()?)
     }
 
     /// Pure resolution logic, separated from env access for testability.
     fn resolve_test_inner(
```
ok300, line 70: This can't error out anymore, so the return can be `Self` instead of `anyhow::Result<Self>`.

This would also simplify some callers.

pubky-testnet/src/ephemeral_testnet.rs
ok300: This block can be removed and `mut config` doesn't have to be `mut`.

The new `create_homeserver_with` called below already takes care of that:

https://github.com/pubky/pubky-homeserver/blob/44e8b44eae337ba7e7ecad784688f52d10c36a62/pubky-testnet/src/testnet.rs#L99-L107

(A)

pubky-testnet/src/ephemeral_testnet.rs
```diff
@@ -335,8 +334,9 @@ impl EphemeralTestnet {
             .clone()
             .or(config.general.database_url);
```
ok300: Same as (A)

pubky-testnet/src/static_testnet.rs
```diff
@@ -759,4 +694,45 @@ mod tests {
...
+            Some(programmatic.clone()),
+            Some(from_config),
+            Some(from_env),
+        )
+        .unwrap();
+        assert_eq!(result.unwrap().as_str(), programmatic.as_str());
+    }
+
+    #[test]
+    fn persistent_db_url_config_used_when_no_programmatic() {
```
ok300: This test supplies both config and env values, then asserts the config value wins.

However the `README` says the environment variable “overrides the `database_url` in the on-disk config”:

https://github.com/pubky/pubky-homeserver/blob/44e8b44eae337ba7e7ecad784688f52d10c36a62/pubky-testnet/README.md?plain=1#L28

This was also mentioned earlier in https://github.com/pubky/pubky-homeserver/pull/500#discussion_r3844908580 , but not sure if the intention was to address it in another follow-up.

pubky-homeserver/src/app_context.rs
86667: This is intentional. The old behavior coupled two independent config vars. `dht_bootstrap_nodes` would implicitly disable all default relays as a hidden side effect. Now the mainline network config is determined by the entry point:

- `AppContext::new_ephemeral` / `isolated_pkarr_builder` starts from a blank network (no defaults) — used by testnets
- `AppContext::from_persistent_dir` starts from mainnet defaults — used in production

If someone wants no default relays, they should use the isolated builder or explicitly configure `dht_relay_nodes`. Bootstrap nodes and relay nodes are independent concerns and shouldn't be implicitly coupled imo.

Updated the `config.sample.toml` docs, good catch.

pubky-homeserver/src/app_context.rs
86667: Explaining this required a few lines which signaled that perhaps the abstraction is not clear enough. So i removed it, in favour of `AppContext` builders configuring their own `PkarrClient` based on their needs. Keeping things flexible.

pubky-testnet/src/static_testnet.rs
86667: Woops - fixed to prioritise env var.

pubky-homeserver/src/app_context.rs
ok300: I think there's still an issue with `dht_relay_nodes = [ ]` , it would result in an error which would prevent startup.

A few lines below, `pubky-homeserver/src/app_context.rs:274` passes `dht_relay_nodes = []` to `pkarr::ClientBuilder::relays`. With pkarr 8.0.0, that later fails `build()` with `EmptyListOfRelays`; it does not disable relays. This contradicts the new guidance in [`pubky-homeserver/config.sample.toml:155`](https://github.com/pubky/pubky-homeserver/blob/c8b0c50f91117ed5433c0465b9d944ac4320849c/pubky-homeserver/config.sample.toml#L155-L156) telling operators to clear relays with an empty list.

For example, this test would fail on the last assertion:

```rust
#[test]
fn empty_relay_config_disables_relays_and_builds() {
    let mut config = ConfigToml::default_test_config();
    config.pkdns.dht_relay_nodes = Some(vec![]);

    let mut builder = pkarr::ClientBuilder::default();
    AppContext::apply_config_to_pkarr(&mut builder, &config);

    for relay in pkarr::DEFAULT_RELAYS {
        assert!(
            !format!("{builder:?}").contains(relay),
            "explicitly empty relay config must not retain default relay {relay}"
        );
    }

    builder
        .build()
        .expect("an explicitly empty relay list should disable relays, not reject startup");
}
```

One solution would be to adjust `AppContext::apply_config_to_pkarr` :

```diff
    fn apply_config_to_pkarr(builder: &mut pkarr::ClientBuilder, config: &ConfigToml) {
        if let Some(bootstrap_nodes) = &config.pkdns.dht_bootstrap_nodes {
            let nodes = bootstrap_nodes
                .iter()
                .map(|node| node.to_string())
                .collect::<Vec<String>>();
            builder.bootstrap(&nodes);
        }

-       if let Some(relays) = &config.pkdns.dht_relay_nodes {
-           builder
-              .relays(relays)
-               .expect("parameters are already urls and therefore valid.");
-       }
+       if let Some(relays) = &config.pkdns.dht_relay_nodes {
+           if relays.is_empty() {
+               builder.no_relays();
+           } else {
+               builder
+                   .relays(relays)
+                   .expect("parameters are already URLs and therefore valid.");
+           }
+       }

        if let Some(request_timeout) = &config.pkdns.dht_request_timeout_ms {
            let duration = Duration::from_millis(request_timeout.get());
            builder.request_timeout(duration);
        }
    }
}
```

pubky-homeserver/src/persistence/sql/database_mode.rs
```diff
@@ -47,28 +59,23 @@ impl DatabaseMode {
     /// 1. Explicitly provided URL (e.g. from Docker Postgres or config) → `EphemeralTest`
     /// 2. `TEST_PUBKY_CONNECTION_STRING` environment variable → `EphemeralTest`
     /// 3. [`DEFAULT_TEST_SERVER`] fallback → `EphemeralTest`
-    ///
-    /// Tests always get ephemeral databases — the decision is encoded here,
-    /// not in a URL query parameter.
     pub fn resolve_test(explicit: Option<ConnectionString>) -> anyhow::Result<Self> {
```
ok300, line 62: This method (`pubky-homeserver/src/persistence/sql/database_mode.rs:62`) as well as  `resolve_persistent_database_url`  ([`pubky-testnet/src/static_testnet.rs:418`](https://github.com/pubky/pubky-homeserver/blob/c8b0c50f91117ed5433c0465b9d944ac4320849c/pubky-testnet/src/static_testnet.rs#L418)) parse `TEST_PUBKY_CONNECTION_STRING` before resolving precedence. This means a malformed environment value makes startup fail even when a valid explicit/programmatic URL should win. The previous `DatabaseMode` flow ignored the env value when an explicit URL existed.

One solution is to resolve the higher-priority source first, and only parse the environment when it can actually be selected.

Some more context from AI below (decided to include it all, as it seems relevant).

---

### Failure example

A caller explicitly supplies a valid test database URL, but their shell/CI has a stale invalid environment variable:

```bash
export TEST_PUBKY_CONNECTION_STRING='not-a-postgres-url'
```

```rust
let explicit = ConnectionString::new(
    "postgres://postgres:***@localhost:5432/postgres",
)
.unwrap();

let mode = DatabaseMode::resolve_test(Some(explicit));
```

Expected: `explicit` wins, so `mode` uses the valid PostgreSQL URL.

Current flow in `pubky-homeserver/src/persistence/sql/database_mode.rs:62`:

1. `resolve_test(Some(explicit))` calls `ConnectionString::from_test_env()?`.
2. The invalid env value fails parsing.
3. The function returns an `Invalid TEST_PUBKY_CONNECTION_STRING` error.
4. It never reaches the later `explicit.or(from_env)` priority logic.

A practical caller affected is:

```rust
EphemeralTestnet::builder()
    .postgres(valid_connection_string)
    .build()
    .await
```

`Testnet::create_homeserver_with` promotes that programmatic URL into the config, then `AppContext::new_ephemeral` calls `DatabaseMode::resolve_test`. An unrelated malformed environment variable prevents the explicit `.postgres(...)` value from being used.

`StaticTestnet` has the same ordering flaw in `pubky-testnet/src/static_testnet.rs:418`: a valid programmatic override is masked by an invalid lower-priority env value.

### Recommended solution

Select the higher-priority typed value before attempting to parse lower-priority sources.

For `DatabaseMode`:

```rust
pub fn resolve_test(explicit: Option<ConnectionString>) -> anyhow::Result<Self> {
    let url = match explicit {
        Some(url) => url,
        None => ConnectionString::from_test_env()?.unwrap_or_else(|| {
            ConnectionString::new(DEFAULT_TEST_SERVER)
                .expect("Default test connection string is valid")
        }),
    };

    Ok(Self::EphemeralTest(url))
}
```

For the persistent `StaticTestnet` path:

```rust
fn resolve_persistent_database_url(
    programmatic: Option<ConnectionString>,
    from_config: Option<ConnectionString>,
) -> anyhow::Result<Option<ConnectionString>> {
    match programmatic {
        Some(url) => Ok(Some(url)),
        None => Ok(ConnectionString::from_test_env()?.or(from_config)),
    }
}
```

That yields the intended policy:

| Source state | Result |
|---|---|
| Valid explicit/programmatic URL + invalid env | Use explicit/programmatic URL |
| No explicit URL + valid env | Use env |
| No explicit URL + invalid env | Return a useful env configuration error |
| Neither explicit nor env | Fall back to config/default as applicable |

I would keep the two small branches rather than introduce a shared abstraction: the fallback behavior differs between ephemeral test DBs and persistent testnets.

### Regression test

For reliable unit coverage without mutating process-wide environment variables, make the resolver accept a lazy env reader internally:

```rust
fn resolve_test_inner(
    explicit: Option<ConnectionString>,
    read_env: impl FnOnce() -> anyhow::Result<Option<ConnectionString>>,
) -> anyhow::Result<DatabaseMode> {
    let url = match explicit {
        Some(url) => url,
        None => read_env()?.unwrap_or_else(|| {
            ConnectionString::new(DEFAULT_TEST_SERVER)
                .expect("Default test connection string is valid")
        }),
    };

    Ok(DatabaseMode::EphemeralTest(url))
}
```

Then the regression test can prove the bad env source is never evaluated:

```rust
#[test]
fn explicit_database_url_does_not_evaluate_invalid_env() {
    let explicit = ConnectionString::new("postgres://custom:5432/mydb").unwrap();

    let mode = DatabaseMode::resolve_test_inner(Some(explicit.clone()), || {
        anyhow::bail!("invalid TEST_PUBKY_CONNECTION_STRING")
    })
    .unwrap();

    assert_eq!(mode.connection_string(), &explicit);
}
```

This is a pure unit test: no database, network, or environment mutation is needed. I have not run it.

pubky-homeserver/README.md
```diff
@@ -64,19 +64,22 @@ async fn main() -> anyhow::Result<()> {
 }
 ```
 
-For testing, use `MockDataDir` to create a temporary directory that is cleaned up on drop. Enable the `testing` feature:
+For testing, use `AppContext::new_ephemeral` to create a context backed by an auto-cleaning temporary directory. Enable the `testing` feature:
 
 ```toml
 [dev-dependencies]
 pubky-homeserver = { version = "0.x", features = ["testing"] }
```
ok300, line 71:
```suggestion
pubky-homeserver = { version = "0.x", features = ["testing"] }
pubky-common = "0.x"
```

The snippet below uses a `pubky_common` import, but in the dependencies snippet, only `pubky-homeserver` is declared. So for someone trying out the code below, it would fail to compile.

Simplest fix would be to add the `pubky-common` dependency.

PR #577, opened 2026-09-01
feat(storage)!: make homeserver storage publication crash-safe

ben-kaufman:
## Summary

Makes homeserver file publication crash-safe by storing content as immutable backend blobs and atomically publishing the logical entry, quota accounting, and events in PostgreSQL.

This provides the durable storage foundation for conditional writes in #578.

## What changed

- Writes immutable blobs before atomically switching logical file entries to them.
- Tracks pending uploads, retained versions, and garbage in one `unreferenced_blobs` table.
- Gives uploads one hour to finish and publish, followed by a five-minute cleanup grace. Expired uploads cannot publish; known failures retain settling time for late backend completion.
- Uses PostgreSQL row locks for cleanup ownership, with bounded batches, five-second backend-delete timeouts, and a 45-second pass budget.
- Runs queued cleanup every 15 minutes, with a separate startup/daily scan for untracked blobs.
- Retains replaced/deleted versions for one hour, but may reclaim them sooner under physical-quota pressure. Current files and active uploads remain protected; reads or copies of reclaimed versions may need retrying.
- Uses the server public key as the immutable-blob namespace.
- Routes user and admin writes through the same file service. Admin PUT streams complete files directly; read ranges, non-overwriting COPY/MOVE, and legacy collision/repair behavior remain supported.
- Consolidates write wrappers and reuses listing/move helpers. Updates storage guides, API documentation, and failure-path tests.

## Deployment and upgrade

Adds an optional blob reference to existing entries and one table for unreferenced blobs. Existing file data is not rewritten during migration.

Instances using the same server key and storage backend must share one authoritative database. A separate or restored database must not use that same key against live storage: its cleanup could delete files belonging to the other database.

Migrations run at startup. Stop all instances and back up PostgreSQL, blob storage, and the server key before upgrading. Mixed-version rolling upgrades are not supported; rollback after the first new-format write requires matching pre-upgrade database and storage backups. See [Storage Operations](https://github.com/pubky/pubky-homeserver/blob/codex/crash-safe-storage-pr/docs/STORAGE.md).

Admin WebDAV supports whole-file PUT, not partial/append writes or PATCH. It does not provide LOCK/UNLOCK, persisted empty collections, conditional mutations, or full WebDAV compliance. See the [admin API](https://github.com/pubky/pubky-homeserver/blob/codex/crash-safe-storage-pr/pubky-homeserver/openapi-admin.yml).

Cleanup holds a database connection during bounded backend I/O. Uploads taking over an hour must retry. These operational trade-offs are documented in the storage guide.

## Verification

At `5a1c48b2`, and on the combined stack at `f89c5308`:

- `cargo fmt --all --check` and `git diff --check`: passed.
- Standalone `cargo test -p pubky-homeserver --all-features --locked --offline --lib`: 558 passed.
- Combined stack: 568 homeserver, 218 SDK, 97 E2E, 3 host WASM, 1 additional native integration test, and 63 SDK doc tests passed; four existing doc examples ignored.
- Native workspace Clippy (all targets/features, excluding WASM) and WASM-target Clippy: passed with warnings denied.
- Browser WASM tests: 2 passed.
- SDK docs: no warnings. Homeserver docs build with two pre-existing private-link warnings in unchanged `database_mode.rs`.
- Independent final review of the cleanup deltas found no actionable correctness or code-quality issues.
- Windows and real GCS execution were not tested.

## Review order

This is the first PR in the storage stack. #578 adds conditional storage writes on top.


pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
```diff
@@ -0,0 +1,1008 @@
...
+        executor: &mut UnifiedExecutor<'a>,
+    ) -> Result<u64, sqlx::Error> {
+        if blob_keys.is_empty() {
+            return Ok(0);
+        }
+        let con = executor.get_con().await?;
+        let result = sqlx::query(
+            r#"
+            INSERT INTO blob_garbage (blob_key, user_id, content_length, available_at)
+            SELECT candidate.blob_key, NULL, 0, statement_timestamp()
```
MCarlomagno, line 233: This treats every blob missing from this database as garbage. 

If two homeservers share one GCS bucket, each can delete the others live blobs, I think we should namespace blobs by homeserver or require some kind of "ownership marker" before queuing

pubky-homeserver/src/persistence/sql/migrations/m20260827_add_immutable_blob_storage.rs
```diff
@@ -0,0 +1,212 @@
...
+
+use crate::persistence::sql::migration::MigrationTrait;
+
+/// Adds immutable blob pointers and durable upload/garbage tracking.
+pub struct M20260827AddImmutableBlobStorageMigration;
+
+#[async_trait]
+impl MigrationTrait for M20260827AddImmutableBlobStorageMigration {
+    async fn up(&self, tx: &mut Transaction<'static, sqlx::Postgres>) -> anyhow::Result<()> {
+        sqlx::query("ALTER TABLE entries ADD COLUMN IF NOT EXISTS blob_key TEXT")
```
MCarlomagno, line 12: Lets add an index on `entries(blob_key)`, it doesnt harm and guarantees better performance

pubky-homeserver/src/admin_server/dav_file_system.rs
```diff
@@ -0,0 +1,912 @@
...
+                } => {
+                    let end = position
+                        .saturating_add(count as u64)
+                        .min(entry.content_length);
+                    if end <= *position {
+                        return Ok(Bytes::new());
+                    }
+                    let bytes = self
+                        .file_service
+                        .get_entry_range(entry, lease, *position..end)
```
MCarlomagno: Each DAV read fetches one range from the backend

With 16 KiB reads, a 1 GiB GCS download can make about 65k requests, I would keep a streaming reader or cache larger ranges (per open file)

pubky-homeserver/src/admin_server/app_state.rs
```diff
@@ -14,10 +13,33 @@ pub(crate) struct AppState {
...
+            DavMethod::Put,
+            DavMethod::Patch,
+            DavMethod::Options,
+            DavMethod::PropFind,
+            DavMethod::Copy,
+            DavMethod::Move,
+            DavMethod::Delete,
+        ] {
+            methods.add(method);
+        }
```
MCarlomagno, line 39: This removes `LOCK`, `UNLOCK`, and `MKCOL` from the previous DAV, if this is intentional, lets evaluate the implications for consumers, if it's not, lets add them back

pubky-homeserver/src/admin_server/routes/dav_handler.rs
```diff
@@ -22,11 +27,184 @@ pub async fn dav_handler(
...
+            .body(Body::from("User collections cannot be deleted"))
+            .expect("This response should always be valid"));
+    }
+    if has_unsupported_mutation_precondition(&req) {
+        return Ok(Response::builder()
+            .status(StatusCode::NOT_IMPLEMENTED)
+            .body(Body::from("Conditional WebDAV mutations are not supported"))
+            .expect("This response should always be valid"));
+    }
+    if req.method().as_str() == "MKCOL" {
```
MCarlomagno: i would use `DavMethod::MkCol` instead of raw strings

pubky-homeserver/src/admin_server/routes/dav_handler.rs
```diff
@@ -22,11 +27,184 @@ pub async fn dav_handler(
...
+        return Ok(Response::builder()
+            .status(StatusCode::NOT_IMPLEMENTED)
+            .body(Body::from("Empty collections are not supported"))
+            .expect("This response should always be valid"));
+    }
+    if matches!(req.method().as_str(), "COPY" | "MOVE") {
```
MCarlomagno: similar here, I would rather do something like:

`if matches!(method, Some(DavMethod::Copy | DavMethod::Move))`

pubky-homeserver/src/admin_server/routes/dav_handler.rs
```diff
@@ -22,11 +27,184 @@ pub async fn dav_handler(
...
+        Ok(_) => Ok(None),
+        Err(crate::persistence::files::FileIoError::NotFound) => Ok(Some(StatusCode::NOT_FOUND)),
+        Err(error) => Err(error),
+    }
+}
+
+fn has_unsupported_mutation_precondition(request: &Request<Body>) -> bool {
+    let is_mutation = matches!(
+        request.method().as_str(),
+        "PUT" | "PATCH" | "DELETE" | "COPY" | "MOVE" | "PROPPATCH"
```
MCarlomagno: I see many cases like this, I would try to avoid avoid repeating raw WebDAV method strings, and use the enum instead

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
ben-kaufman: Agreed, this could delete another homeserver's blobs. Fixed in [76ddf51](https://github.com/pubky/pubky-homeserver/commit/76ddf5130d4feadd305143de780c6b8f93c0582b): blob creation and reconciliation now use a namespace persisted in PostgreSQL. Instances sharing that database share the namespace; independent databases get separate ones. Added a regression covering two databases sharing the same storage backend.

pubky-homeserver/src/persistence/sql/migrations/m20260827_add_immutable_blob_storage.rs
ben-kaufman: Added the index on `entries(blob_key)` and a migration assertion checking it in [76ddf51](https://github.com/pubky/pubky-homeserver/commit/76ddf5130d4feadd305143de780c6b8f93c0582b).

pubky-homeserver/src/admin_server/dav_file_system.rs
ben-kaufman: Agreed. Added a bounded 1 MiB read-ahead buffer per open file in [b467247](https://github.com/pubky/pubky-homeserver/commit/b467247d6fb168a763ac6d790fe932c9b82a85f1), so small sequential reads reuse fetched data. Seeking invalidates the buffer, and read-lease checks remain in place. Added coverage for backend read counts and seeks.

pubky-homeserver/src/admin_server/app_state.rs
ben-kaufman: This was intentional, but you're right that it affects existing clients. The previous LOCK/UNLOCK implementation used `FakeLs`, which provided client compatibility rather than actual locking. MKCOL was removed because the database-backed view doesn't persist empty directories. I haven't restored these yet. Do we need to preserve writable desktop DAV mounts and empty-directory creation?

pubky-homeserver/src/admin_server/routes/dav_handler.rs
ben-kaufman: Updated to `DavMethod::MkCol` in [b467247](https://github.com/pubky/pubky-homeserver/commit/b467247d6fb168a763ac6d790fe932c9b82a85f1).

pubky-homeserver/src/admin_server/routes/dav_handler.rs
ben-kaufman: Updated COPY/MOVE checks to match on `DavMethod` in [b467247](https://github.com/pubky/pubky-homeserver/commit/b467247d6fb168a763ac6d790fe932c9b82a85f1).

pubky-homeserver/src/admin_server/routes/dav_handler.rs
ben-kaufman: Updated the remaining checks too in [b467247](https://github.com/pubky/pubky-homeserver/commit/b467247d6fb168a763ac6d790fe932c9b82a85f1). The handler converts the request method once and uses `DavMethod` for the WebDAV-specific checks.

pubky-homeserver/src/admin_server/dav_file_system.rs
```diff
@@ -0,0 +1,987 @@
...
+            });
+            let entry = if self.create_new {
+                self.file_service
+                    .admin_create_stream_with_size_hint(&self.entry_path, stream, content_length)
+                    .await
+            } else {
+                self.file_service
+                    .admin_write_stream_with_size_hint(&self.entry_path, stream, content_length)
+                    .await
+            }
```
MCarlomagno, line 695: A chunked `PUT` can commit here before dav server rejects its body length. We should assert a matching `Content-Length` or validate before publication.

MCarlomagno (commented):
Generally looks ok to me, but I would get a few more eyes in this PR just in case, also consider adding the breaking change marker (!) and migration docs (if needed)

`feat(<scope>)!: make homeserver storage publication crash-safe`

pubky-homeserver/src/admin_server/dav_file_system.rs
ben-kaufman: Good catch. Fixed in [6485d7e](https://github.com/pubky/pubky-homeserver/commit/6485d7ec1168583f4168e2c783c247d67cfbbed1): upload lengths are checked in the body stream before DAV can flush, including `X-Expected-Entity-Length` and `Content-Range`. Invalid uploads cannot create or overwrite a file; valid chunked uploads remain supported. Added regressions for short/excess PUT and PATCH bodies, interrupted uploads, and range-length overflow. All 51 admin/DAV tests pass.

pubky-homeserver/src/services/file_service/cleanup_task.rs
```diff
@@ -0,0 +1,58 @@
...
+                interval.tick().await;
+                if let Err(error) = cleanup_service.recover_blob_storage().await {
+                    tracing::error!(%error, "Failed to recover blob storage");
+                }
+            }
+        });
+        let reconciliation_handle = tokio::spawn(async move {
+            let mut interval = tokio::time::interval(RECONCILIATION_INTERVAL);
+            interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            loop {
```
86667, line 33: This seems like a lot of work for cleaning up after a failure mode which i believe is unknown?

It certainly can run less frequently than hourly and im even wondering if it could be an admin command or only on startup. @MCarlomagno wdyt?

pubky-homeserver/src/persistence/files/file/file_service.rs
```diff
@@ -70,161 +356,1135 @@ impl FileService {
...
+        stream: impl Stream<Item = Result<Bytes, WriteStreamError>> + Unpin + Send,
+        size_hint: u64,
+    ) -> Result<EntryEntity, FileIoError> {
+        self.write_stream_inner(path, stream, WriteMode::AdminCreate, Some(size_hint))
+            .await
+    }
 
-        Ok(Bytes::from(collected_data))
+    pub(crate) async fn admin_users(&self) -> Result<Vec<String>, FileIoError> {
+        Ok(UserRepository::get_all(&mut self.db.pool().into())
```
86667: use user_service

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
```diff
@@ -0,0 +1,1017 @@
+use crate::persistence::sql::UnifiedExecutor;
+
+/// Persists immutable blob upload and cleanup bookkeeping.
+pub struct BlobRepository;
```
86667, line 5: This file doesnt follow the existing entities pattern - there should be a domain objects eg `BlobGarbageEntity` which manages logic and validation around the db type. See eg `EventEntry` or `UserEntry`

pubky-homeserver/src/persistence/files/file/file_service.rs
```diff
@@ -4,54 +4,340 @@ use crate::{
...
+const READ_LEASE_SECONDS: i64 = 2 * 60;
+const READ_LEASE_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(30);
+const READ_LEASE_DATABASE_TIMEOUT: Duration = Duration::from_secs(10);
+const READ_LEASE_CACHE_CAPACITY: usize = 4096;
+// Bound physical data to the active version, one retained version, and one replacement upload.
+const PHYSICAL_STORAGE_QUOTA_MULTIPLIER: u64 = 3;
+
+/// Coordinates logical file entries in PostgreSQL with immutable backend blobs.
 #[derive(Debug, Clone)]
 pub struct FileService {
```
86667: This `FileService` was a simple wrapper coupling SqlDb and the OpenDAL service. Now its turned into a large service managing OpenDAL file operations, admin operations and manages blob uploads, garbage collections, read leases etc. 

It should live in `/services` and probably be split into abstractions with lesser concerns, or at the least the file should be split up into its groupings of functionality. Seems like the read-lease subsystem (ActiveBlobReadLease, LeaseProtectedStream, LRU cache), upload heartbeats, the write flow (reserve/commit/abandon), and admin operations are each self-contained enough to be their own modules.

wdyt?

86667 (changes requested):
The whole read lease concept worries me. Its a lot of logic which adds a maintenance burden and is not simple to understand. Performance-wise it adds multiple database writes to each read and continues to each 30s of read time.

As far as i can tell all it really achieves is to protect a read which gets deleted or overwritten whilst the read is still happening. Why not simply fail the read after a timeout of say an hour (by raising `ACTIVE_BLOB_RETENTION_SECONDS`)? This is an edge case and the data is overwritten anyway so is arguably the incorrect blob to serve.

I may be missing something here. but i'd really rather not merge this level of complexity unless absolutely necessary.

pubky-homeserver/src/services/file_service/cleanup_task.rs
ben-kaufman: Agreed, changed the full backend scan to startup and daily. The lightweight cleanup queue still runs every minute to recover interrupted uploads and retry pending deletions.

pubky-homeserver/src/persistence/files/file/file_service.rs
ben-kaufman: Done, now uses the existing `UserService::get_all()`.

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
ben-kaufman: Updated the garbage and read-lease queries to return typed entities using `FromRow`, matching the existing pattern. Kept the lifecycle checks in SQL so the atomicity guarantees remain unchanged.

pubky-homeserver/src/persistence/files/file/file_service.rs
ben-kaufman: Done, moved it into `services/file_service` and split reads, writes, admin operations, and cleanup into separate modules. Kept the same behavior and interface without adding another abstraction layer.

pubky-homeserver/src/services/file_service/cleanup_task.rs
86667: Looking at the lightweight cleanup task:
```
    /// Recover orphaned uploads and retry deferred backend deletion.
    pub(crate) async fn recover_blob_storage(&self) -> Result<(), FileIoError> {
        BlobRepository::enqueue_stale_uploads(
            ABANDONED_UPLOAD_AGE_SECONDS,
            STALE_UPLOAD_RECOVERY_GRACE_SECONDS,
            &mut self.db.pool().into(),
        )
        .await?;
        BlobRepository::prune_expired_read_leases(&mut self.db.pool().into()).await?;
        self.drain_blob_garbage().await;
        Ok(())
    }
```

the only one which arguably should be ran frequently is `drain_blob_garbage` in order to avoid large storage costs. Though we rely on rate limiting for keeping user's activity bounded. 

How about running this every 15 mins, or even hourly?

ben-kaufman: Changed the cleanup task to every 15 minutes in e2e94c8. The full backend scan remains at startup and daily.

Also documented the tradeoff: old blobs from normal overwrites count toward the physical-storage limit until deleted, so frequent large overwrites may temporarily block uploads while waiting for cleanup.

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
86667: Thanks, better. Since its a larger repository could you split the file into a `/blob` dir similar to how `/entry` is structured please

ben-kaufman: Done in e630cec. Split it into `blob/entity.rs`, `blob/repository.rs`, and `blob/mod.rs`, matching `entry/`. Kept the queries and tests unchanged, with the same public imports. All 565 homeserver tests and Clippy pass.

docs/INSTALL.md
```diff
@@ -218,6 +218,15 @@ sed -i 's|^# \[general\]|[general]|; s|^# database_url = .*|database_url = "post
 
 ## Run
 
+Database migrations run automatically at startup. Stop all homeserver instances
```
86667: migration docs shoudlnt be dumped in install instructions

docs/INSTALL.md
```diff
@@ -326,10 +335,30 @@ The generated `config.toml` works out of the box for local use. Here are a few s
 | `general.database_url` | PostgreSQL connection string. | `postgres://localhost:5432/pubky_homeserver` |
 | `general.signup_mode` | `"open"` or `"token_required"`. | `"token_required"` |
 | `storage.type` | Storage backend: `file_system`, `google_bucket`, or `in_memory`. | `file_system` |
+| `storage.admin_dav_spool_limit_mb` | Combined local disk limit for in-progress admin DAV uploads. | `1024` |
 | `admin.admin_password` | Password for the admin API. | `"admin"` |
 
 The full list of options is documented in [`pubky-homeserver/config.sample.toml`](../pubky-homeserver/config.sample.toml).
 
+When using `google_bucket`, disable bucket soft delete and Object Versioning so
```
86667: this is a huge amount of info which just isnt relevant for a user-friendly install doc.  I see that we dont explicitly support google bucket in any of our docs. Maybe lets move this to a googel bucket specific doc for now?

And please make it more user friendly. this is not easy to read, eg "incomplete multipart uploads are
not visible to the homeserver's completed-object reconciliation"

pubky-homeserver/src/services/file_service/writes.rs
```diff
@@ -0,0 +1,438 @@
...
+};
+use bytes::Bytes;
+use futures_util::Stream;
+
+use super::{upload_heartbeat::UploadHeartbeat, FileService};
+
+// A failed remote close can complete after the client loses the response.
+const ABANDONED_UPLOAD_SETTLE_SECONDS: i64 = 60 * 60;
+const ACTIVE_BLOB_RETENTION_SECONDS: i64 = 60 * 60;
+// Bound physical data to the active version, one retained version, and one replacement upload.
```
86667, line 23: Retention pins each replaced blob for an hour and cleanup sweeps every 15 minutes, so garbage isn't reclaimable for 60-75 minutes, which i think is fine. 

Did you consider a force-cleanup in the circumstance where a user is hitting quota exceeded even though they have space but because they have blobs which are awaiting GC?

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
```diff
@@ -279,20 +409,26 @@ impl EntryRepository {
...
+                    ELSE entries.path \
+                 END AS regpath",
                 vec![sea_query::Value::from(dir_path.clone())],
             ))
             .left_join(
                 USER_TABLE,
                 Expr::col((ENTRY_TABLE, EntryIden::User)).eq(Expr::col((USER_TABLE, UserIden::Id))),
             )
-            .and_where(Expr::col((ENTRY_TABLE, EntryIden::Path)).like(format!("{}%", dir_path))) // Everything that starts with the path
+            .and_where(Expr::cust_with_values(
```
86667, line 428: I guess this is a fix of pre-existing poor code? Thanks, AI tells me that `list_deep` has the same pattern, could you fix there too? Also, a separate commit for these unrelated changes would be nice, else they are buried in with the large `f0794a40 "Make blob publication crash-safe"`

pubky-homeserver/src/services/file_service/writes.rs
```diff
@@ -0,0 +1,438 @@
...
+            })
+        }
+        .await;
+        match result {
+            Ok(reservation) => {
+                tx.commit().await?;
+                Ok(reservation)
+            }
+            Err(error) => {
+                tx.rollback().await?;
```
86667:
```suggestion
                if let Err(rollback_error) = tx.rollback().await {
                    tracing::error!(%rollback_error, "Failed to roll back blob publication");
                }
```

86667: like in commit_write() below

pubky-homeserver/src/services/file_service/writes.rs
```diff
@@ -0,0 +1,438 @@
...
+    shared::webdav::EntryPath,
+};
+use bytes::Bytes;
+use futures_util::Stream;
+
+use super::{upload_heartbeat::UploadHeartbeat, FileService};
+
+// A failed remote close can complete after the client loses the response.
+const ABANDONED_UPLOAD_SETTLE_SECONDS: i64 = 60 * 60;
+const ACTIVE_BLOB_RETENTION_SECONDS: i64 = 60 * 60;
```
86667: Add comment here please explaining that this now sets the limit on a read's max duration

86667: coudl this be nenamed now? `UNREFERENCED_BLOB_GRACE_SECONDS` or `INFLIGHT_READ_GRACE_SECONDS`? wdyt?

86667 (commented):
Cool, thanks.

A partial review, more to come

docs/INSTALL.md
ben-kaufman: Moved the upgrade and rollback instructions into a separate storage guide, leaving a short link in INSTALL.md. Done in [6b3ce20](https://github.com/pubky/pubky-homeserver/commit/6b3ce2019e1cee7b9deb52bad10cac04aff1050b).

docs/INSTALL.md
ben-kaufman: Moved Google Bucket configuration into its own guide and simplified the wording. General backup and cleanup information now lives in the storage guide. Done in [6b3ce20](https://github.com/pubky/pubky-homeserver/commit/6b3ce2019e1cee7b9deb52bad10cac04aff1050b).

pubky-homeserver/src/services/file_service/writes.rs
ben-kaufman: Added a bounded cleanup pass for the affected user when retained blobs block an upload, followed by one retry, in [8b02f77](https://github.com/pubky/pubky-homeserver/commit/8b02f77b37a0c897f2f8bd79cdd12285b42be0a3).

This can remove replaced/deleted blobs before the one-hour grace period ends. The tradeoff is that an ongoing download or copy using an old version may fail and need retrying. Current files and active uploads remain protected, and write crash safety is unchanged. Is that tradeoff acceptable?

pubky-homeserver/src/persistence/sql/entities/entry/repository.rs
ben-kaufman: Yes, this was pre-existing. Fixed `list_deep` too, with tests covering literal path characters such as `%` and `_`. Extracted the shallow/deep listing and directory-existence fixes into a separate commit, [6b7a186](https://github.com/pubky/pubky-homeserver/commit/6b7a18690f52a040b1e5ab53ac41a5d2ceb3c730), before the storage changes. The reordered history preserves the exact reviewed final file tree.

pubky-homeserver/src/services/file_service/writes.rs
ben-kaufman: Updated this to log rollback failures without replacing the original error, matching `commit_write`. Applied the same fix to file deletion. Done in [8b02f77](https://github.com/pubky/pubky-homeserver/commit/8b02f77b37a0c897f2f8bd79cdd12285b42be0a3).

pubky-homeserver/src/services/file_service/writes.rs
ben-kaufman: Renamed it to `UNREFERENCED_BLOB_GRACE_SECONDS` and added the explanation in [8b02f77](https://github.com/pubky/pubky-homeserver/commit/8b02f77b37a0c897f2f8bd79cdd12285b42be0a3). More precisely, this is grace after a blob is replaced/deleted, not a maximum duration for every read. Unchanged files have no read deadline. Quota-pressure cleanup can shorten the grace.

pubky-homeserver/src/admin_server/dav_file_system.rs
```diff
@@ -0,0 +1,1046 @@
...
+    shared::webdav::{EntryPath, StoragePath},
+};
+
+const DAV_READ_AHEAD_BYTES: usize = 1024 * 1024;
+
+#[derive(Clone)]
+pub(crate) struct AdminDavFileSystem {
+    file_service: FileService,
+    spool_directory: Arc<PathBuf>,
+    spool_budget: Arc<DavSpoolBudget>,
```
86667: From what i can tell all this spool infrastructure is to support the addition of partial writes to the admin webdav server. I dont think this is needed?

pubky-homeserver/src/persistence/sql/migrations/m20260827_add_immutable_blob_storage.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        sqlx::query(
+            "CREATE INDEX IF NOT EXISTS blob_uploads_updated_at_idx ON blob_uploads (updated_at)",
+        )
+        .execute(&mut **tx)
+        .await?;
+
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS blob_garbage (
```
86667: Could blob_upload and blob_garbage be combined into a single table with a `state` column? they both refer to a blob which exists in the backend but no entry point to it.

86667: AI suggestion (please do own investigation into the best solution):

```
CREATE TABLE unreferenced_blobs (
    blob_key        TEXT PRIMARY KEY,
    user_id         INTEGER,            -- NULL for reconciled orphans
    content_length  BIGINT NOT NULL,
    state           TEXT NOT NULL,      -- 'uploading' | 'replaced' | 'orphan'
    eligible_at     TIMESTAMP NOT NULL  -- when the drain may delete it
);
The trick is eligible_at. Today an upload has updated_at (heartbeat) and garbage has available_at (grace) — different columns with different semantics. But both reduce to "the drain may take this after time T":

heartbeat → eligible_at = now + 1h + 5min (settle included)
replace/delete → eligible_at = now + 1h, state = 'replaced'
reconciled orphan → eligible_at = now, state = 'orphan'
Then enqueue_stale_uploads disappears entirely. A dead upload is just a row whose eligible_at has passed; the drain picks it up with no separate promotion step. abandon_upload becomes a one-line UPDATE … SET state='replaced', eligible_at=now. activate_upload stays a DELETE … WHERE state='uploading'. The retained_for_reads boolean is subsumed by state = 'replaced' — that's exactly what quota-pressure cleanup keys on.

recover_blob_storage becomes one step instead of two, and the repository drops from ~15 methods to roughly 9.
```

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
```diff
@@ -0,0 +1,780 @@
...
+    pub async fn claim_garbage_for_quota<'a>(
+        user_id: i32,
+        limit: i64,
+        stale_claim_seconds: i64,
+        executor: &mut UnifiedExecutor<'a>,
+    ) -> Result<Vec<BlobGarbageEntity>, sqlx::Error> {
+        Self::claim_garbage_inner(limit, stale_claim_seconds, Some(user_id), executor).await
+    }
+
+    async fn claim_garbage_inner<'a>(
```
86667: Thoughts on this:

## Replace the claim lease with a held transaction
`claimed_at` + `claim_token` + `STALE_GARBAGE_CLAIM_SECONDS` exist to lease rows to a worker that might die. Postgres already has that primitive:

```
BEGIN;
SELECT … FOR UPDATE SKIP LOCKED LIMIT 64;   -- rows locked by this connection
-- delete from backend, 8 concurrent, 5s timeout each
DELETE … / UPDATE eligible_at …             -- finish or defer, same tx
COMMIT;
```
If the worker dies, the lock vanishes with the connection — instantly, not after a 5-minute stale-claim window. No token-mismatch reasoning, no "was my finish a no-op". Two columns and a constant gone.

The trade is a transaction held open across network deletes — normally a smell, but this is a background job with a bounded batch and a 5 s per-delete cap, so worst case is ~40 s and typically well under 1 s. You should know that trade-off is being made, but I think it's the right one.

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
86667: multiple homeservers sharing a storage backend is an interesting edge case worth considering. I think that if multiple homeserver instances are sharing a postgres instance then they are also sharing a keypair. Ie, keypair + config (including database_url) are tied. Therfore, could we drop this table and instead use the frist few bytes of the keypair?

pubky-homeserver/src/services/file_service/upload_heartbeat.rs
```diff
@@ -0,0 +1,80 @@
+use crate::persistence::{
```
86667: Similar to my comment on read leases, i feel that all of this machinery could be replaced by a generous fixed window. this becomes easier to implement with the unified blob-with-no-entry table changes from earlier in the review?

pubky-homeserver/src/services/file_service/admin.rs
```diff
@@ -0,0 +1,421 @@
...
+    ) -> Result<EntryEntity, FileIoError> {
+        self.write_stream_inner(path, stream, WriteMode::AdminOverwrite, Some(size_hint))
+            .await
+    }
+
+    pub(crate) async fn admin_create_stream_with_size_hint(
+        &self,
+        path: &EntryPath,
+        stream: impl Stream<Item = Result<Bytes, WriteStreamError>> + Unpin + Send,
+        size_hint: u64,
```
86667: No need to have 3 fns here, just expose write_stream?. Same for the non-admin side

86667 (changes requested):
A few more simplification suggestions. 

I really like this solution so am keen to find out what the minimum implementation is.  

Please look for simplification opportunities yourself too and discuss trade-offs with the core team

pubky-homeserver/src/admin_server/dav_file_system.rs
ben-kaufman: Agreed. Removed the partial/append write support, temporary-file spool, budget, and related config in f38744b. Admin PUT now streams the complete file directly through the file service. PATCH and partial-upload headers are rejected; range reads and normal COPY/MOVE/DELETE remain. Tests and the admin API docs reflect that smaller surface.

pubky-homeserver/src/persistence/sql/migrations/m20260827_add_immutable_blob_storage.rs
ben-kaufman: Implemented in f38744b: one `unreferenced_blobs` table with `uploading`, `retained`, and `garbage` states plus `eligible_at`. The stale-upload promotion step is gone. Two safety details differ from the sketch: abandonment preserves a settling delay and can recreate missing tracking, while activation checks expiry after taking the row lock. Only retained read versions can bypass the grace under quota pressure, not failed uploads or delete backoff.

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
ben-kaufman: Implemented in f38744b. Selection, backend deletion, and finish/defer now use one transaction with `FOR UPDATE SKIP LOCKED`; claim timestamps/tokens and stale-claim recovery are removed. Batches remain limited to 64 blobs, 8 concurrent deletes, 5 seconds per delete, and a 45-second total budget including DB waits. The connection-occupancy trade-off is documented. Rollback/retry and expired-upload publication tests pass.

pubky-homeserver/src/persistence/sql/entities/blob/repository.rs
ben-kaufman: Yes, adopted that deployment assumption in f38744b and removed the namespace table. The prefix uses the full server public key, avoiding shortened-prefix collisions. The docs now explicitly require one authoritative database for a given server key + storage backend. An independent/restored database must not run with that key against live storage, because its orphan cleanup could delete the other database's files. Different server keys remain isolated for new immutable blobs.

pubky-homeserver/src/services/file_service/upload_heartbeat.rs
ben-kaufman: Implemented in f38744b. The heartbeat task is gone: uploads have a fixed one-hour publication window, then five minutes of cleanup settling time. Expiry is checked under the publication transaction, so a delayed writer cannot publish after cleanup has taken ownership. Uploads exceeding the window must retry; known failed uploads retain their settling delay for late backend completion. Documented the trade-off and updated the timeout/recovery tests.

pubky-homeserver/src/services/file_service/admin.rs
ben-kaufman: Done in f38744b. Both client and admin sides expose one stream-write method with an optional size hint instead of parallel wrappers. They share the same internal publication path; create-only semantics needed for non-overwriting COPY remain internal. Removed the now-unused admin spool helpers as well.

ben-kaufman:
Thanks, I've made the simplifications we discussed:

- Uploads, retained versions and files waiting for deletion are tracked in one table.
- Removed the upload heartbeat task. Uploads have one hour to finish and publish.
- Cleanup uses database row locks, without separate claim tokens or leases.
- Removed the namespace table. Blob paths use the full server public key.
- Admin DAV uploads whole files directly, without partial-write support or temporary-file buffering.
- Shared more of the existing read/write/list/move code and removed duplicate checks and tests.
- Fixed DAV path/collision handling and expired-upload filesystem cleanup.
- In #578, authenticated listing checks that credentials belong to the target homeserver, and paths/cursors containing spaces, percent signs or Unicode work correctly.

The trade-offs to be clear about:

- The same server key and storage backend must share the same database. Two separate databases using that combination could delete each other's files during cleanup. Restores and test databases need isolated storage.
- Uploads taking over an hour must retry. The extra five minutes before cleanup does not extend that deadline. Failed uploads also get time for backend writes to finish before cleanup.
- Cleanup holds a database connection while deleting files. Each delete has a five-second timeout, and each pass is limited to 45 seconds including database waits. Interrupted cleanup can be retried.
- Admin DAV no longer supports partial/append updates or PATCH. Whole-file uploads and range reads still work.
- Old versions are kept for an hour, but can be removed earlier if storage is full. A read/copy of an old version may then need retrying. Current files and active uploads stay protected.
- Physical storage is capped at three times the user's quota, including current files, old versions and uploads in progress. Writes can be blocked if cleanup cannot free enough space, even if the user's current files are below quota.

This still gives us what we need for Paykit: file contents are stored durably before the database publishes the file entry, quota and events together. Cleanup does not delete current files when the database/storage setup follows the rule above.

Issue #506, opened 2026-07-19
feat(homeserver|sdk): conditional writes, create-only PUT, compare-and-swap.


SHAcollision:
**Premise:** the homeserver supports HTTP conditionals on reads (strong content-hash ETags, `If-None-Match` on GET). Writes have none: every PUT is an unconditional upsert, every DELETE unconditional.

**Problem:** two write patterns race today. (1) Idempotent creation: check-then-put has a window in which the losing writer silently overwrites existing data; there is no server-enforced "create only if absent". (2) Read-modify-write: between a client's GET and its PUT, another device or app can write, and the PUT silently clobbers (multi-device edits of one file; one app destroying another app's additions to a shared file).

**Phase 1 (small, implementable against the current write path): create-only PUT.** On PUT, honor `If-None-Match: *`: proceed only if no entry exists at the path, otherwise `412 Precondition Failed` with zero side effects (nothing stored, no event emitted). Requests without the header keep today's behavior exactly. `If-Match` with an ETag value is rejected as unsupported rather than ignored, so a client cannot assume compare-and-swap semantics that phase 1 does not provide. Sketch: a pre-open existence check in the same architectural position as the path-collision check, carrying the same documented residual window, backstopped by the entry table's unique (user, path) index; no atomicity claim, since the blob write and the entry-row update are separate steps today. Acceptance: 412 on existing path; create on absent path; no event on 412; no-header requests byte-identical to today; e2e on fs and memory backends.

**Phase 2 (gated on two prerequisites): full `If-Match` on PUT and DELETE.** Proceed only if the current entry's strong ETag matches, else 412. Prerequisite A: the write path finalizes blob and entry together (temp-key-then-swap, or the planned entry/events layer consolidation), because a compare that runs after the blob is flushed cannot offer 412-with-no-side-effects. Prerequisite B: capability discovery (a version or capability signal on the server), because servers that lack support silently ignore conditional headers, and a client library cannot offer safety it cannot verify. Semantic note: ETags are content hashes, so `If-Match` prevents lost updates but is subject to ABA (a value changed and changed back is indistinguishable from unchanged); it is not a linearizability claim. Create-only is unaffected by ABA.

[...]

PR #578, opened 2026-09-01
feat(storage)!: add conditional writes to the homeserver and SDK

ben-kaufman:
## Summary

Adds optimistic conditional storage writes using strong ETags and the standard HTTP `If-Match` and `If-None-Match` headers.

Clients can update or delete shared state without silently overwriting or removing a newer version written by another client. Failed preconditions return `412 Precondition Failed`, allowing the client to reload, merge, and retry.

## What changed

- Returns strong content-hash ETags for stored resources.
- Enforces `If-Match` and `If-None-Match` on storage PUT and DELETE, including compare-and-set updates, create-if-absent writes, and conditional deletion.
- Rejects already-stale uploads before reserving storage or reading the body, then rechecks preconditions atomically within the durable publication boundary.
- Advertises PUT and DELETE support through the `conditional-writes` homeserver feature.
- Adds `ResourceStats::etag`, `put_if_match`, `put_if_absent`, and `delete_if_match` to the Rust SDK.
- Adds `getBytesWithEtag`, `putBytesIfMatch`, `putBytesIfAbsent`, and `deleteIfMatch` to the JavaScript bindings.
- Exposes structured unsupported-feature and precondition-failure errors.
- Ensures authenticated storage credentials are only attached to the homeserver that issued them, including after a PKARR homeserver change.
- Uses normalized paths for authenticated directory listings and decodes URI cursors for pagination.
- Documents the conditional read-modify-write flow.
- Adds concurrent-write, stale-write, create-if-absent, conditional-delete, feature-discovery, and credential-boundary tests.

## Protocol and API impact

- New advertised feature: `conditional-writes`, guaranteeing conditional PUT and DELETE enforcement.
- Storage responses expose strong `ETag` headers.
- Conditional PUT and DELETE requests may return `412 Precondition Failed`.
- New public Rust and JavaScript SDK methods listed above.
- **Breaking Rust API change:** `RequestError::UnsupportedFeature { feature }` requires downstream exhaustive matches to handle the new variant. See the [SDK upgrade note](https://github.com/pubky/pubky-homeserver/blob/f89c530861786be3ca332339bcd78d0e16b3c408/pubky-sdk/README.md#upgrade-notes).

These are compare-and-set operations, not general transactions or WebDAV locks.

## Verification

At `f89c5308`:

- `cargo fmt --all --check` and `git diff --check`: passed.
- `cargo test -p pubky-homeserver -p pubky -p e2e -p pubky-wasm --all-features --locked --offline`: 568 homeserver, 218 SDK, 97 E2E, 3 host WASM, 1 additional native integration test, and 63 SDK doc tests passed; four existing doc examples ignored.
- `cargo clippy --workspace --all-targets --all-features --exclude pubky-wasm --locked --offline -- -D warnings`: passed.
- `cargo clippy -p pubky-wasm --target wasm32-unknown-unknown --locked --offline -- -D warnings`: passed.
- `wasm-pack test --headless --chrome --offline`: 2 browser tests passed.
- `RUSTDOCFLAGS='-D warnings' cargo doc -p pubky --all-features --no-deps --locked --offline`: passed.
- LIST credential isolation and encoded-directory regression tests failed before their fixes and pass afterward; coverage includes cookie/grant credentials, public/authenticated listings, and cursor pagination.
- Independent final review of the cleanup deltas found no actionable correctness or code-quality issues.

## Stack

This PR is based on #577 (`codex/crash-safe-storage-pr`) and should be reviewed after it.


pubky-common/src/constants.rs
```diff
@@ -18,6 +18,8 @@ pub mod storage {
 
 /// Features advertised by the homeserver client API.
 pub mod features {
+    /// Homeserver enforces `If-Match` and `If-None-Match` on storage `PUT` requests.
+    pub const CONDITIONAL_WRITES: &str = "conditional-writes";
```
MCarlomagno, line 22: `CONDITIONAL_WRITES` only has `PUT` support, but `delete_if_match` trusts it for `DELETE`. 

A server implementing this could ignore `If-Match` on `DELETE` and accept a stale deletion. I think it would make sense to cover `PUT` and `DELETE`

pubky-homeserver/src/persistence/files/file/file_service.rs
```diff
@@ -951,7 +1022,9 @@ impl FileService {
             return Err(FileIoError::UploadLeaseLost);
         }
 
-        let result = self.commit_write(path, &blob_key, &metadata, mode).await;
+        let result = self
+            .commit_write(path, &blob_key, &metadata, mode, &preconditions)
+            .await;
```
MCarlomagno: Preconditions are checked only after the body was fully uploaded. Then a stale 1 GB `If-Match` uploads 1 GB and leaves it in garbage storage for at least an hour before returning 412 error.

Please add a pre-check described in #506 and keep this check for races

pubky-sdk/src/errors.rs
```diff
@@ -147,6 +147,13 @@ pub enum RequestError {
         /// Error message from the JSON deserializer (with context if available).
         message: String,
     },
+
+    /// The target homeserver does not advertise a feature required by the operation.
+    #[error("Homeserver does not support required feature: {feature}")]
+    UnsupportedFeature {
+        /// Stable feature identifier advertised by the homeserver `/info` endpoint.
+        feature: String,
+    },
```
MCarlomagno, line 156: Adding this variant breaks potential exhaustive downstream matches on `RequestError`. 

A `0.11` (or older) client matching the previous four variants will stop compiling, if this intended lets add it to migration notes

pubky-sdk/src/actors/auth/cookie/flow.rs
MCarlomagno: This paragraph is now stale

pubky-common/src/constants.rs
ben-kaufman: Agreed. Updated the feature contract and OpenAPI docs to explicitly cover both PUT and DELETE. Both are already enforced by the server.

Updated in 8727157.

pubky-homeserver/src/persistence/files/file/file_service.rs
ben-kaufman: Added a pre-check before reserving storage or reading the upload body. The transactional check remains for races. Added regression tests for immediate rejection and for state changing during upload.

Fixed in 8727157.

pubky-sdk/src/errors.rs
ben-kaufman: Yes, this is an intentional source-breaking addition. Added an SDK upgrade note explaining that exhaustive matches must handle `RequestError::UnsupportedFeature`, without falling back to unconditional writes.

Updated in 8727157.

pubky-sdk/src/actors/auth/cookie/flow.rs
ben-kaufman: Removed the stale paragraph. The linked `await_credential()` documentation already describes binding the credential during sign-in.

Fixed in 8727157.

PR #583, opened 2026-09-02
deps: bump `jsonwebtoken` to v11, fix `JWS` tests

ok300:
This PR bumps `jsonwebtoken`, which fixes a few known vulnerabilities.

Full changelog: https://github.com/Keats/jsonwebtoken/blob/master/CHANGELOG.md

---

This bump also uncovered older issues with a few auth tests, which started failing with `InvalidEddsaKey` after the `jsonwebtoken` v11 upgrade.

These tests were using a test-only code path to generate a legacy PKCS#8 encoding accepted by the previous `ring` backend, but which is rejected by RustCrypto.

The commit which fixed the tests ( a5b712cb57b932296c848eb5b439f182ddf9171c ) switched to using the existing `pubky_common::auth::jws::sign_jws` helper already used by the SDK.

### Changes
- Switch grant, PoP and grant service fixtures to the shared JWS signer.
- Remove the obsolete private-key PEM encoder, encoding-key helper and header helper.
- Consolidate duplicate roundtrip tests while preserving signature rejection, header validation, expiry and other negative-case coverage.

Production verification logic is unchanged. Tests now exercise SDK-signed tokens through the homeserver’s existing verification paths.

### DIFF
--- a/pubky-homeserver/Cargo.toml
+++ b/pubky-homeserver/Cargo.toml
@@ -86,7 +86,7 @@ opentelemetry = "0.32"
 opentelemetry_sdk = { version = "0.32", features = ["rt-tokio", "metrics"] }
 opentelemetry-prometheus = "0.32"
 prometheus = "0.14"
-jsonwebtoken = "9"
+jsonwebtoken = { version = "11", features = ["rust_crypto"] }
 chrono = { version = "0.4", features = ["serde"] }
 sha2 = "0.10"
 rand.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -656,6 +656,12 @@ dependencies = [
 
 [[package]]
 name = "base16ct"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
+
+[[package]]
+name = "base16ct"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fd307490d624467aa6f74b0eabb77633d1f758a7b25f12bceb0b22e08d9726f6"
@@ -1239,6 +1245,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
+name = "crypto-bigint"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0dc92fb57ca44df6db8059111ab3af99a63d5d0f8375d9972e319a379c6bab76"
+dependencies = [
+ "generic-array",
+ "rand_core 0.6.4",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
 name = "crypto-common"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


MCarlomagno (dismissed):
LGTM

86667 (changes requested):
AI found:
Crypto backend is feature-selected; wrong resolution panics at request time
jsonwebtoken ≥10 requires exactly one of rust_crypto / aws_lc_rs. When zero or both are enabled, CryptoProvider::from_crate_features() silently falls through to:

```
static INSTANCE: CryptoProvider = CryptoProvider {
    verifier_factory: |_, _| panic!("{}", NOT_INSTALLED_ERROR),
    ...
};
```
pubky-homeserver is a default-member published to crates.io, and pubky-testnet re-exports it. Cargo unifies features across the graph, so any downstream crate that also depends on jsonwebtoken v11 with aws_lc_rs turns every call to verify_grant / PopProof::verify into a panic — compiles clean, fails only at runtime, in the auth hot path. Our committed Cargo.lock protects our own builds but not library consumers.

Fix in HomeserverApp::start():

```
// Pin the JWS crypto backend explicitly so a downstream feature clash
// can't swap it for the panicking fallback provider.
let _ = jsonwebtoken::crypto::rust_crypto::DEFAULT_PROVIDER.install_default();
```

It's a OnceLock::set, so idempotent and safe to ignore the Err. This is the same footgun class as rustls's provider install, and the same remedy.

86667 (commented):
Another from AI which seems like a nice idea:

`pubky-common` already owns `sign_jws`, `jws_signing_input`, and `ed25519-dalek`. A ~30-line `verify_jws` there would let the homeserver drop `jsonwebtoken` entirely, deleting its whole dependency tree and the need to fix the above issue also?

pubky-common/src/auth/jws.rs
```diff
@@ -70,7 +70,77 @@ pub fn finish_jws(signing_input: String, signature: impl AsRef<[u8]>) -> String
...
+        b64: bool,
+    }
+
+    let header_bytes = URL_SAFE_NO_PAD
+        .decode(header_b64)
+        .map_err(|_| VerifyError::InvalidFormat("invalid base64url in JWS header"))?;
+    let header: Header = serde_json::from_slice(&header_bytes)
+        .map_err(|_| VerifyError::InvalidFormat("invalid JSON in JWS header"))?;
+    if header.crit || header.b64 {
+        return Err(VerifyError::UnsupportedHeader);
```
MCarlomagno, line 129: Now valid JWS tokens with `b64: true` (ordinary `base64url` encoding) will fail verification. The old verifier accepts it, but this implementation returns `InvalidSignature`.  Is this intended?

pubky-homeserver/Cargo.toml
```diff
@@ -86,7 +86,6 @@ opentelemetry = "0.32"
 opentelemetry_sdk = { version = "0.32", features = ["rt-tokio", "metrics"] }
 opentelemetry-prometheus = "0.32"
 prometheus = "0.14"
-jsonwebtoken = "9"
```
talosmachina, line 89: **<sub><sub>P1</sub></sub> The vulnerable `jsonwebtoken` still compiles into the homeserver**

Removing the direct dependency does not remove the crate: `jsonwebtoken v9.3.1` still resolves under `opendal` (reached through both `pubky-homeserver` and `dav-server-opendalfs`, with no `--no-default-features` in `Dockerfile:41`), so the shipped binary still carries a crate that OSV flags as GHSA-h395-gr6q-cpjc / CVE-2026-25537, fixed in 10.3.0. The homeserver's own verification no longer goes through it, but the commit titled "bump `jsonwebtoken` to v11" does not deliver that fix for the artifact.

Verified by: `cargo tree -p pubky-homeserver -i jsonwebtoken` on the PR head prints `jsonwebtoken v9.3.1 -> reqsign -> opendal -> pubky-homeserver`; OSV query for 9.3.1 returns the advisory.

talosmachina (commented):
**1 finding** (1 P1). Replaces the homeserver's `jsonwebtoken` usage with an in-house `verify_jws` and switches the grant/PoP test fixtures to the shared SDK signer. Reviewed `72083db`, full tier.

<details><summary>What I checked, and 3 candidates I ruled out</summary>

**Read in full**: `pubky-common/src/auth/jws.rs`, `grant_verifier.rs`, `pop_verifier.rs`, `jws_compact.rs`, `error_mapping.rs`, `routes.rs`, `service.rs`, both `Cargo.toml`, `Cargo.lock`
**Call sites traced**: `verify_grant`, `PopProof::verify`, `sign_jws`/`JwsCompact`, `AuthServiceError` -> HTTP mapping
**CI**: green (fmt, clippy, msrv, doc, test pubky-common/homeserver/testnet, e2e, docker builds)
**Checked by running**: `cargo tree -p pubky-homeserver -i jsonwebtoken` on the PR head, and an OSV query for 9.3.1

**Ruled out**
- *Verification behaviour changed*: `verify_jws` keeps the Ed25519 check, the `typ` check and the manual `exp`/`aud` handling, and is stricter than the old path (rejects `crit` and `b64` headers, extra compact segments).
- *Grant error status moved*: a malformed header/compact now maps to 401 instead of 400 on the grant path (`error_mapping.rs:14`), but `InvalidHeaderType` -> 400 is preserved and the split looks deliberate.
- *Unverified parse reaches logic*: the signature is verified before the header and the claims are parsed.

</details>

Merge confidence: 3/5, the open P1 on the dependency that still ships.

Issue #575, opened 2026-08-31
`Pubky::restore_session()` returned errors expansion


dzdidi:
Could `Pubky::restore_session()` return more varied error types so that it is more clear what is retrieable and what is terminal?

> Pubky 0.11 exposes homeserver failures only as status + message. Even 401 can mean a recoverable PoP audience or timestamp failure, so no server status proves this stored grant is invalid. Keep these retryable until upstream preserves a typed rejection cause.

Does this make sense?

dzdidi:
Somewhat related:

Signed approval becomes externally inaccessible when Pubky SDK ACKs/deletes relay inbox message.
Current sequence:

```
Locks calls await_credential()

1.  SDK polls relay
2.  receives encrypted approval
3.  decrypts message
4.  ACKs/deletes message from relay   ← no longer retrievable
5.  decodes/verifies signed AuthToken
6.  resolves creator homeserver
7.  POST /session
8.  constructs cookie credential
```

So message from relay becomes unavailable at step 4. 
If there is an error on step 6 or 7

• relay copy is gone;
• SDK may temporarily hold token in local function memory, but returns error without exposing it;

Thus auth is irrecoverably fails.

This in combination unrliable DHT on GCP cause high and unpredictable error rates


UPDATE

looks like current auth flow does allow for retry at all because it treats them as a replay. Is it possible to accommodate retry mechanism for auth?

PR #586, opened 2026-09-07
feat(sdk): make two stage auth flow publicly available

dzdidi:
- changes visibility of `pub(crate) async fn from_auth_token` to `pub async fn from_auth_token`
- adds documentation and test

Helps addressing the issue described in https://github.com/pubky/pubky-homeserver/issues/575#issuecomment-5561879414

### DIFF
--- a/pubky-sdk/src/actors/auth/cookie/credential.rs
+++ b/pubky-sdk/src/actors/auth/cookie/credential.rs
@@ -157,8 +157,27 @@ impl CookieCredential {
         Ok(Self::new(user, cookie, record, homeserver))
     }
 
-    /// Establish a cookie credential from a signed [`AuthToken`] (legacy flow).
-    pub(crate) async fn from_auth_token(
+    /// Establish a cookie credential from a verified, signed [`AuthToken`]
+    /// (legacy flow).
+    ///
+    /// This is the second stage of split cookie authentication. Obtain the
+    /// token with [`crate::PubkyCookieAuthFlow::await_token`], retain it, then
+    /// call this method to resolve the homeserver and exchange the token at
+    /// `/session`. Because the token is borrowed, callers may retry failures
+    /// independently from relay polling while the token remains valid.
+    ///
+    /// [`AuthToken`] is short-lived and one-shot at the homeserver. A retry is
+    /// therefore not guaranteed to recover an exchange whose request reached
+    /// the homeserver but whose response was lost.
+    ///
+    /// `homeserver` binds signup credentials to the homeserver named by the
+    /// signup deep link. Pass `None` for signin, whose deep link does not name
+    /// a homeserver.
+    ///
+    /// # Errors
+    /// - Propagates homeserver resolution, HTTP transport, authentication, and
+    ///   response decoding failures.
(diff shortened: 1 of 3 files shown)


talosmachina (commented):
**No findings.** Makes the second stage of the split cookie auth flow public (`CookieCredential::from_auth_token` was `pub(crate)`), with docs and a compile-time regression test, addressing the irrecoverable-auth report from @dzdidi in #575.

One CI check is red: `Validate PR title` requires a conventional-commit prefix, so the title needs to become e.g. `feat(sdk): make two stage auth flow publicly available`. No code change implied. `wasm-test` and `docker-build` were still pending when I looked; everything else is green.

<details><summary>What I checked, and candidates I ruled out</summary>

**Read in full**: `pubky-sdk/src/actors/auth/cookie/credential.rs`, `flow.rs`, `pubky-sdk/tests/cookie_auth_stages.rs`
**Call sites traced**: `from_auth_token` is also called by `PubkyCookieAuthFlow::await_credential`, `try_poll_credential_once` and `PubkySigner::signin_cookie_with_publish` — all stay on the same behaviour, since only visibility changed.
**Verified locally**: checked out head `fd79963`; `cargo test -p pubky --test cookie_auth_stages` passes; `cargo doc -p pubky` renders `from_auth_token` in the public docs of `CookieCredential`.
**Ruled out**
- *Publicising a deprecated API's method widens a deprecated surface*: deliberate, that is the point of the PR per the linked issue.
- *Doc examples promote a deprecated flow*: both examples wrap in `#[allow(deprecated)]`; the doc check is green.
- *Retry claim in the docs is misleading*: the doc itself states the token is one-shot at the homeserver and a lost response is not recoverable, which matches the implementation.

</details>

86667 (dismissed):

pubky-sdk/src/actors/auth/cookie/credential.rs
```diff
@@ -157,8 +157,27 @@ impl CookieCredential {
...
+    /// the homeserver but whose response was lost.
+    ///
+    /// `homeserver` binds signup credentials to the homeserver named by the
+    /// signup deep link. Pass `None` for signin, whose deep link does not name
+    /// a homeserver.
+    ///
+    /// # Errors
+    /// - Propagates homeserver resolution, HTTP transport, authentication, and
+    ///   response decoding failures.
+    pub async fn from_auth_token(
```
SHAcollision, line 180: nit, not blocking: for signup the caller needs to pass the homeserver here, but `await_token` consumes the flow and `target_homeserver()` is private, so they have to keep it from the deep link themselves. Maybe a public `homeserver()` getter on the flow so it's symmetric with `await_credential`? can be a follow up, wdyt?

SHAcollision (dismissed):
LGTM

pubky-sdk/src/actors/auth/cookie/credential.rs
dzdidi: Added in https://github.com/pubky/pubky-homeserver/pull/586#event-30681324257

SHAcollision (approved):

86667 (approved):

PR #587, opened 2026-09-07
ci: use mbx for clippy and test matrix

ok300:
The recent test https://github.com/pubky/pubky-homeserver/pull/585#issuecomment-5568417815 showed promising speedups when using `mbx` cache instead of `Swatinem/rust-cache`.

This PR integrates `mbx` into the CI `clippy` and test jobs.

Partially addresses #486

## Summary

- Replace `Swatinem/rust-cache@v2` with pinned `jdx/mr-boxington-action` in the `clippy` job and six-lane test matrix.
- Run the existing Clippy, doctest, and nextest commands through mbx without changing their flags or control flow.
- Pin the action to `7234d3dd1a6ca8f6c381eea8e4dfb03f18fcf777` (`v1.3.0`) and mbx to `1.9.0`.
- Add `permissions: contents: read` at workflow scope, as extra hardening.
- Use independent cache generations:
  - `ci-mbx-v1-clippy`
  - `ci-mbx-v1-test-${{ matrix.crate }}`

## Rationale

The existing warm-cache pilot showed substantial speedups for the full Clippy workload. The test matrix is included in this PR because it is the largest remaining direct-Cargo workload in `PR Check`.

Each matrix lane receives its own cache generation so concurrent `main` jobs do not compete to save the same immutable GitHub Actions cache entry.

The integration keeps `mbx` on its default GitHub Actions backend:

- pushes to `main` seed caches;
- PRs, including fork PRs, restore only;
- no OIDC permission, secret, cache server, or manual cache-seeding workflow is added.

## Out of scope

- Other `rust-cache` jobs remain unchanged.
- Docker, `wasm-pack`/npm, `cross`, artifact-build, release, and publishing workflows remain unchanged.
- No production release path restores compiler outputs from an `mbx` cache.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches: [main]
 
+permissions:
+  contents: read
+
 jobs:
   msrv:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


andrei-21 (approved):
We should also look into speeding up `docker-build`: it took ~10 minutes in this run and is the slowest CI job, so it will still determine the overall CI time even after the mbx caches are warm. Could be a good follow-up to check Docker layer and build caching.
Maybe @talosmachina you can check this?

PR #594, opened 2026-09-07
build(docker)!: support debug builds and default to homeserver

andrei-21:
Add BUILD_PROFILE with release as the default. Use debug mode for
PR Docker checks and document usage.

BREAKING CHANGE: Docker builds now default to homeserver instead of
testnet. Pass --build-arg BUILD_TARGET=testnet to build the testnet image.

### DIFF
--- a/.github/workflows/docker-check.yml
+++ b/.github/workflows/docker-check.yml
@@ -1,0 +1,14 @@
+name: Check Release Docker Images
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    uses: ./.github/workflows/docker-build.yml
+    with:
+      publish: false
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -106,7 +106,7 @@ Requires [Docker Engine](https://docs.docker.com/engine/install/).
 Build the homeserver image using the [Dockerfile](../Dockerfile):
 
 ```bash
-docker build --build-arg BUILD_TARGET=homeserver -t pubky-homeserver .
+docker build -t pubky-homeserver .
 ```
 
 Verify the image built correctly:
@@ -114,6 +114,8 @@ Verify the image built correctly:
 ```bash
 docker run --rm pubky-homeserver homeserver --version
 ```
+
+For debug builds and testnet images, see [Docker build options](./TESTING.md#docker-build-options).
 
 ## Initialise the Data Directory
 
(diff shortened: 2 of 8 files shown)


docs/INSTALL.md
```diff
@@ -109,6 +109,13 @@ Build the homeserver image using the [Dockerfile](../Dockerfile):
 docker build --build-arg BUILD_TARGET=homeserver -t pubky-homeserver .
```
86667:
```suggestion
docker build -t pubky-homeserver .
```

docs/INSTALL.md
```diff
@@ -109,6 +109,13 @@ Build the homeserver image using the [Dockerfile](../Dockerfile):
...
 ```
 
+Builds use the release profile by default. For faster development builds, select
+the debug profile, which preserves debug symbols:
+
```
86667: I dont think this is necessary here. Im trying to keep these docs as succinct as possible by not giving more info that is absolutely necessary.

.github/workflows/docker-check.yml
```diff
@@ -0,0 +1,40 @@
...
+          registry: registry-1.docker.io/synonymsoft
+          context: .
+          push: false
+          latest: false
+          platforms: |
+            linux/arm64
+            linux/amd64
+          image: ${{ matrix.image }}
+          cache_from: type=gha,scope=docker-release-${{ matrix.target }}
+          cache_to: type=gha,scope=docker-release-${{ matrix.target }},mode=max
```
86667: AI which i dont fully understand, please chekc:

Cache scopes don't match the release workflow. docker-check.yml:37-38 uses per-target scopes, while docker.yml:46-47 uses the unscoped default. Tag builds can read caches created on the default branch, so if the scopes matched, every main push would pre-warm the release build. As written, the check builds everything and the release rebuilds from scratch. The unscoped default in docker.yml also means the testnet and homeserver matrix entries overwrite each other's cache, which is pre-existing but now easy to fix by copying the scoped form.

andrei-21: Fair. I reused one workflow for all docker builds: PR check, `main` check, release build.

docs/INSTALL.md
andrei-21: Done.

docs/INSTALL.md
andrei-21: Agree, `INSTALL.md` is not a place for such details. I just added a link to `TESTING.md` where it is described in detail.

86667 (approved):

PR #599, opened 2026-09-09
feat: add `read_timeout` to Pubky client

ok300:
This PR adds `read_timeout` to `PubkyHttpClientBuilder`.

This expands the existing `request_timeout` as follows:
- `request_timeout`: total timeout, for the entire connection, from when it starts connecting until the response body is fully received
- `read_timeout` (new): read timeout. When connection is established, this is the timeout until the response headers have to arrive. After that, chunks of the body response can arrive with at most `read_timeout` delay between them.

Similar to the existing `request_timeout`, the WASM variant is marked as deprecated.

The new config is more suitable for detecting stalled connections. This can be useful to Nexus when indexing.

### DIFF
--- a/pubky-sdk/src/client/core.rs
+++ b/pubky-sdk/src/client/core.rs
@@ -338,9 +338,9 @@ impl PubkyHttpClientBuilder {
     /// Set the maximum duration of each HTTP read operation.
     ///
     /// # Deprecated
-    /// This setter has no effect on WASM: reqwest uses the browser `fetch`
-    /// backend and does not support a global HTTP read timeout.
-    #[deprecated(note = "HTTP read timeout is not supported on WASM and has never had any effect")]
+/// This setter has no effect on WASM: reqwest's browser `fetch` backend
+/// does not support configuring a timeout for each read operation.
+#[deprecated(note = "HTTP read timeout is not supported on WASM")]
     pub fn read_timeout(&mut self, _timeout: Duration) -> &mut Self {
         self
     }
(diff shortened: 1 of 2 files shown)


pubky-sdk/src/pubky.rs
```diff
@@ -135,7 +135,7 @@ pub struct Pubky {
 }
 
 impl Pubky {
-    /// Construct with defaults (mainnet relays, standard timeouts).
+    /// Construct with the default mainnet configuration and no HTTP timeouts.
```
andrei-21: There is at least one (`/info` endpoint) and might be more timeouts in SDK. It is better to be precise here:
```suggestion
    /// Construct with the default mainnet configuration and no HTTP client timeouts.
```

pubky-sdk/src/client/core.rs
```diff
@@ -311,6 +334,16 @@ impl PubkyHttpClientBuilder {
     pub fn request_timeout(&mut self, _timeout: Duration) -> &mut Self {
         self
     }
+
+    /// Set the maximum duration of each HTTP read operation.
+    ///
+    /// # Deprecated
+    /// This setter has no effect on WASM: reqwest uses the browser `fetch`
+    /// backend and does not support a global HTTP read timeout.
+    #[deprecated(note = "HTTP read timeout is not supported on WASM and has never had any effect")]
```
andrei-21:
```suggestion
/// This setter has no effect on WASM: reqwest's browser `fetch` backend
/// does not support configuring a timeout for each read operation.
#[deprecated(note = "HTTP read timeout is not supported on WASM")]
```

andrei-21 (commented):
Nice, only minor things.

andrei-21 (approved):

PR #605, opened 2026-09-10
ci(docker): run PR builds on native amd64 and arm64 runners

andrei-21:
Add ARM64 coverage to debug Docker checks using native runners.
Allow runner selection and cache scope suffixes in the shared workflow
to isolate caches by architecture.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -3,6 +3,14 @@ name: Build Docker Images
 on:
   workflow_call:
     inputs:
+      runner:
+        description: GitHub runner label for the build
+        type: string
+        default: ubuntu-latest
+      cache_scope_suffix:
+        description: Optional suffix to isolate caches for separate architecture builds
+        type: string
+        default: ''
       build_profile:
         description: Docker build profile (debug or release)
         type: string
@@ -28,7 +36,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ inputs.runner }}
     timeout-minutes: 150
     strategy:
       matrix:
@@ -59,8 +67,8 @@ jobs:
             type=ref,event=tag
           platforms: ${{ inputs.platforms }}
           image: ${{ matrix.image }}
-          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}
-          cache_to: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }},mode=max
+          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}${{ inputs.cache_scope_suffix }}
+          cache_to: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}${{ inputs.cache_scope_suffix }},mode=max
           build_args: |
             BUILD_TARGET=${{ matrix.target }}
             BUILD_PROFILE=${{ inputs.build_profile }}
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -57,7 +57,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build Docker image
-        uses: pubky/ci-workflows/.github/actions/docker/build_and_push@e6d3b38b5ebfa97b86595cb44f3a86695e473b09 # v1.0.0
+        uses: pubky/ci-workflows/.github/actions/docker/build_and_push@5faaf004d36bb93a204b58cf7190ffe7fbcce16f
         with:
           registry: registry-1.docker.io/synonymsoft
           context: .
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-check.yml
```diff
@@ -9,6 +9,17 @@ permissions:
...
+      matrix:
+        include:
+          - architecture: amd64
+            runner: ubuntu-24.04
+          - architecture: arm64
+            runner: ubuntu-24.04-arm
     uses: ./.github/workflows/docker-build.yml
     with:
+      runner: ${{ matrix.runner }}
+      cache_scope_suffix: -${{ matrix.architecture }}
```
ok300, line 22: Preserve release-cache reuse for tag publishing

Adding `cache_scope_suffix: -${{ matrix.architecture }}` makes the release checks on `main` write architecture-specific caches. However, `.github/workflows/docker.yml` still calls the shared workflow with only `publish: true` (e.g. with no `cache_scope_suffix`), so publishing retains the empty suffix and reads the old, unsuffixed scope. For example, for the homeserver target:

```text
Release checks write:
  docker-release-homeserver-amd64
  docker-release-homeserver-arm64

Tag publishing reads:
  docker-release-homeserver
```

The same applies to `testnet`.

(A)

.github/workflows/docker-build.yml
```diff
@@ -59,8 +67,8 @@ jobs:
             type=ref,event=tag
           platforms: ${{ inputs.platforms }}
           image: ${{ matrix.image }}
-          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}
-          cache_to: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }},mode=max
+          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}${{ inputs.cache_scope_suffix }}
```
ok300: Related to (A)

ok300 (dismissed):
The above are NITs, generally LGTM.

.github/workflows/docker-check.yml
andrei-21: Fixed.

.github/workflows/docker-build.yml
andrei-21: Fixed.

.github/workflows/docker-build.yml
```diff
@@ -59,8 +67,12 @@ jobs:
             type=ref,event=tag
           platforms: ${{ inputs.platforms }}
           image: ${{ matrix.image }}
-          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}
-          cache_to: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }},mode=max
+          # Publishing imports only the caches warmed by native release checks on main.
+          cache_from: |
+            ${{ !inputs.publish && format('type=gha,scope=docker-{0}-{1}{2}', inputs.build_profile, matrix.target, inputs.cache_scope_suffix) || '' }}
```
86667: nit: bleh, bit ugly. idk if theres a nicer way though

86667 (commented):
Instead of rebuilding and relying on caching could we not use the build artifacts in merge to `main` as the artifacts for docker publishing on tag?

Edit: maybe not the way to go if that requires pushing to Docker on each main build. I'll let @ok300 approve.

.github/workflows/docker-build.yml
andrei-21: I juggled for a bit but didn't find a better way to do it.

andrei-21 (commented):
> Instead of rebuilding and relying on caching could we not use the build artifacts in merge to `main` as the artifacts for docker publishing on tag?
> 
> Edit: maybe not the way to go if that requires pushing to Docker on each main build. I'll let @ok300 approve.

Exactly — we would need to pollute the registry with useless images for that.

.github/workflows/docker-build.yml
```diff
@@ -59,8 +67,12 @@ jobs:
             type=ref,event=tag
           platforms: ${{ inputs.platforms }}
           image: ${{ matrix.image }}
-          cache_from: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }}
-          cache_to: type=gha,scope=docker-${{ inputs.build_profile }}-${{ matrix.target }},mode=max
+          # Publishing imports only the caches warmed by native release checks on main.
+          cache_from: |
```
ok300: NIT: how about smth like this?

```diff
diff --git a/.github/workflows/docker-build.yml b/.github/workflows/docker-build.yml
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -56,6 +56,31 @@
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Select cache settings
+        id: cache
+        shell: bash
+        env:
+          PUBLISH: ${{ inputs.publish }}
+          CACHE_SCOPE: docker-${{ inputs.build_profile }}-${{ matrix.target }}
+          CACHE_SCOPE_SUFFIX: ${{ inputs.cache_scope_suffix }}
+        run: |
+          if [[ "$PUBLISH" == "true" ]]; then
+            # Publishing imports native caches from main without exporting.
+            {
+              echo 'cache_from<<EOF'
+              echo "type=gha,scope=${CACHE_SCOPE}-amd64"
+              echo "type=gha,scope=${CACHE_SCOPE}-arm64"
+              echo 'EOF'
+              echo 'cache_to='
+            } >> "$GITHUB_OUTPUT"
+          else
+            cache="type=gha,scope=${CACHE_SCOPE}${CACHE_SCOPE_SUFFIX}"
+            {
+              echo "cache_from=$cache"
+              echo "cache_to=$cache,mode=max"
+            } >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Build Docker image
         uses: pubky/ci-workflows/.github/actions/docker/build_and_push@5faaf004d36bb93a204b58cf7190ffe7fbcce16f
         with:
@@ -67,12 +92,8 @@
             type=ref,event=tag
           platforms: ${{ inputs.platforms }}
           image: ${{ matrix.image }}
-          # Publishing imports only the caches warmed by native release checks on main.
-          cache_from: |
-            ${{ !inputs.publish && format('type=gha,scope=docker-{0}-{1}{2}', inputs.build_profile, matrix.target, inputs.cache_scope_suffix) || '' }}
-            ${{ inputs.publish && format('type=gha,scope=docker-{0}-{1}-amd64', inputs.build_profile, matrix.target) || '' }}
-            ${{ inputs.publish && format('type=gha,scope=docker-{0}-{1}-arm64', inputs.build_profile, matrix.target) || '' }}
-          cache_to: ${{ !inputs.publish && format('type=gha,scope=docker-{0}-{1}{2},mode=max', inputs.build_profile, matrix.target, inputs.cache_scope_suffix) || '' }}
+          cache_from: ${{ steps.cache.outputs.cache_from }}
+          cache_to: ${{ steps.cache.outputs.cache_to }}
           build_args: |
             BUILD_TARGET=${{ matrix.target }}
             BUILD_PROFILE=${{ inputs.build_profile }}
```

ok300 (dismissed):
Generally looks good.

NIT: proposed a small diff to extract `cache_from` and `cache_to`, to make things more readable.

.github/workflows/docker-build.yml
andrei-21: Nice idea with the explicit step. I made it a bit differently to avoid explicitly mentioning platform suffices.

ok300 (approved):

PR #608, opened 2026-09-11
feat(sdk): configure error responses limits

### LINKED ISSUE #604: Bounded raw/streaming GET for authenticated storage (unbounded error-body allocation)
Context: pubky-noise's encrypted session-backup loading needs to bound memory allocation when talking to a potentially malicious homeserver (pubky/pubky-noise#23 review thread: https://github.com/pubky/pubky-noise/pull/23#discussion_r3894926949).

**Problem**

`SessionStorage::get()` consumes *every* response body — including non-2xx error bodies — via `response.text().await` before returning control to the caller. A malicious homeserver can therefore return a small `HEAD` response (or a small `Content-Length`) followed by an arbitrarily large error body, forcing an unbounded allocation in the client before any caller-side size cap can run.

pubky-noise mitigates what it can from the caller side (HEAD pre-check via `storage.stats()`, chunked capped reads for 2xx bodies), but the error-body gap is only closable inside the SDK.

**Ask**

Expose a raw/streaming response path for authenticated storage that lets callers apply a byte cap *before* status handling consumes the body — e.g.:

- a `SessionStorage::request(...)`-style API returning the raw `reqwest::Response` (currently crate-private), or
- a `get_bounded(path, max_bytes)` helper that streams and aborts once the cap is exceeded, regardless of status code.

**Environment**

Observed with pubky 0.11.0 (crates.io).

MCarlomagno:
Closes [https://github.com/pubky/pubky-homeserver/issues/604](https://github.com/pubky/pubky-homeserver/issues/604)

### Summary

Bounds HTTP error bodies across SDK requests. Previously, a server could make the SDK collect an arbitrarily large error body.

### Behavior

- Defaults to capturing 4096 bytes.
- Stops reading after detecting overflow and appends `[response body truncated at N bytes]`.
- Setting `0` skips error-body reads and uses the HTTP status reason.
- Preserves error status codes and successful-response handling.

Configure the limit when creating the client:

```rust
use pubky::{Pubky, PubkyHttpClient};

let client = PubkyHttpClient::builder()
    .max_error_body_bytes(1024)
    .build()?;

let pubky = Pubky::with_client(client);
```

```typescript
import { Client, Pubky } from "@synonymdev/pubky";

const pubky = Pubky.withClient(
  new Client({ maxErrorBodyBytes: 1024 }),
);
```

### Acceptance criteria

- [x] Covers default, custom and zero limits, boundary sizes, UTF-8 decoding and body-read failures.
- [x] Returns after overflow without waiting for EOF; verifies browser stream cancellation.
- [x] Preserves grant and legacy-cookie private reads.
- [x] Applies the configured limit to refresh errors and prevents the storage GET when refresh fails.
- [x] Passes native tests, Node/browser regressions, formatting, Clippy and WASM checks.

### DIFF
--- a/pubky-sdk/README.md
+++ b/pubky-sdk/README.md
@@ -70,6 +70,28 @@ println!("Your current homeserver: {:?}", resolved);
 # Ok(()) }
 ```
 
+## Error-body limits
+
+HTTP status checking leaves successful response bodies unread. For HTTP errors,
+the client captures up to 4096 body bytes for the error message. Change this limit when creating the client:
+
+```rust,no_run
+use pubky::{Pubky, PubkyHttpClient};
+
+let client = PubkyHttpClient::builder()
+    .max_error_body_bytes(1024)
+    .build()?;
+let pubky = Pubky::with_client(client);
+# Ok::<_, pubky::BuildError>(())
+```
+
+The setting applies to all checked requests made through that client, including
+credential-refresh errors. Set it to `0` to skip error-body reads and use the
+HTTP status reason as the message. Positive limits preserve short messages and
+mark longer ones with `[response body truncated at N bytes]`.
+
+In JavaScript, use `Pubky.withClient(new Client({ maxErrorBodyBytes: 1024 }))`.
+
 ## Key formats (display vs transport)
 
 `PublicKey` has two string representations:
--- a/pubky-sdk/bindings/js/src/client/constructor.rs
+++ b/pubky-sdk/bindings/js/src/client/constructor.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 use tsify::{Ts, Tsify};
 use wasm_bindgen::prelude::*;
 
-use crate::js_error::{JsResult, PubkyError, PubkyErrorName, deserialize_ts};
+use crate::js_error::{JsResult, PubkyError, PubkyErrorName};
 
 // ------------------------------------------------------------------------------------------------
 // JS style config objects for the client.
@@ -31,7 +31,9 @@ pub struct PubkyClientConfig {
     /// Configuration on how to access pkarr packets on the mainline DHT.
     #[tsify(optional)]
     pub(crate) pkarr: Option<PkarrConfig>,
-    // NOTE: user_max_record_age belonged to the old client — removed here.
+    /// Maximum HTTP error-body bytes to capture. Defaults to 4096; zero skips the body.
+    #[tsify(optional)]
+    pub(crate) max_error_body_bytes: Option<usize>,
 }
 
 /// Low-level HTTP bridge used by the Pubky facade and actors.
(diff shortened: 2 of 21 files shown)


pubky-sdk/src/actors/auth/cookie/credential.rs
```diff
@@ -196,7 +195,7 @@ impl CookieCredential {
         .await?;
         let response = request.body(token.serialize()).send().await?;
 
-        let response = check_http_status(response).await?;
+        let response = client.check_http_status(response).await?;
```
MCarlomagno, line 198: most code changes in this PR are like this one, now we need the `check_http_status` function embedded in the `client` as the `client` determines the max body size to process

pubky-sdk/src/util.rs
```diff
@@ -1,24 +1,64 @@
...
+            }
+        }
+        drop(chunks);
 
-    Err(Error::from(RequestError::Server { status, message }))
+        // Native reqwest retains a UTF-8 BOM; browser Response.text() strips it.
+        let body = body.as_slice();
+        #[cfg(target_arch = "wasm32")]
+        let body = body.strip_prefix(b"\xef\xbb\xbf").unwrap_or(body);
+        let mut message = String::from_utf8_lossy(body).into_owned();
```
andrei-21, line 57: `String::from_utf8_lossy(body)` can expand invalid UTF-8: each invalid byte may become the three-byte `U+FFFD` replacement character. Consequently, although `body.len()` is capped at `self.max_error_body_bytes`, `message.len()` can reach roughly `3 * limit`, plus the truncation marker. If this setting is intended to bound the final error message or its memory usage, the decoded output should also be capped; otherwise, the documentation should clarify that the limit applies only to raw response bytes captured.

pubky-sdk/README.md
```diff
@@ -70,6 +70,28 @@ println!("Your current homeserver: {:?}", resolved);
...
+    .max_error_body_bytes(1024)
+    .build()?;
+let pubky = Pubky::with_client(client);
+# Ok::<_, pubky::BuildError>(())
+```
+
+The setting applies to all checked requests made through that client, including
+credential-refresh errors. Set it to `0` to skip error-body reads and use the
+HTTP status reason as the message. Positive limits preserve short messages and
+mark longer ones with `[response body truncated at N bytes]`.
```
andrei-21: Maybe we need to mention the new line appended:
```suggestion
mark longer ones with `\n[response body truncated at N bytes]`.
```

pubky-sdk/src/util.rs
```diff
@@ -1,24 +1,64 @@
...
-        return Ok(response);
+impl PubkyHttpClient {
+    pub(crate) async fn check_http_status(&self, response: Response) -> Result<Response> {
+        if response.status().is_success() {
+            return Ok(response);
+        }
+        let status = response.status();
+        let message = if self.max_error_body_bytes == 0 {
+            // Dropping the stream also cancels the browser reader without polling it.
+            drop(response.bytes_stream());
```
andrei-21:
```suggestion
            // `bytes_stream()` acquires the browser body reader; dropping the unpolled
            // stream cancels the underlying `ReadableStream`.
            drop(response.bytes_stream());
```

pubky-sdk/src/util.rs
```diff
@@ -1,24 +1,64 @@
...
+                break;
+            }
+        }
+        drop(chunks);
 
-    Err(Error::from(RequestError::Server { status, message }))
+        // Native reqwest retains a UTF-8 BOM; browser Response.text() strips it.
+        let body = body.as_slice();
+        #[cfg(target_arch = "wasm32")]
+        let body = body.strip_prefix(b"\xef\xbb\xbf").unwrap_or(body);
```
andrei-21:
```suggestion
        let body = body.as_slice();
        // Preserve the previous WASM `Response::text()` behavior, which strips a leading UTF-8 BOM.
        #[cfg(target_arch = "wasm32")]
        let body = body.strip_prefix(b"\u{FEFF}").unwrap_or(body);
```

pubky-sdk/src/util.rs
```diff
@@ -1,24 +1,64 @@
...
-pub async fn check_http_status(response: Response) -> Result<Response> {
-    if response.status().is_success() {
-        return Ok(response);
+impl PubkyHttpClient {
+    pub(crate) async fn check_http_status(&self, response: Response) -> Result<Response> {
+        if response.status().is_success() {
+            return Ok(response);
+        }
+        let status = response.status();
+        let message = if self.max_error_body_bytes == 0 {
```
andrei-21, line 18: The following statement will become easier to read if we introduce:
```rust
fn status_reason(status: StatusCode) -> String {
    status
        .canonical_reason()
        .unwrap_or("Unknown Error")
        .to_owned()
}
```

pubky-sdk/bindings/js/src/client/constructor.rs
```diff
@@ -52,26 +54,36 @@ impl Client {
...
+            .transpose()
+            .map_err(|error| PubkyError::new(PubkyErrorName::InvalidInput, error))?;
         let mut builder = pubky::PubkyHttpClient::builder();
 
+        if let Some(config) = config_opt.as_ref()
+            && let Some(limit) = config.max_error_body_bytes
+        {
+            builder.max_error_body_bytes(limit);
+        }
         if let Some(config) = config_opt
```
andrei-21: We can merge these two `if`s to make them simpler.

pubky-sdk/src/util.rs
```diff
@@ -1,24 +1,64 @@
...
+        while let Some(chunk) = chunks.next().await {
+            let chunk = chunk?;
+            let remaining = limit - body.len();
+            body.extend_from_slice(&chunk[..chunk.len().min(remaining)]);
+            if chunk.len() > remaining {
+                truncated = true;
+                break;
+            }
+        }
+        drop(chunks);
```
andrei-21, line 52:
```suggestion
        // Cancel the unread remainder before decoding the captured prefix.
        drop(chunks);
```

pubky-sdk/src/util.rs
andrei-21: Maybe we should strip it on native targets.

pubky-sdk/src/util.rs
MCarlomagno: I think just explicit documentation is fine for this case, I added it to the readme section

andrei-21 (approved):

PR #610, opened 2026-09-14
chore: bump `opendal` dependency

ok300:
This PR bumps `opendal` from 0.54 to the latest version that doesn't need an MSRV bump (0.57).

Changelog: https://github.com/apache/opendal/blob/main/CHANGELOG.md


andrei-21 (approved):
As a follow-up PR, we could use the async `delete()` API to check permissions directly and remove local buffering. We should cover the behavior change with tests: earlier allowed deletions could proceed before a later forbidden path fails.

PR #624, opened 2026-09-15
feat(sdk): implement guard against oversized payloads

### LINKED ISSUE #457: Events-stream API: add defense against Oversized Payloads
The SDK exposes an API for interfacing with the events stream, via `EventStreamBuilder`.

Once subscribed, it reads the bytestream and parses events via `parse_sse_event`.

However, it doesn't enforce any upper bounds on the byte size limit for a valid event. This means a malicious HS could, for example, return extremely long resource paths which, from what we can tell, can be arbitrarily large as long as they have the right format (Oversized Payload). This could be used to DoS a client reading events via the events stream API.

A few ideas on how this can be mitigated:
- the events stream API could enforce a hard cap on a response's byte length. If the stream sends more than this (configurable?) limit, the API throws an error.
  -  would only work for cases with `live = false`
- the event parsing (or `ResourcePath::parse`, which is the only variable-length component of an event) could itself enforce an upper limit on the max size of a valid resource path
  - would work for both `live = true` and `live = false`

MCarlomagno:
Closes [https://github.com/pubky/pubky-homeserver/issues/457](https://github.com/pubky/pubky-homeserver/issues/457)

### Summary

Adds an optional client side byte limit for SSE blocks received through the SDK’s `/events-stream` subscriptions:

- Rust: `.max_event_bytes(8192)`
- JavaScript: `.maxEventBytes(8192)`

Subscriptions remain unbounded by default and retain the existing parser. No breaking changes or downstream migration required, clients should enable protection explicitly.

This change is scoped to SSE parsing. HTTP error-body limits and general response limits remain separate.

### Behavior

- Applies to historical and live streams, with a fresh byte budget for each block and no cumulative stream limit.
- Counts all fields, comments, line endings, and a leading UTF-8 BOM. Excludes the final blank separator.
- On overflow, immediately closes the response, yields one validation error, and terminates without waiting for the block to finish or the connection to close. Previously delivered events remain available.
- Requires a positive limit. Rust rejects zero at subscription time; JavaScript accepts integers in `1..=4294967295`.
- JavaScript `maxEventBytes` and `addUsers` return new builders and preserve the original builder, including after validation errors.
- Bounded decoding replaces invalid UTF-8 and discards incomplete events at EOF. The default parser retains its existing behavior, including trailing UTF-8 errors.

Parser memory is proportional to the configured limit. Transport buffers and events retained by application code are outside its scope.

### Acceptance criteria

- [x] Rejects oversized and unterminated blocks and releases the response.
- [x] Preserves large legacy events when no limit is configured.
- [x] Handles fragmented SSE, split UTF-8, and long streams.
- [x] Allows JS builders to subscribe after catching validation errors.
- [x] Passes 225 Rust tests, 26 event integration tests, and 636 assertions each in Node and browser.
- [x] Passes formatting and native/WASM Clippy checks.


pubky-sdk/bindings/js/src/actors/event_stream.rs
```diff
@@ -142,8 +166,9 @@ impl EventStreamBuilder {
     ///
     /// @returns {EventStreamBuilder} - Builder for chaining
     #[wasm_bindgen]
-    pub fn live(self) -> Self {
-        EventStreamBuilder(self.0.live())
+    pub fn live(&mut self) -> Self {
+        self.0 = self.0.clone().live();
+        EventStreamBuilder(self.0.clone())
```
andrei-21: Why did you change the behavior?
Note the PR description states it differently:
> JavaScript `maxEventBytes` and `addUsers` return new builders and preserve the original builder, including after validation errors.

Now you do not preserve the original builder.

pubky-sdk/src/actors/event_stream/decoder.rs
andrei-21, line 1: Could we consider replacing `eventsource-stream` with [`sse-core`](https://docs.rs/sse-core/0.2.3/sse_core/struct.SseDecoder.html) instead of maintaining a custom SSE decoder? It provides configurable payload limits via `SseDecoder::with_limit`, potentially leaving us with a small adapter for Pubky event parsing and termination on overflow.

Its limits apply separately to accumulated data, event name, and ID—not total block bytes—so the semantics would differ, but may still satisfy #457. We should verify that against the overflow, cancellation, and WASM tests.

pubky-sdk/bindings/js/src/actors/event_stream.rs
MCarlomagno: AI did it for DX convenience after it was flagged by a security review, but this change is not related to the issue, so I will rollback the change and keep it as before

pubky-sdk/Cargo.toml
```diff
@@ -29,13 +29,15 @@ pubky-common.workspace = true
 thiserror.workspace = true
 async-trait.workspace = true
 eventsource-stream.workspace = true
+sse-core = { version = "0.2.3", default-features = false, features = ["stream"] }
```
andrei-21, line 32:
```suggestion
sse-core = { version = "0.2", default-features = false, features = ["stream"] }
```

pubky-sdk/src/actors/event_stream.rs
```diff
@@ -483,30 +513,88 @@ impl EventStreamBuilder {
...
+    {
+        let limit =
+            NonZeroUsize::new(limit).expect("event byte limit was validated before decoding");
+        let mut events = SseStream::with_decoder(SseDecoder::with_limit(limit));
+        events.attach(source.then(|chunk| async move {
+            // Allow cancellation even when incoming chunks are always ready.
+            futures_lite::future::yield_now().await;
+            chunk.map(Cursor::new)
+        }));
+        // Drop the response before yielding an error; sse-core otherwise allows recovery.
```
andrei-21:
```suggestion
        // Consume the decoder state on error so the attached response body is
        // dropped before yielding it; sse-core can otherwise recover after overflow.
```

pubky-sdk/src/actors/event_stream.rs
```diff
@@ -483,30 +513,88 @@ impl EventStreamBuilder {
...
-        let event_stream = sse_stream.filter_map(|result| async move {
+        Ok(Self::response_event_stream(response, self.max_event_bytes))
+    }
+
+    fn response_event_stream(
+        response: reqwest::Response,
+        max_event_bytes: Option<usize>,
+    ) -> impl Stream<Item = Result<Event>> {
+        let source = response.bytes_stream();
+        let sse_stream = if let Some(limit) = max_event_bytes {
```
andrei-21: Maybe we can set the limit to `usize::MAX` by default? It is more than any machine can handle.
Then we can remove the `eventsource-stream` dependency.

pubky-sdk/Cargo.toml
MCarlomagno: We need to stick to `0.2.3` here, `0.2.0` has a bug where some input bypasses the size limit. I tested locally an 8byte limit, and it accepted 99 bytes. The version `0.2.3` correctly rejected the same input.

pubky-sdk/src/actors/event_stream.rs
MCarlomagno: Agree

pubky-sdk/Cargo.toml
andrei-21: Ah-nice. Do you mind adding a comment about it?

andrei-21 (dismissed):

andrei-21 (approved):

PR #625, opened 2026-09-16
chore: track ci-workflows main

mdozhdev:
Moves the `pubky/ci-workflows` references back to `@main`.

## Why

`ci-workflows` holds only our own CI code. The third-party actions it calls — `docker/build-push-action`, `docker/login-action` and the rest — are SHA-pinned **inside** that repo, and stay pinned. So the supply-chain property is unchanged by this: we are not resolving anyone else's code at a moving ref.

What pinning bought on top of that was a second review of our own changes, once per consuming repo, when each bump PR landed. That is 8 pull requests per release for code we wrote ourselves, and it delays fixes reaching the pipelines that need them.

The same property is better served by requiring review **once, in the repo where the change is written**. Tracked in pubky/ci-workflows#10.

## What is unaffected

- **Third-party pins inside `ci-workflows`** — untouched
- **Dependabot in this repo** — unchanged. It has never proposed a `ci-workflows` bump; every PR it has raised here is for third-party actions, cargo or npm dependencies. That coverage stays.
- **Tags and releases on `ci-workflows`** — still cut, still the changelog

## What changes

A merge to `ci-workflows` `main` now reaches this repository's next build directly, with no bump PR in between.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SpontaneousOverthrow (approved):
LGTM

PR #628, opened 2026-09-21
docs(sdk): fix grant session persistence example

gcomte:
The Rust persistence example signs in to a grant session, then calls the cookie-only and deprecated `write_secret_file`, which panics. Use `export_local_secret` and `restore_session` with an existing account's recovered key instead.


talosmachina (commented):
**No findings.** Fixes the grant-session persistence example in the SDK README: it now signs in, reads the grant view, exports the local secret with `export_local_secret()`, and restores with `Pubky::restore_session()`, instead of calling the cookie-only `write_secret_file()`, which panics on a grant session. Reviewed `c3def65`, docs-only diff, every API and behaviour claim checked against the implementation.

<details><summary>What I checked, and 4 candidates I ruled out</summary>

**Read in full**: `pubky-sdk/README.md`; `pubky-sdk/src/pubky.rs` (`restore_session`, `signer_from_recovery_file`), `pubky-sdk/src/actors/auth/cookie/legacy_api.rs`, `pubky-sdk/src/actors/auth/cookie/view.rs`, `pubky-sdk/src/actors/auth/grant/view.rs`, `pubky-sdk/src/actors/signer/session.rs`
**Call sites traced**: `signin_with_publish` -> `credential_from_grant_exchange` -> `from_grant_credential`; `as_grant` -> `try_downcast_credential`; `Pubky::restore_session` -> `GrantCredential::is_secret_token` -> `import_grant_secret`
**CI**: green (18/18, incl. doc (pubky), examples, test (pubky))

**Ruled out**
- *`signin` may return a cookie session, so `as_grant()` would return `None` and the example's `expect` would panic*: `signer/session.rs` `signin_with_publish` builds a `GrantCredential` and `PubkySession::from_grant_credential`; `as_grant` only returns `None` for cookie-backed sessions.
- *`write_secret_file` fails with an `Err` rather than panicking*: `cookie/legacy_api.rs` calls `self.as_cookie().expect("write_secret_file() is only valid for cookie sessions")`, so a grant session panics exactly as the new note says.
- *`restore_session` rejects a grant secret*: `pubky.rs` dispatches on `GrantCredential::is_secret_token` and routes to `import_grant_secret`.
- *The new intra-doc links to the deprecated items break the doc build*: `doc (pubky)` is green.

</details>

Merge confidence: 5/5, docs-only change, CI green, and the example is the one the README's own doctest block compiles.

PR #630, opened 2026-09-21
feat(homeserver): Basic Webdav locks

86667:
# feat(homeserver): WebDAV locks on storage files

Alternative to #577 + #578, and a companion or alternative to #606. Gives clients an atomic read-modify-write on a small storage file with locks alone. The change lives in the route handlers plus one table; the storage and finalization layers are untouched.

This PR is deliberately only the lock protocol. Working through it surfaced several gaps, some in the write path on `main` and some in the locks themselves. Closing these will come as a follow-up PR. They are listed under "Missing parts" with the reason each is needed. None is included here.

## Summary

`/storage/{user}/{path}` accepts the WebDAV `LOCK` and `UNLOCK` methods (RFC 4918) on file paths. A lock is an exclusive write lock on one path, held as a bearer token. While it lives, `PUT` and `DELETE` on that path must present the token in the `If` header; without it they return `423 Locked`. A token that is not the live lock returns `412 Precondition Failed`.

A write checks the path's lock once, before its body streams. A write that presents the token runs under that lock, and the server keeps the lock alive until the write ends. A write that presents none holds nothing: it runs if no live lock exists and is refused otherwise. A client that holds the lock from before its `GET` until after its `PUT` therefore gets an atomic read-modify-write against every writer that starts after the lock is granted. Other lock-aware writers are refused at `LOCK` until it unlocks; writers that send no token, or a stale one, are refused before their write starts. No content comparison is needed.

## Behaviour

- **LOCK** on a file path under `/pub/` or `/priv/`, with a session holding write capability. The path need not exist; locking an unmapped path reserves it without creating anything. `Timeout: Second-N` requests a lifetime: at most 60 seconds, 30 when absent, 60 for `Infinite`. A client that needs longer refreshes. A `lockinfo` body is optional; only an exclusive write lock is granted, and a shared request is `400`. The body is read only once the request is authorized, and one over 16 KiB is `413`. Response: `200` with a `lockdiscovery` XML body, a `Timeout` header and `Lock-Token: <opaquelocktoken:...>`. `423` when another live lock exists.
- **Refresh**: `LOCK` with the held token in `If`. `412` when the token is not the live lock.
- **UNLOCK** with `Lock-Token`: `204`, or `409` when no lock with that token exists on the path, `400` without the header.
- **PUT / DELETE**: locked and no token → `423`; token matches → proceeds; tokens that name no live lock on the path → `412`, so a client whose lock expired never writes unlocked by accident. Only tokens are read from `If`; resource tags, entity tags and `Not` are ignored.
- **No behaviour change for clients that never lock.** Unlocked writes to one path remain last-writer-wins and cost one primary-key lookup each. They see `423` only on a path someone has explicitly locked.
- `LOCK` and `UNLOCK` exist on `/storage/{user}/{path}` only. The deprecated owner-relative routes do not offer them, but their `PUT` and `DELETE` share the handlers, so they obey locks the same way: `423` while a path is locked, unless they carry the token in `If`.
- `/info` advertises `webdav-locks`. CORS exposes `Lock-Token` and `Timeout` so browser clients can read a granted lock.
- **SDK**: `SessionStorage` gains `lock`, `refresh`, `unlock`, `put_locked` and `delete_locked`. `put_locked` and `delete_locked` take the path from the lock, so a token can never be presented for the wrong path.

## Guarantees

1. **A held lock excludes every write that starts after it.** The handler checks the lock before the body streams: a write without a token is refused while a lock is live, and a write with a token is refused unless it names the live lock. A `LOCK` granted while an unlocked upload is already streaming does not stop that upload from landing; see "Deliberately not included".
2. **One winner per path.** Acquisition is a single `INSERT ... ON CONFLICT DO UPDATE ... WHERE expired` statement, so of two racing `LOCK` requests exactly one takes the path.
3. **A lock cannot expire under a write made with its token.** The handler extends the client's lock every 10 seconds to at least 30 seconds ahead, until the write ends. A longer `Timeout` is never shortened; a shorter one is lengthened, so after a write under it a client's lock lasts at least 30 seconds more. Without this, a slow upload could outlive its lock, another client could lock, read and write, and one of the two updates would be lost silently.
4. **The keep-alive stops with the write.** It is aborted when the write ends, committed or failed, and when the request future is dropped because the client disconnected. If the process dies, the keep-alive dies with it and the lock expires within 30 seconds.
5. **Shared across instances.** Locks live in one table in Postgres. Expired rows are ignored by every read, overwritten by the next acquisition, and swept on each `LOCK`; no background task.

## Missing parts, and why each is needed

These are known gaps in this PR. The first group is in the write path on `main`; locks do not cause them, but locks make them matter, because the protocol tells a holder that what it reads is the latest committed content. The second group is in the locks themselves.

### Write path on `main`

1. **Staged uploads on the filesystem backend.** That backend streams a body straight into the final file. A client disconnect, a broken stream, or a quota rejection at finalization has therefore already overwritten part of the file, while the entry row still describes the old content. The next lock holder's `GET` reads those torn bytes. For a file that is a client's single source of truth, such as an encrypted state snapshot, the old content is gone and the new is incomplete; if no client still holds a good copy, the state is lost. Fix: set `atomic_write_dir` to a sibling `data/files-tmp`, so an upload is a temp file renamed into place on close and the file is always one complete body. Memory and GCS already publish only on close.
2. **Abort rejected and abandoned uploads.** Needed alongside staging, or rejected uploads pile up in the staging directory. The finalizer should abort the backend writer when it rejects on quota or a collision, and an upload whose request future is dropped should abort through a guard. The existing abort on a broken stream only takes effect once staging is on, and today its error can mask the stream error.
3. **Run finalization on its own task.** Finalizing a write publishes the blob and then commits the entry row; a delete commits the row and then removes the blob. Both run inside the request future, so a client that disconnects at that moment cancels it between the two steps and leaves row and blob disagreeing. Fix: `tokio::spawn` the writer's `close()` and the deletes and await the handle, so the client only loses the response. When this lands, the keep-alive should move onto that task too, so a dropped request does not stop extending the lock while the write is still committing.
4. **A stale entry row after a process death between publish and commit.** Even with 1 to 3, a server that dies in that window leaves the row, the event feed and the quota describing the old content. `GET` takes `Content-Length` and `ETag` from the row and the body from the blob, so the response is cut short when the lengths differ, and a revalidating client gets a wrong `304`. No data is lost and the next `PUT` repairs the row, but a holder that must read first can be stuck. Small mitigations: take `Content-Length` from the blob, and never answer `304` for a locked path. Closing it properly is what #606's repair step or #577's pointer switch are for.

### Locks

5. **No check when the write commits.** A write under a token checks the lock before the body streams and then keeps it alive. The lock can still go away under the write in two ways, and in both another client takes the path while the first write still publishes. The keep-alive cannot reach the database for a whole lifetime, so the lock expires. Or another holder of the same token sends `UNLOCK`, or refreshes it with a `Timeout` shorter than the keep-alive interval, while the upload is in flight. The second is self-inflicted, since only the token's holder can do it, but the server logs "Lock lost" and carries on. Fix: for uploads that ran longer than one keep-alive interval, confirm the lock once the body has arrived and abandon the write if it is gone. With staged uploads an abandoned write touches nothing. A check inside the finalization transaction would be stricter still, but needs the tokens threaded through OpenDAL into the finalization layer.
6. **A deleted file keeps its lock** until it expires or is unlocked. The RFC drops it on delete. Harmless, and it lets a holder delete and recreate under one lock, but it is a deviation.
7. **The admin API** does not honour these locks, neither WebDAV at `/dav` nor the entry delete endpoint, so an admin write can land under a client's lock.
8. **A lock can only be released with its token.** There is no lock discovery and no override, so the owner cannot break a lock that another of their sessions keeps refreshing; revoking that session frees the path within 60 seconds. Fix, if it comes up: let `UNLOCK` with a root capability ignore the token.
9. **No JS bindings.** The Rust SDK has the lock methods; the JS bindings do not expose them yet.

Items 1, 2, 3, 5 and 6, and the mitigations in item 4, have been prototyped on top of this commit and are small. They were held back to keep this PR to the protocol.

## Client contract

For a client that needs read-modify-write on a mutable file:

1. `LOCK` the path. A lock lasts at most 60 seconds; refresh it with `LOCK` and the token in `If` if the cycle runs longer. On `423`, wait and retry; someone is mid-update, or has crashed and the lock will expire.
2. `GET` the file. No other write can start while you hold the lock. Verify the body before merging into it, for example by decrypting it or checking its hash against the `ETag`: until missing parts 1 to 4 land it may be torn, and an unlocked upload that was already streaming when you locked can still land after your read.
3. `PUT` with `If: (<opaquelocktoken:...>)`. On `412`, the lock expired: re-`LOCK`, re-read, retry. The upload itself need not fit in the lock's lifetime; the server keeps the lock alive while it runs.
4. `UNLOCK` with `Lock-Token`.

Immutable, write-once files: `LOCK` the unmapped path, check it is still absent, `PUT` with the token, `UNLOCK`.

Clients that never lock need no change. They only see `423` on a path someone has explicitly locked, and can retry once the lock ends.

A crashed lock holder blocks its path until the lock expires, at most 60 seconds. It never deadlocks a path and never causes a lost update.

## Design notes

- `LOCK` and `UNLOCK` are not in axum's method filter set, so they reach the route through the method router's fallback (`routes/tenants/lock.rs`), which answers `405` with `Allow` to anything else. The handler takes the body unread, so nothing is buffered for an unknown method or an unauthorized request; an authorized `LOCK` reads at most 16 KiB.
- One table, `entry_locks` (`path` primary key, `token`, `expires_at`), via one migration. No foreign key: the path key carries the owner, and a row is dead the moment it expires.
- `with_write_lock` is the single entry point for `PUT` and `DELETE`: it takes the write as a future, checks the lock, and for a write under a client's lock keeps that lock alive until the future returns. The keep-alive is a task per such write, aborted when the future returns, failed or not, and by `Drop` when the request future is dropped. In `PUT` the check happens after the `Content-Length` quota pre-check, so a request rejected there never touches the lock table.
- A write without tokens is one lookup of the live lock (free, or `423`). A write with tokens is one `keep_alive ... WHERE token IN (presented)` (the live lock, or `412`), so nothing is read first and the lock cannot change between the check and the hold. Refresh works the same way.
- `refresh` and `keep_alive` are one statement that differs only in the new expiry: the requested one, or `GREATEST` of the current and the keep-alive horizon, so a keep-alive only ever extends.
- The `lockinfo` body and the `If` and `Lock-Token` headers are read with small string scanners matched on local element names, so any namespace prefix works and no XML dependency is added.
- Cost on the write path: one primary-key lookup per unlocked `PUT` and `DELETE`. A write under a lock adds one update before it starts, and a keep-alive query every 10 seconds only if it runs that long.
- `LOCK` reads and checks its `lockinfo` body before it reads the clock, so a slowly sent body cannot eat into the lifetime that is granted and reported.
- New `HttpError` helpers for `409`, `412` (token mismatch) and `423`.

## Deliberately not included

- **Everything under "Missing parts".** Kept out so the protocol can be reviewed on its own.
- **Excluding uploads already in flight.** An earlier revision had every unlocked write take an implicit lock for its duration, so a `LOCK` during an upload was refused. That made concurrent unlocked writes to one path exclude each other, added an upsert and a delete to every write, and needed an upload idle timeout so a silent client could not hold a path. It was dropped so that clients which never lock see no change. The cost is that a lock taken while an unlocked upload is streaming does not stop that upload from landing, which is why the client contract asks holders to verify what they read.
- **Shared locks and `Depth: infinity`.** Exclusive write locks on single files are the whole requirement.
- **Lock-null resources.** The RFC creates an empty resource when locking an unmapped path. Recording the lock without creating anything serves create-once flows and avoids an empty-file write and event.
- **A configuration knob for timeouts.** Constants in `lock.rs`; 30 and 60 seconds are easy to change if a client needs more.

## Relation to #606

#606 gives compare-and-set through `If-Match` on content-hash ETags. Its correctness depends on the entry row matching the blob, which is why it verifies and repairs the row under the user lock on every conditional write, and why it changes the finalization layer. Locks do not depend on the row for mutual exclusion: nobody else can start a write while the lock is held. They do depend on the blob being whole, which is what missing parts 1 to 3 provide; #606 already contains those three. The two approaches are independent and can coexist. ETags remain the better fit for optimistic clients that want to avoid the `LOCK` and `UNLOCK` round trips, cannot hold a lock open, or write large files.

## Relation to #577

#577 makes publication crash-safe with immutable blobs and a pointer switch, which removes missing parts 1 to 4 structurally. This PR keeps the storage model and adds one table and one migration.


e2e/src/tests/storage/locks.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    // A locked delete leaves the lock in place.
+    let response = storage.delete_locked(&lock).await.unwrap();
+    assert_eq!(response.status(), StatusCode::NO_CONTENT);
+    let plain = storage.put(path, vec![2]).await;
+    assert_eq!(status_of(plain.unwrap_err()), StatusCode::LOCKED);
+
+    // Unlocked: the old lock is refused everywhere.
+    storage.unlock(&lock).await.unwrap();
+    let refresh = storage.refresh(&mut lock, Duration::from_secs(10)).await;
```
SeverinAlexB: `storage.refresh()` seems like it's refreshing the lock, is this correct? If yes, can we rename it to `refresh_lock()`? `refresh()` could mean anything.

talosmachina (commented):
**1 finding (1 P1).** Adds RFC 4918 `LOCK`/`UNLOCK` on storage files, a lock-token check in `If` on `PUT`/`DELETE`, and the SDK plus OpenAPI surface for both. Reviewed `3463e2e`, oversized tier: the lock protocol, the migration and the SDK/OpenAPI contract in full, the rest of the 2k-line diff not read.

<details><summary>What I checked, and 3 candidates I ruled out</summary>

**Read in full**: `routes/tenants/lock.rs`, `persistence/sql/entities/entry_lock.rs`, `m20260921_create_entry_locks`, `routes/tenants/write.rs`, `routes/tenants/authorize.rs`, `pubky-sdk/src/actors/storage/lock.rs`, `openapi-client.yml`
**CI**: `test (e2e)` red on this head; every other job green.

**Ruled out**
- *A token used on another path*: `keep_alive` filters on `path = ...`, so a token for a sibling path or another user is 412 and extends nothing. Pinned by `token_is_refused_on_any_other_path`.
- *Two writers racing for one path*: `acquire`'s `ON CONFLICT (path) DO UPDATE ... WHERE expires_at <= now()` decides in one statement, and the race test passes.
- *A lock outliving its lifetime*: expiry is compared against the database clock, and `expired_lock_frees_the_path` covers it.

</details>

Merge confidence: 1/5, oversized tier and CI red on the change itself: the P1 below turns the e2e concurrent-write test red.

pubky-homeserver/src/client_server/routes/tenants/lock.rs
```diff
@@ -0,0 +1,951 @@
...
+    let mut executor: UnifiedExecutor = sql_db.pool().into();
+    let held = if_header_tokens(headers);
+    if held.is_empty() {
+        if EntryLockRepository::get_active(entry_path, &mut executor)
+            .await?
+            .is_some()
+        {
+            return Err(HttpError::locked());
+        }
+        return write.await;
```
talosmachina: **<sub><sub>P1</sub></sub> Do not hold the pooled connection across `write.await`**

`executor` keeps its `PoolConnection` until it is dropped, and it stays alive across `write.await` on both exit paths, so every unlocked `PUT`/`DELETE` pins one pool connection for as long as its body streams. The write then needs a second connection for its finalization transaction (`write_finalization_layer/write.rs`, `self.sql_db.pool().begin()`), so with the default 10-connection pool ten concurrent uploads each take one and all wait for a second; nothing releases, and at the 30s acquire timeout every request returns 500. `tests::rate_limiting::test_concurrent_write_read` (10 concurrent PUTs) fails exactly that way on this head and passed on `1f2b84b`, where the executor was local to a `guard_write` that returned before the write. Scope the check so the connection is released before `write.await`, as the existing `uploads_beyond_pool_size_complete_for_same_and_cross_user_writes` test requires.

Verified by: `test (e2e)` red on this head (500 on all ten PUTs, 36.6s run = ~3s stream plus the 30s acquire timeout); the same job is green on `1f2b84b`, whose only functional change to this path is the executor lifetime.

SeverinAlexB (commented):
One issue, otherwise good.

pubky-homeserver/src/client_server/routes/tenants/lock.rs
86667: fixed!

86667: @talosmachina please confirm the fix.

talosmachina: Confirmed at `eab96cc`.

The fix is the right shape: `check_lock` now keeps its `UnifiedExecutor` local and returns only the `KeepAlive`, so the pool connection is released before `write.await`; `with_write_lock` therefore holds nothing across the body stream. That is exactly the lifetime that caused the starvation.

Verified on this head:

- `pubky-homeserver` `tenants::lock` lib tests: 15 passed, including `write_runs_only_when_the_lock_allows_it`, which pins the pool to one connection and takes a connection inside the write body, so it fails if any connection is still held across the write.
- `e2e` `rate_limiting::test_concurrent_write_read`: green in 11.9s, both 10-way phases under the 5s assertion. On `3463e2e` the same test returned 500 on all ten PUTs after the 30s acquire timeout, which was the finding.

Both run against `postgres://…@localhost:5432/postgres` as in CI. No remaining hold on the connection from this path.

SeverinAlexB (approved):

PR #637, opened 2026-09-22
feat(homeserver): Harden write path against client disconnects and late rejections

86667:
Here we implement two mechanisms:

1) The filesystem operator stages uploads in` data/files-tmp` and renames into place on close, so the file on disk is always one complete body. The finalizer aborts a write it rejects on quota or a collision, a guard aborts an upload whose request is dropped, and the broken-stream abort now works and no longer masks the stream error.

2) The writer's `close()`, `delete` and `admin_delete` run on a spawned task, so a client that disconnects after its last byte cannot cancel the commit between publishing the blob and updating the row.


pubky-homeserver/src/persistence/files/opendal/staged_upload.rs
```diff
@@ -0,0 +1,195 @@
...
+            writer.write(chunk_result?).await?;
+        }
+        Ok(())
+    }
+
+    /// Publish the staged bytes and commit their entry, on a task that
+    /// outlives a dropped request.
+    pub(super) async fn publish(mut self) -> Result<(), FileIoError> {
+        let mut writer = self.take_writer();
+        spawn_finalization(async move { writer.close().await })
```
talosmachina, line 50: **<sub><sub>P1</sub></sub> Put the finalization spawn in the layer, or the admin WebDAV mount keeps the bug being fixed**

The detached task is installed here and in `OpendalService::delete_with`, both above the layers, so it covers only the callers that go through `OpendalService`. The admin `/dav` mount hands `admin_operator` straight to `dav-server-opendalfs` (`admin_server/app_state.rs:17`), which drives `operator.writer(...)` / `operator.delete(...)` inline: a WebDAV PUT or DELETE whose client hangs up mid-finalization still drops the future between the blob rename and the row commit, leaving the blob published with the old entry row (or an unreferenced blob after a delete) exactly as the new `files` module docs say cannot happen. Moving the `spawn_finalization` calls into `WriteFinalizationWriter::close` and the deleter's `close` covers every operator consumer, including this one.

Verified by: read `OpendalFs::new(admin_operator)` at `admin_server/app_state.rs:17` and `dav-server-opendalfs-0.7.2/src/fs.rs:170` (delete) plus its `file.rs` write paths; the spawn is only reachable from `OpendalService` (`opendal_service.rs:166`, `staged_upload.rs:50`), and no `Drop` on opendal's `Writer` compensates.

talosmachina (commented):
**1 finding** (1 P1). Stages filesystem uploads in `data/files-tmp` and renames them into place on close, and moves write/delete finalization onto a detached task so a client disconnect cannot cancel the commit. Reviewed `6b368e4`, full tier.

<details><summary>What I checked, and 4 candidates I ruled out</summary>

**Read in full**: `staged_upload.rs`, `opendal_service.rs`, `opendal/finalization.rs`, `write_finalization_layer/{write,delete,layer}.rs`, `files/mod.rs`, `client_server/routes/tenants/{write,lock}.rs`, `admin_server/{app_state.rs,routes/dav_handler.rs}`, `file/file_service.rs`
**Call sites traced**: `FileService::{write_stream,delete,admin_delete}`, `client_server` PUT/DELETE, admin `/webdav` and `/dav`, `dav-server-opendalfs` 0.7.2
**opendal 0.57 checked**: `FsWriter::create` stages in `atomic_write_dir` (created at builder build), `close` = `rename`, `abort` = `remove_file`, no `Drop` on the async `Writer`
**CI**: green (clippy, fmt, msrv, 6 test jobs, 2 docker-build jobs pending)

**Ruled out**
- *Pool starvation / row lock across the body stream*: `check_lock` returns its connection before the write runs (`lock.rs:96-128`), so the body stream holds none; the spawned finalization needs one connection, same as before.
- *`FileService::write_stream`'s entry re-read*: `spawn_finalization` awaits the task, so the commit is done before the re-read.
- *Abort on a rejected upload inside the open transaction* (P2, dropped): `write_in_transaction` calls `abort_rejected_upload` while the tx holds `FOR NO KEY UPDATE` on the user row, so the backend abort now runs with the lock and a pooled connection held; on GCS that abort is a network call. Real but P2, below the posting bar.
- *Staging dir is never swept* (P3, dropped): an unclean shutdown or a failed `rename` leaves a file in `data/files-tmp` with nothing to prune it.

</details>

Merge confidence: 3/5, the P1 on the WebDAV surface.

pubky-homeserver/src/persistence/files/opendal/staged_upload.rs
86667: Good find, but this is fine. Admin webdav can act outside of Locks.

pubky-homeserver/src/persistence/files/opendal/staged_upload.rs
```diff
@@ -0,0 +1,195 @@
...
+            .await
+            .map(|_backend_metadata| ())
```
SeverinAlexB, line 52: AI finding:

```
Abort staged uploads when starting the database transaction fails
staged_upload.rs:49–53 (pubky-homeserver/src/persistence/files/opendal/staged_upload.rs#L49-L53)
publish() removes the writer from the drop guard and delegates cleanup to writer.close(). However, Finalizer::finalize_write() returns immediately if pool().begin() fails, before reaching the new abort-on-rejection handler.
Consequently, a database outage or pool-acquisition timeout after an upload finishes leaves its entire body permanently in data/files-tmp. Repeated failures can consume disk space without creating entries or charging quota.
Reproduced: stage an upload, close the database pool, then call publish(). Publication fails, but one staged file remains.
Suggested fix: abort the backend writer when transaction creation fails, covering this pre-publication failure alongside prepare_write() failures.
```

Image

@talosmachina You didn't flag this. Is this legit?

SeverinAlexB (commented):
Does this PR also work in case we horizontally scale the homeserver?

pubky-homeserver/src/persistence/files/opendal/staged_upload.rs
SeverinAlexB: @talosmachina hello?

talosmachina: @SeverinAlexB Legit. I checked it on `6b368e4` by reading the path, not by rerunning the repro.

`Finalizer::finalize_write` opens the transaction first (`write_finalization_layer/write.rs:126`). When `pool().begin()` fails it returns through `?` and never reaches `write_in_transaction`, so the new `abort_rejected_upload` at `write.rs:172` never runs. By then `StagedUpload::publish` has taken the writer out of the drop guard (`opendal/staged_upload.rs:49`) and moved it into the detached close task, so when that task drops the writer on the error path nothing aborts the backend writer. The full body stays in `data/files-tmp` with no entry row and no quota charge, and nothing sweeps that directory.

A pool-acquire timeout or a short DB outage at close is enough to trigger it. It is the concrete case of the "nothing prunes `files-tmp`" gap I left at P3 in my review, so I'd fix them as one thing rather than two.

Fix I would take: abort the backend writer on the begin failure too, i.e. call `abort_rejected_upload(backend_writer, entry_path)` before the `?`. That is under every caller, so the WebDAV mount gets it as well. Aborting is safe there because the backend writer has not been closed, so nothing has been renamed into place yet.

Severity: P2, a storage leak that needs the database to fail at close rather than data corruption, but I would not merge with it open.

86667:
> Does this PR also work in case we horizontally scale the homeserver?

Yes, since staged writes append 8 random chars to temp file names

pubky-homeserver/src/persistence/files/opendal/staged_upload.rs
86667: @talosmachina fixed. Please verify and re-review.

talosmachina: @86667 Verified on 8980d54. `Finalizer::finalize_write` now aborts the staged writer when `pool().begin()` fails (`write.rs:106-115`), before the error is returned, so the path I confirmed on 6b368e4 no longer leaves the staged bytes behind. `upload_is_aborted_when_the_transaction_cannot_begin` exercises it on the filesystem backend: one chunk written, pool closed, and it fails if the staged file survives the failed close. Re-review posted.

talosmachina (commented):
**Re-review of 8980d54 after 6b368e4.** Staged-upload hardening: an upload whose finalization transaction cannot begin is now aborted instead of left staged. 2 earlier findings: 1 fixed, 1 closed as intended by @86667. No new findings.

<details><summary>What I checked, and 4 candidates I ruled out</summary>

**Read in full**: `write_finalization_layer/write.rs`, `write_finalization_layer/layer.rs`, `opendal/staged_upload.rs`
**Diff scope**: 2 files, 78 changed lines
**Call sites traced**: `WriteFinalizationWriter::close` -> `Finalizer::finalize_write` -> `StagedUpload::publish`
**CI**: green (fmt, clippy, msrv, test (pubky-homeserver), docker-postgres-test, e2e)

**Earlier findings**
- *Finalization spawned above the layers leaves the admin WebDAV mount unfixed*: closed by @86667 as intended; the admin path acts outside the locks.
- *No abort when `pool().begin()` fails*: fixed in 8980d54 at `write.rs:106-115`, the staged writer is aborted before the begin error is returned, and `upload_is_aborted_when_the_transaction_cannot_begin` fails if it is not.

**Ruled out**
- *Abort after the backend published*: the abort sits on the `begin()` failure, upstream of `backend_writer.close()` (`write.rs:150`); nothing has been renamed, and the commit path still leaves published bytes alone.
- *The new test counting the wrong directory*: `get_fs_operator` stages in `<tmp>/files-tmp` (`opendal_test_operators.rs:152-156`), which is what the test counts, and its first assertion (one staged file while the upload is open) fails if staging moves.
- *Cleanup failure masking the real error*: `abort_unpublished_upload` logs and returns, so the caller still reports the begin failure to the client.
- *The failing `begin()` being test-only*: a closed, saturated or shutting-down pool gives the same error in production; `db.pool().close()` is a real route to it.

</details>

Merge confidence: 5/5, CI green and the new test covers the abort path.

86667:
@talosmachina Sev suggested bringing the finalisation task logic inside of an OpenDal layer, which ive implemented in [d56da9d](https://github.com/pubky/pubky-homeserver/pull/637/commits/d56da9dfc126abbd7af92d5fb819b67afb017481). Please check that commit.

talosmachina (commented):
**Re-review of `8980d54` after `d56da9d`.** Moves the finalization spawn out of `OpendalService` and into `WriteFinalizationWriter::close` and the deleter's `close`, so every operator consumer gets it, and gives the writer a `Drop` that discards a staged upload that is never closed. 2 earlier findings: 2 fixed. No new findings.

<details><summary>What I checked, and 4 candidates I ruled out</summary>

**Read in full**: `write_finalization_layer/{write,layer,delete}.rs`, `opendal/opendal_service.rs`, `write_path_layer.rs`
**Diff scope**: 9 files, 460/463 lines; `opendal/finalization.rs` and `opendal/staged_upload.rs` deleted
**Call sites traced**: `OpendalService::{delete,admin_delete,write_stream}`, `OpendalFs::new(admin_operator)` at `admin_server/app_state.rs:17`, `WritePathAccessor`/`WritePathDeleter` pass-through
**opendal 0.57**: no `Drop` on the high-level `Writer`, so dropping it drops the layer's writer and the new abort fires
**CI**: green (clippy, fmt, msrv, doc, examples, helper-crates, docker-build x2, docker-postgres-test, e2e, test x5, wasm-test)

**Earlier findings**
- *P1, spawn above the layers leaves the admin WebDAV mount unfixed*: fixed. The spawn now sits in `WriteFinalizationWriter::close` (`write.rs:115`) and `WriteFinalizationDeleter::close` (`delete.rs:161`); `admin_operator` carries its own `WriteFinalizationLayer` (`opendal_service.rs:71`) and is what `OpendalFs` is built from, so a `/dav` PUT or DELETE whose client hangs up no longer drops the future between the rename and the row commit.
- *No abort when `pool().begin()` fails*: still fixed, and reached through `WriteFailure::BeforePublication` after the rollback, so the abort runs with no transaction holding the user row.

**Ruled out**
- *Second `close` re-runs the deleter*: `self.queued` is restored after close, so opendal could close again; nothing in opendal closes a deleter twice and the queue is empty after a successful one.
- *Double abort on drop after a failed close*: `close` takes `inner` before it spawns, so `Drop` and `abort()` both see `None` and no-op.
- *Join-error path losing the domain error*: `spawn_finalization` passes the finalization's own error through unchanged, so the quota and collision sources still downcast; only a panic produces the join error.
- *Staged file left when the finalization task is cancelled*: pre-existing and named as a limit in the module docs, not this diff.

</details>

Merge confidence: 5/5, CI green and the writer-drop, caller-drop-on-close and begin-failure tests cover the changed paths.

talosmachina:
@86667 Checked `d56da9d`. Moving the spawn into the layer's writer and deleter `close` covers every operator consumer, including the admin WebDAV mount, so it closes the P1 from my first review. No new findings; re-review posted above. Merge confidence: 5/5.

